PR #11, opened 2026-04-07
Use SubagentTool to register subagent(speedwagon) to main agent

khj809:
## Overview

This PR refactors how Speedwagon knowledge-base agents are integrated into the main `ChatAgent`, replacing a spawn-per-call pattern with persistent in-memory subagents registered via ailoy's new `SubagentTool` API (depends on https://github.com/brekkylab/ailoy/pull/369). It also adds web search as a builtin tool on the main agent.

---

### 1. Speedwagon: Spawn-per-call → In-memory `SubagentTool`

**Before:** `build_speedwagon_tool` created a single `ask_speedwagon` tool with `kb_id` + `question` parameters. On each invocation, it spawned a fresh `AgentRuntime` via `dispatch_speedwagon()`, opened the search index, and tore everything down after the call.

**After:** `register_speedwagon_subagents` builds one `AgentRuntime` per KB entry at `ChatAgent` construction time, wraps each in `Arc<TokioMutex<AgentRuntime>>`, and registers it with `ToolSet::with_subagent_in_memory`. Each KB entry becomes its own named tool (`ask_speedwagon_{id}`), allowing the LLM to call the right knowledge base directly without a `kb_id` dispatch parameter.

**Files changed:**
- `chat-agent/src/speedwagon/dispatch.rs` — replaced `build_speedwagon_tool` + `dispatch_speedwagon` with `register_speedwagon_subagents`
- `chat-agent/src/speedwagon/mod.rs` — updated exports; removed `ASK_SPEEDWAGON_TOOL` constant
- `chat-agent/src/chat_agent.rs` — updated `build_tool_set` to call `register_speedwagon_subagents`

### 2. Tool naming: safe identifiers with human-readable frontend display

Speedwagon names (e.g. "새 Speedwagon") can contain Korean characters and spaces, which are not safe as tool names. Tool names are set to `ask_speedwagon_{uuid}` for API safety.

To keep the UI readable, the frontend resolves the UUID back to a human-readable name:

- `KbEntry` gains a `name` field populated from `Speedwagon.name` in `state.rs`
- `ToolCallBlock` gains a `variant` prop (`"default"` | `"subagent"`) that changes status text:
  - `subagent` calling: `"Speedwagon(Brekkylab)에 물어보는 중.."`
  - `subagent` done: `"Speedwagon(Brekkylab) 답변 완료"`
- `ToolCallFallback` in `thread.tsx` detects `ask_speedwagon_*` tool names, looks up the speedwagon name from the app store, and passes the formatted display name + `variant="subagent"` to `ToolCallBlock`

**Files changed:**
- `chat-agent/src/speedwagon/mod.rs` — added `name: String` to `KbEntry`
- `backend/src/state.rs` — populate `KbEntry.name` from `sw.name`
- `backend/src/prompt.rs` — updated test helper to include `name` field
- `frontend/components/chat/tool-call-block.tsx` — added `variant` prop
- `frontend/components/assistant-ui/thread.tsx` — speedwagon name resolution in `ToolCallFallback`

### 3. Web search as builtin tool on main agent

Adds web search as a builtin tool registered on the main `ChatAgent`, phasing out previous test/placeholder tools.

**Files changed:**
- `chat-agent/src/tools.rs` — `build_default_tool_set` now includes web search builtin
- `backend/src/agent/spec.rs` — updated agent spec assembly

### 4. ailoy dependency update

Updated ailoy to the latest revision which introduced `ToolSet::with_subagent_in_memory` and `ToolKind::Subagent`.

---

## Key design decisions

- **One agent per KB entry, built upfront.** The previous approach re-opened the search index and re-initialized an `AgentRuntime` on every tool call. The new approach pays that cost once at session start. If an index fails to open, that entry is skipped with a warning rather than failing the whole session.
- **Tool name = `ask_speedwagon_{uuid}`.** UUIDs are guaranteed to be safe as tool identifiers regardless of the speedwagon's display name.
- **Name resolution in the frontend, not the backend.** The SSE event carries the raw tool name; the frontend maps it to a display name using the already-loaded speedwagon store. This avoids adding a display-name field to the event schema.


Cargo.toml
```diff
@@ -7,4 +7,5 @@ members = [
 ]
 
 [workspace.dependencies]
-ailoy = { git = "https://github.com/brekkylab/ailoy.git", rev = "e21c7a9b437c594a071f9173f39fb0508b61ead3" }
+ailoy = { git = "https://github.com/brekkylab/ailoy.git", rev = "31b04b3eef3fd510196fb0e143bf9f7989c6ad68" }
+# ailoy = { path = "../ailoy" }
```
nuri-yoo: LGTM 

nit: 불필요한 주석으로 보입니다

nuri-yoo (approved):

Cargo.toml
khj809: @nuri-yoo 넵 사실 https://github.com/brekkylab/ailoy/pull/369 이게 먼저 merge되고 나면 rev 변경 후에 제거하려고 했습니다.

ljhh-0611 (commented):
I have some questions. Now that speedwagon subagents are no longer ephemeral runtimes, is history management handled on the ailoy side?

  If there's no separate history management, it seems like history would accumulate proportionally to sessions × speedwagons, which could become a problem. Or is the history cleared on each invocation?

  Also, with agents now resident in memory, are there any concerns about server memory overhead? And when speedwagon settings change and agents get recreated, there shouldn't be any management issues..?

khj809:
@ljhh-0611 That's a really sharp observation! I've updated ailoy (https://github.com/brekkylab/ailoy/pull/369/changes/1627f529b9ef7e7f6406706bfa34ff60f0a3fdd8) to reset the subagent's message history for each invocation.

About memory overhead, I think it's not a serious problem right now, but we can consider some solutions later, such as LRU eviction or limiting max subagent runtimes.

About subagent state management, it seems there's already a cache invalidation mechanism, so I believe there would be no particular issues.

ljhh-0611 (approved):
@khj809 Thank you for your super speed feedback!

PR #24, opened 2026-04-12
feat: Find-or-create Agent pattern with copy-on-write and  cleanup orphan agent in update/delete a session

### LINKED ISSUE #13: Apply find-or-create Agent pattern

ljhh-0611:
## Overview

Apply immutable, spec-based Agent reuse.

This PR changes Agent lifecycle from "mutate an existing Agent" to "find or create an Agent by normalized spec, then point the Session at that Agent." Identical `AgentSpec`s now converge to one DB row, model changes use copy-on-write, and orphaned Agents are cleaned up when Sessions stop referencing them.

Closes #13

## Background

Previously, every new chat or mid-session model change could create or mutate Agent rows in ways that did not match the domain model:

- `POST /agents` inserted rows even when the same spec already existed.
- `PUT /agents/{id}` mutated an Agent that could be shared by multiple Sessions.
- Model changes could accidentally affect other Sessions pointing at the same Agent.
- Agent rows could accumulate after the last Session reference disappeared.

The intended invariant is:

```text
Agent identity = normalized AgentSpec
Session owns the choice of which Agent it points to
Agent specs are immutable after creation
```

### Before

```mermaid
flowchart TD
    A["User changes model"] --> B["Frontend calls updateAgent"]
    B --> C["PUT agents by id"]
    C --> D["Backend mutates Agent spec_json"]
    D --> E["Invalidate runtimes by Agent id"]
    E --> F["Other Sessions sharing that Agent can observe the changed spec"]
```

Problems:

- Shared Agent mutation could leak model/spec changes across Sessions.
- Agent identity was row-id based, not spec based.
- Runtime invalidation needed to reason about every Session referencing that Agent.

### After

```mermaid
flowchart TD
    A["User changes model"] --> B["Load current Agent"]
    B --> C["Preserve instruction and tools; replace lm"]
    C --> D["POST agents with new spec"]
    D --> E{"Same normalized spec exists"}
    E -->|"yes"| F["Return existing Agent with 200 OK"]
    E -->|"no"| G["Insert new Agent with 201 Created"]
    F --> H["Update Session agent_id when changed"]
    G --> H
    H --> I["Invalidate only this Session runtime"]
    I --> J["Cleanup previous Agent if orphaned"]
```

This keeps model switching scoped to the active Session while still deduplicating identical Agent specs.

## Key Decisions

| Decision | Current behavior | Why |
|----------|------------------|-----|
| Agent identity | Normalized `spec_json` | Keeps scope small while making identical specs reusable. |
| Deduplication | SQLite `UNIQUE(spec_json)` + `ON CONFLICT ... RETURNING` | Makes find-or-create atomic at the DB layer. |
| Agent updates | Remove `PUT /agents/{id}` | Agents are immutable; changing model/spec means selecting another Agent. |
| Session model change | `createAgent()` then `updateSession({ agent_id })` | Keeps API responsibilities explicit. |
| Orphan cleanup | Best-effort after Session delete or agent swap | Cleanup failure should not turn a successful Session operation into a misleading client failure. |

### 1. Agent spec normalization defines identity

`normalize_spec()` canonicalizes only the fields that matter for comparing current `AgentSpec`s:

```rust
pub fn normalize_spec(spec: &AgentSpec) -> RepositoryResult<String> {
    let mut normalized = spec.clone();
    normalized.tools.sort();
    normalized.tools.dedup();
    normalized.instruction = normalized
        .instruction
        .map(|s| s.trim().to_string())
        .filter(|s| !s.is_empty());
    serde_json::to_string(&normalized).map_err(RepositoryError::from)
}
```

So these match:

```json
{ "lm": "gpt-4.1", "instruction": null, "tools": ["a", "b"] }
{ "lm": "gpt-4.1", "instruction": "",   "tools": ["b", "a"] }
```

### 2. `POST /agents` is now atomic find-or-create

SQLite now enforces uniqueness:

```sql
CREATE TABLE IF NOT EXISTS agents (
    id TEXT PRIMARY KEY,
    spec_json TEXT NOT NULL UNIQUE,
    created_at TEXT NOT NULL,
    updated_at TEXT NOT NULL
);
```

The insert path returns either the newly inserted row or the existing conflict row:

```rust
INSERT INTO agents (id, spec_json, created_at, updated_at)
VALUES (?, ?, ?, ?)
ON CONFLICT(spec_json) DO UPDATE SET updated_at = excluded.updated_at
RETURNING id, spec_json, created_at, updated_at;
```

The repository returns `(Agent, created)` by comparing the returned id to the id attempted for this insert. That avoids timestamp-based `201`/`200` ambiguity.

### 3. `PUT /agents/{id}` is removed

The route now exposes read/delete for individual Agents, but no update:

```rust
.api_route("/agents", get(list_agents).post(create_agent))
.api_route(
    "/agents/{id}",
    get(get_agent).delete(delete_agent),
)
```

This prevents direct mutation of shared Agent specs.

### 4. Session updates can relink `agent_id`

`UpdateSessionRequest` now accepts an optional `agent_id`:

```rust
pub struct UpdateSessionRequest {
    pub title: Option<String>,
    pub agent_id: Option<Uuid>,
    pub provider_profile_id: Option<Uuid>,
    pub speedwagon_ids: Option<Vec<Uuid>>,
    pub source_ids: Option<Vec<Uuid>>,
}
```

`update_session_atomic()` returns the updated Session plus the previous `agent_id` when a real swap happened:

```rust
RepositoryResult<Option<(Session, Option<Uuid>)>>
```

The service then invalidates only the changed Session runtime and cleans up the previous Agent if it is now orphaned:

```rust
state.invalidate_session_runtime(id);

if let Some(prev) = previous_agent_id {
    super::agent::cleanup_orphaned_agent(state, prev).await;
}
```

### 5. Frontend model changes preserve the rest of the Agent spec

The model selector no longer calls `updateAgent()`. It fetches the current Agent, changes only `lm`, creates/reuses an Agent, and relinks the Session if needed:

```tsx
const currentAgent = await getAgent(currentAgentId);
const agent = await createAgent({
  spec: { ...currentAgent.spec, lm: model },
});

const sessionUpdate: { agent_id?: string; provider_profile_id?: string } = {};
if (agent.id !== currentAgentId) {
  sessionUpdate.agent_id = agent.id;
}
if (profileId !== currentProfileId) {
  sessionUpdate.provider_profile_id = profileId;
}

if (Object.keys(sessionUpdate).length > 0) {
  await updateSession(sessionId, sessionUpdate);
}
```

This preserves `instruction` and `tools` across model switches.

### Backend

| File | Change |
|------|--------|
| `backend/src/repository/mod.rs` | Add `normalize_spec()`. Change `create_agent()` to return `(Agent, created)`. Add previous-agent return value to `update_session_atomic()`. |
| `backend/src/repository/sqlite.rs` | Add `UNIQUE(spec_json)` and explicit unique index. Implement atomic upsert with `ON CONFLICT ... RETURNING`. Return previous `agent_id` from session relinks. |
| `backend/src/repository/postgres.rs` | Keep matching trait stubs for the updated repository contract. |
| `backend/src/services/agent.rs` | Add `find_or_create_agent()` and orphan cleanup helper. |
| `backend/src/services/session.rs` | Validate optional `agent_id`, relink Sessions, invalidate per-session runtime, clean up previous orphan Agents. |
| `backend/src/handlers.rs` | Remove `PUT /agents/{id}`. Return `201 Created` for new Agents and `200 OK` for reused Agents. Add integration coverage. |
| `backend/src/models/agent.rs` | Remove `UpdateAgentRequest`. |
| `backend/src/models/session.rs` | Add `agent_id` to `UpdateSessionRequest`. |
| `backend/src/state.rs` | Remove agent-wide runtime invalidation path that was only needed for mutable Agents. |

### Frontend

| File | Change |
|------|--------|
| `frontend/components/chat/chat-view.tsx` | Replace model-change `updateAgent()` flow with fetch-current-Agent → `createAgent()` → `updateSession()` copy-on-write flow. |
| `frontend/lib/api.ts` | Remove `updateAgent()`. |
| `frontend/lib/types.ts` | Remove `UpdateAgentRequest`; add `agent_id?: string` to `UpdateSessionRequest`. |

### Docs

| File | Change |
|------|--------|
| `docs/conventions.md` | Add Rust formatting conventions. |

## Test Coverage

Backend integration/unit tests cover:

| Test | Verifies |
|------|----------|
| `find_or_create_same_spec_returns_same_id` | Same spec returns same Agent id; first request is `201`, reuse is `200`. |
| `find_or_create_tools_order_independent` | Tool order does not affect Agent identity. |
| `find_or_create_empty_instruction_variants_match` | `null`, empty, and whitespace-only instructions normalize together. |
| `find_or_create_different_lm_creates_new_agent` | Different model creates a distinct Agent. |
| `concurrent_create_agent_deduplicates_by_spec` | Concurrent identical creates converge to one row. |
| `update_session_agent_id_relinks_session` | `PUT /sessions/{id}` can move a Session to another Agent. |
| `delete_session_cascade_deletes_orphaned_agent` | Last Session deletion removes its orphaned Agent. |
| `delete_session_keeps_agent_when_other_sessions_exist` | Shared Agent survives when another Session still references it. |
| `update_session_agent_swap_cascade_deletes_orphaned_previous_agent` | Agent swap deletes the previous Agent when it becomes orphaned. |
| `update_session_agent_swap_keeps_previous_agent_when_other_sessions_exist` | Agent swap preserves the previous Agent when still referenced. |

Verified locally:

```text
cargo test -p agentwebui_backend -- --test-threads=1
28 passed

cd frontend && pnpm lint
exit 0
```

Note: backend tests are run serially because runtime/tool initialization can contend on shared local cache paths when run in parallel.

## Known Limitations

1. **Orphan cleanup is best-effort.** The primary Session operation succeeds even if orphan cleanup fails. A failed cleanup can leave an unused Agent row, but it does not corrupt Session references.
2. **Existing duplicate `spec_json` rows block the new unique index.** This is acceptable for the current dev-stage DB policy; deleting `backend/data/app.db` resolves old duplicate data.
3. **`normalize_spec()` must be reviewed when `ailoy::AgentSpec` changes.** The unique identity depends on deterministic serialization of the current fields.

## Architectural Note

`normalize_spec()` and JSON blob comparison are still a scoped solution. They make the current behavior correct without mixing in a larger schema refactor. A future schema could flatten Agent spec fields into SQL-native columns such as `lm`, `instruction`, and `tools_json` for clearer indexing and comparison.

This PR keeps that separate so reviewers can focus on the lifecycle behavior:

```text
Agent rows are immutable spec snapshots.
Sessions choose which Agent snapshot they use.
Identical snapshots are reused.
Unused snapshots are cleaned up when possible.
```

### DIFF
--- a/backend/src/handlers.rs
+++ b/backend/src/handlers.rs
@@ -24,6 +24,7 @@ use crate::models::{
     UpdateSpeedwagonRequest,
 };
 use crate::repository::RepositoryError;
+use crate::services::agent::{self as agent_service};
 use crate::services::session::{self as session_service, SessionError, SseEvent};
 use crate::services::speedwagon::{self as speedwagon_service, SpeedwagonError};
 use crate::state::AppState;
@@ -197,12 +198,15 @@ async fn create_agent(
     Json(payload): Json<CreateAgentRequest>,
 ) -> Result<(StatusCode, Json<AgentResponse>), (StatusCode, Json<AppError>)> {
     let CreateAgentRequest { spec } = payload;
-    let agent = state
-        .repository
-        .create_agent(spec)
+    let (agent, created) = agent_service::find_or_create_agent(&state, spec)
         .await
         .map_err(repo_err)?;
-    Ok((StatusCode::CREATED, Json(to_agent_response(&agent))))
+    let status = if created {
+        StatusCode::CREATED
+    } else {
+        StatusCode::OK
+    };
+    Ok((status, Json(to_agent_response(&agent))))
 }
 
 async fn list_agents(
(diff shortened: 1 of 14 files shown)


backend/src/repository/sqlite.rs
```diff
@@ -560,29 +570,34 @@ impl SqliteRepository {
...
+        // DO UPDATE (not DO NOTHING) is required for RETURNING to emit the conflict row.
+        // The updated_at bump on conflict is a benign side effect; agents are immutable
+        // under the copy-on-write model so updated_at has no semantic meaning.
+        let row = sqlx::query(
             r#"
             INSERT INTO agents (id, spec_json, created_at, updated_at)
-            VALUES (?, ?, ?, ?);
+            VALUES (?, ?, ?, ?)
+            ON CONFLICT(spec_json) DO UPDATE SET updated_at = excluded.updated_at
+            RETURNING id, spec_json, created_at, updated_at;
```
nuri-yoo, line 589: ### `updated_at` is being bumped on every reuse

Small thing that may be worth a second look — the `DO UPDATE SET updated_at = excluded.updated_at` here means every duplicate `POST /agents` executes a real `UPDATE` on the existing row, even though from the client's perspective nothing changed. The comment on L581 already notes that `updated_at` has no semantic meaning under CoW, so the write is purely incidental.

A couple of practical side effects:

1. Each reuse incurs a WAL write + dirty page + write-lock contention. Cheap individually, but reuse is the common path by design, so it adds up.
2. `agent.updated_at` effectively becomes a "last-touched" timestamp, which drifts from the "immutable snapshot" framing elsewhere in the PR. If a client ever keys cache off `updated_at`, it'll see invalidations that don't reflect real changes.
3. `find_or_create_same_spec_returns_same_id` only asserts `id` equality, so this behavior isn't currently pinned by a test either way.

### Suggested tweak (one line)

```sql
ON CONFLICT(spec_json) DO UPDATE SET updated_at = agents.updated_at
```

The self-assignment still triggers the conflict branch so `RETURNING` emits the existing row, which keeps the `(agent, created)` detection (`row.get("id") == id_str`) working unchanged. `DO NOTHING` would also work but needs a follow-up `SELECT` on conflict, so the self-no-op feels simpler.

backend/src/handlers.rs
```diff
@@ -1538,4 +1461,365 @@ mod tests {
...
+        let update_resp = put_json(
+            &app,
+            &format!("/sessions/{session_id}"),
+            json!({ "agent_id": agent_id2 }),
+        )
+        .await;
+        assert_eq!(update_resp.status(), StatusCode::OK);
+        let updated = response_json(update_resp).await;
+        assert_eq!(updated["agent_id"].as_str().unwrap(), agent_id2.to_string());
+    }
```
nuri-yoo: Observation — this new test asserts the DB row after the swap (`agent_id` updated in the response), but not the runtime side of the flow. The deleted `update_agent_resets_session_runtime_cache` used to cover that shape via mock-server request counts:

```text
turn-1: system(1) + streaming-user(1) = (baseline)
turn-2 after swap: system(1) + restored-history(...) + streaming-user(1) = (expected)
```

Since "invalidate only this Session runtime" is one of the mechanics called out in the PR body, a request-count test for the agent-swap case would make that invariant observable from the test layer too — today it's implicit in `update_session` calling `invalidate_session_runtime(id)` and in the cache-validity check inside `get_or_create_runtime_for_session`.

`update_provider_profile_resets_session_runtime_cache` already uses the `start_mock_chat_completion_server` + request-count pattern, so adapting it for agent-swap is mostly mechanical:

```rust
#[tokio::test]
async fn update_session_agent_id_resets_runtime_and_restores_history() {
    // 1. create agents A and B with distinct specs, point session at A
    // 2. send user message → request count after turn-1
    // 3. PUT /sessions/{id} with { agent_id: B }
    // 4. send another user message → runtime should rebuild with B and restore history
    // 5. assert the expected request-count shape (mirrors the deleted test)
}
```

backend/src/repository/sqlite.rs
ljhh-0611: I agree with your reply, so I accepted and committed your suggestion in d0574577727037a68f26313aa61b92b27a6412b7.

grf53: In this discussion, the `updated_at` field always remains the same as `created_at`. If so, is it necessary to continue maintaining the `updated_at` field?
Considering that operations like `update` might occur again in the future, would it be better to leave it as is?

backend/src/handlers.rs
ljhh-0611: I applied it in 0cb375647c08d64250d976749f205819ed5a4f0b

backend/src/repository/sqlite.rs
grf53: As discussed separately, to minimize the impact of this change and accelerate the process, we will maintain the current DB schema for now and readjust it later if necessary.

grf53 (approved):
LGTM

nuri-yoo (approved):

nuri-yoo:
Looks good to me, too

PR #30, opened 2026-04-20
test(chat-agent): fix routing tests that are missed to be modified along with #6

grf53:
## Description
Chat agent APIs to call knowledge agent are updated in #6.
But some `ignore`d tests are not fixed along with that.
This PR fixes the tests to use the latest APIs.

## Request for comments or direct fixes
The changes in this PR are written with Claude code without deep understanding of relevant codes.
So I think a confirmation by the author of #6(@ljhh-0611) is needed.

## Test
`OPENAI_API_KEY=sk-proj-... cargo test -p chat-agent -- --ignored --nocapture`


chat-agent/tests/routing_test.rs
```diff
@@ -64,36 +79,29 @@ fn create_agent(model: &str) -> ChatAgent {
...
-        .expect("tool result should be present");
+        .map(|e| e.tool.as_str())
+        .collect();
     assert!(
-        result.get("answer").is_some(),
-        "query={query:?} → tool result missing 'answer': {result}"
+        called.contains(&expected_tool.as_str()),
+        "query={query:?} → expected tool {expected_tool:?} not in calls {called:?}"
     );
 }
```
nuri-yoo, line 104: The old `assert_routes_to` also verified that the tool returned a non-empty result — might be worth restoring that coverage here too. Since `ask_speedwagon_*` returns a plain string rather than `{"answer": "..."}`, the check needs a small adaptation. Suggest replacing lines 96–106 with:

```rust
let expected_tool = format!("ask_speedwagon_{expected_kb}");
let entry = agent
    .tool_call_log()
    .iter()
    .find(|e| e.tool == expected_tool);

let entry = entry.expect(&format!(
    "query={query:?} → {expected_tool:?} was never called"
));

let result = entry.result.as_ref().expect("tool result should be present");
assert!(
    result.as_str().map(|s| !s.is_empty()).unwrap_or(false),
    "query={query:?} → tool returned empty result: {result}"
);
```

Verified locally — test passes.

grf53: Applied and check test passes too.

nuri-yoo (approved):

ljhh-0611 (approved):
I just arranged routing func a little bit :)

It works well to me too, and LGTM! Thank you

PR #34, opened 2026-04-21
fix(backend): handle ChatCompletion schema when bootstrapping openai-default profile

### LINKED ISSUE #33: backend: panic on startup when OPENAI_API_KEY is set
## Summary

`agentwebui_backend` panics on startup whenever `OPENAI_API_KEY` is present
in the environment, preventing the server from binding its listener.

```
thread 'main' panicked at backend/src/state.rs:299:26:
unsupported api schema
```

This is the most common local configuration for anyone developing against
OpenAI (`export OPENAI_API_KEY=...`), and it reproduces 100% of the time.
The server cannot be started without a workaround.

## Reproduction

```sh
export OPENAI_API_KEY=sk-...
cargo run --release -p agentwebui_backend
# → thread 'main' panicked at backend/src/state.rs:299:26: unsupported api schema
```

Workaround:

```sh
env -u OPENAI_API_KEY ./agentwebui_backend
# Then create the provider profile manually via REST:
# POST /provider-profiles { "provider": { "lm": { "schema": "chat_completion", ... } } }
```

## Root cause

`backend/src/state.rs::overwrite_default_provider_profiles_from_env` upserts
a default `provider_profile` row for each recognized API-key env var. Its
lookup table is:

```rust
let profile_definitions = [
    ("OPENAI_API_KEY",    "openai-default",    LangModelAPISchema::ChatCompletion),
    ("ANTHROPIC_API_KEY", "anthropic-default", LangModelAPISchema::Anthropic),
    ("GEMINI_API_KEY",    "gemini-default",    LangModelAPISchema::Gemini),
];
```

Notice that `OPENAI_API_KEY` is classified as `LangModelAPISchema::ChatCompletion`.

However, the `match` that converts `schema` into a `LangModelProvider`
(`state.rs:293-302`) omits the `ChatCompletion` arm and falls through to a
wildcard panic:

```rust
lm: match schema {
    LangModelAPISchema::Anthropic => LangModelProvider::anthropic(api_key),
    LangModelAPISchema::Gemini    => LangModelProvider::gemini(api_key),
    LangModelAPISchema::OpenAI    => LangModelProvider::openai(api_key),
    _ => panic!("unsupported api schema"),
},
```

### Why the variant names mislead

`LangModelAPISchema` has four variants (`ailoy/src/agent/spec.rs:53-68`):

| variant | wire protocol | what OpenAI users typically want |
|---|---|---|
| `ChatCompletion` | `POST /v1/chat/completions` (and OpenAI-compatible servers: vLLM, Azure, Together, Groq, …) | **Yes — the mainstream OpenAI endpoint** |
| `Anthropic`      | Anthropic Messages API | — |
| `Gemini`         | Google Gemini API | — |
| `OpenAI`         | `POST /v1/responses` (Responses API) | Only if they explicitly want Responses API |

The variant literally named `OpenAI` does **not** cover OpenAI's standard
chat endpoint — it targets the newer Responses API. Likewise,
`LangModelProvider::openai(api_key)` hard-codes `/v1/responses`
(`ailoy/src/agent/rt/lang_model/api/openai.rs:15`), whereas the standard
endpoint is built by `LangModelProvider::chat_completion(url, Some(key))`
(`ailoy/src/agent/rt/lang_model/api/chat_completion.rs:23`).

So the table correctly routes `OPENAI_API_KEY → ChatCompletion`, but the
match forgot to handle that arm — the wildcard catches it and panics.

### Why the compiler didn't catch this

The `_ => panic!(...)` wildcard suppresses Rust's exhaustiveness check.
Without it, omitting `ChatCompletion` would have surfaced as a
`non-exhaustive patterns` compile error.

### Why Anthropic/Gemini don't hit the same bug

Their table entries and match arms share the same variant name
(`Anthropic` ↔ `Anthropic`, `Gemini` ↔ `Gemini`), so they line up by
accident. Only OpenAI has the table/arm mismatch (`ChatCompletion` vs
`OpenAI`), and that is the single input guaranteed to fall through.

## Proposed fix

Add the missing arm and drop the wildcard so future additions to
`LangModelAPISchema` surface as compile errors:

```rust
let provider = AgentProvider {
    lm: match schema {
        LangModelAPISchema::ChatCompletion => LangModelProvider::chat_completion(
            "https://api.openai.com/v1/chat/completions",
            Some(api_key),
        )
        .map_err(|e| RepositoryError::InvalidData(format!(
            "invalid OpenAI chat_completion URL: {e}"
        )))?,
        LangModelAPISchema::Anthropic => LangModelProvider::anthropic(api_key),
        LangModelAPISchema::Gemini    => LangModelProvider::gemini(api_key),
        LangModelAPISchema::OpenAI    => LangModelProvider::openai(api_key),
    },
    tools: vec![],
};
```

Rationale for keeping `OPENAI_API_KEY → ChatCompletion` (rather than
remapping it to `OpenAI`):

- Preserves the mainstream OpenAI wire protocol that existing users expect.
- Avoids migrating any `openai-default` rows already written to existing
  SQLite databases, which would silently switch live sessions to the
  Responses API format.

## Verification

- [ ] `OPENAI_API_KEY=sk-... ./agentwebui_backend` boots without panic
- [ ] `GET /provider-profiles` returns `openai-default` with
      `lm.schema == "chat_completion"`
- [ ] Existing rows in SQLite (`openai-default` with `ChatCompletion`) are
      unchanged after restart
- [ ] (Optional) Add an integration test that boots `AppState` with each
      of `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GEMINI_API_KEY` set and
      asserts no panic

## Scope

- Single-file change: `backend/src/state.rs`
- No changes to the public REST API, DB schema, or agent/tool behavior

nuri-yoo:
## Summary

`agentwebui_backend` panics on startup when `OPENAI_API_KEY` is set, because `overwrite_default_provider_profiles_from_env` routed `OPENAI_API_KEY` to `LangModelAPISchema::ChatCompletion` but the match that converts `schema` into a `LangModelProvider` omitted that arm and fell through to `_ => panic!("unsupported api schema")`.

## Fix

- Route `OPENAI_API_KEY` to `LangModelAPISchema::OpenAI` (`/v1/responses`) instead of `ChatCompletion`.
- Drop the `_ => panic!(...)` wildcard. The match now handles every `LangModelAPISchema` variant explicitly, so future additions surface as compile errors instead of runtime panics.

## Verification

- [x] `OPENAI_API_KEY=sk-... ./agentwebui_backend` boots without panic (`server listening on http://127.0.0.1:...`)
- [x] `GET /provider-profiles` returns `openai-default` with `lm.schema == "openai"` and `lm.url == "https://api.openai.com/v1/responses"`
- [x] `cargo check -p agentwebui_backend` clean
- [x] `cargo test -p agentwebui_backend` — 19 passed, 0 failed

## Scope

Single-file change: `backend/src/state.rs`. No changes to public REST API, DB schema, or agent/tool behavior.

Closes #33

### DIFF
--- a/backend/src/state.rs
+++ b/backend/src/state.rs
@@ -287,10 +287,18 @@ impl AppState {
 
             let provider = AgentProvider {
                 lm: match schema {
+                    LangModelAPISchema::ChatCompletion => LangModelProvider::chat_completion(
+                        "https://api.openai.com/v1/chat/completions",
+                        Some(api_key),
+                    )
+                    .map_err(|e| {
+                        RepositoryError::InvalidData(format!(
+                            "invalid OpenAI chat_completion URL: {e}"
+                        ))
+                    })?,
                     LangModelAPISchema::Anthropic => LangModelProvider::anthropic(api_key),
                     LangModelAPISchema::Gemini => LangModelProvider::gemini(api_key),
                     LangModelAPISchema::OpenAI => LangModelProvider::openai(api_key),
-                    _ => panic!("unsupported api schema"),
                 },
                 tools: vec![],
             };
--- a/backend/src/state.rs
+++ b/backend/src/state.rs
@@ -263,7 +263,7 @@ impl AppState {
             (
                 "OPENAI_API_KEY",
                 "openai-default",
-                LangModelAPISchema::ChatCompletion,
+                LangModelAPISchema::OpenAI,
             ),
             (
                 "ANTHROPIC_API_KEY",
@@ -287,18 +287,15 @@ impl AppState {
 
             let provider = AgentProvider {
                 lm: match schema {
-                    LangModelAPISchema::ChatCompletion => LangModelProvider::chat_completion(
-                        "https://api.openai.com/v1/chat/completions",
-                        Some(api_key),
-                    )
-                    .map_err(|e| {
-                        RepositoryError::InvalidData(format!(
-                            "invalid OpenAI chat_completion URL: {e}"
-                        ))
-                    })?,
+                    LangModelAPISchema::OpenAI => LangModelProvider::openai(api_key),
                     LangModelAPISchema::Anthropic => LangModelProvider::anthropic(api_key),
                     LangModelAPISchema::Gemini => LangModelProvider::gemini(api_key),
-                    LangModelAPISchema::OpenAI => LangModelProvider::openai(api_key),
+                    LangModelAPISchema::ChatCompletion => {
+                        return Err(RepositoryError::InvalidData(format!(
+                            "`{name}` cannot be bootstrapped with ChatCompletion schema; \
+                             create the provider profile via POST /provider-profiles"
+                        )));
+                    }
                 },
                 tools: vec![],
             };


nuri-yoo:
@grf53 @ljhh-0611

grf53:
> Keeping `OPENAI_API_KEY → ChatCompletion` (rather than remapping it to `OpenAI`, which targets the newer Responses API) preserves the wire protocol existing users and SQLite rows expect.

Could you help me to understand this sentence more precisely?
At a first glance, remapping to `OpenAI` with Responses API would be a better solution we now have the support for that.

I cannot clearly recall a situation where "expected aspects are not met." by using `OpenAI` instead of `ChatCompletion`. Is there any information that is not obtained when using the Responses API compared to using the ChatCompletion API?

nuri-yoo:
The two serve different purposes.

`ChatCompletion` is the de-facto industry standard — OpenAI's 2023 format that every OpenAI-compatible provider (vLLM, Azure, Together, Groq, …) mirrors. `Responses` is OpenAI's 2025 successor for agentic/reasoning workflows: typed content blocks (`output_text` + `reasoning` + `function_call`), optional server-side state via `previous_response_id`, and native reasoning-summary round-trips. OpenAI-only, not mirrored by third parties.

To your question directly: no, nothing is lost by going through `OpenAI` (Responses) — at the level ailoy actually uses, Responses is a strict superset. The only thing `ChatCompletion` uniquely offers is ecosystem reach (OpenAI-compatible servers).

So I agree with your suggestion: remap `OPENAI_API_KEY → OpenAI`. If OpenAI-compatible endpoints ever need to share the default profile, we can revisit.

One note regardless of direction: dropping the `_ => panic!` wildcard so the compiler enforces exhaustive matching would prevent this class of bug from recurring.

grf53 (approved):

ljhh-0611 (approved):

PR #36, opened 2026-04-22
fix: tool call tracking in live stream and on reload

### LINKED ISSUE #31: bug: Session tool-call history is out of order on reload
## Title

Tool call tracking issues: reversed pairing, unstable ordering, and dropped entries

---

## Summary

When multiple tool calls are executed within a single turn, three separate issues across backend and frontend layers cause incorrect pairing, unstable ordering, and missing tool call entries in the UI.

---

## Bug 1 — Reverse pairing of tool call and result

**Where**

* `backend/src/services/session.rs:253-259`
* `chat-agent/src/chat_agent.rs:288-296`

**Behavior**
When the same tool is called multiple times in a single turn, arguments and results are paired in reverse order.
The result from Call #1 is attached to Call #N, and vice versa.

**Observed symptom**

```
Args(start=240, end=365) ↔ Result(start=80, end=160)
Args(start=80, end=160) ↔ Result(start=240, end=365)
```

**Root cause**
The following logic scans in reverse order:

```rust
.iter_mut().rev().find(|(n, _, r)| n == &tool && r.is_none())
```

However, ailoy emits tool results in FIFO order, so matching should also proceed forward.

**Fix**
Remove `.rev()` so matching happens from the oldest pending entry:

```rust
.iter_mut().find(|(n, _, r)| n == &tool && r.is_none())
```

The same fix is required in `chat_agent.rs` for streaming `tool_name` lookup.

---

## Bug 2 — Tool call order scrambled on session reload

**Where**

* `backend/src/services/session.rs:305-318`

**Behavior**
Tool calls appear in correct order during live streaming, but after reloading the session, the order becomes inconsistent (e.g., `open_document` appearing before `convert_pdf_to_md`).

**Root cause**

* All `SessionToolCall` entries in a turn share the same `chrono::Utc::now()` timestamp
* SQLite stores timestamps with millisecond precision
* Query:

  ```sql
  ORDER BY created_at ASC, id ASC
  ```

  falls back to UUID ordering when timestamps are identical

**Fix**
Assign a base timestamp per turn and offset each entry by its index:

```rust
let base_ts = chrono::Utc::now();

tool_calls_for_db
    .iter()
    .enumerate()
    .map(|(i, (name, args, result))| SessionToolCall {
        ...
        created_at: base_ts + chrono::Duration::milliseconds(i as i64),
    })
```

**Note**
Maximum skew is `(tool_call_count - 1)` ms, which is negligible and requires no schema change.

---

## Bug 3 — Live UI drops tool calls within the same millisecond

**Where**

* `frontend/components/chat/api-runtime-provider.tsx:67`

**Behavior**
The live UI shows fewer tool calls than actually persisted.
Example: live view shows 2 entries, but reloading shows 5.

**Root cause**

```ts
const callId = `tc_${Date.now()}_${data.tool ?? "unknown"}`;
```

* `Date.now()` has millisecond precision
* Multiple SSE events within the same ms generate identical IDs
* `Map.set()` overwrites previous entries silently

**Fix**
Add a monotonic counter to guarantee unique IDs:

```ts
let callCounter = 0;

const callId = `tc_${Date.now()}_${callCounter++}_${data.tool ?? "unknown"}`;
```

---

## Relate issue
#26 

### Related Image

Image

Image

selloriwoo:
## Ticket
resolves #31

## Description
- Fix issue where consecutive tool calls were misordered or missing in the UI.

- When multiple tool calls (e.g., `open tool`) were triggered within the same millisecond during an `LLM chat`, some calls were not displayed in the live UI, and their order appeared inconsistent after a session reload.

- This change ensures that tool calls are consistently recorded and displayed in the correct order, even when invoked in rapid succession.

## Bugs
- Tool call order scrambled on session reload `SessionToolCall` rows within one turn were being stamped with the same `chrono::Utc::now()` tick (millisecond precision in SQLite), so `ORDER BY created_at ASC, id ASC` fell back to random UUID order.

- Live UI drops tool calls within the same ms `callId = tc_${Date.now()}_${tool}` collided whenever multiple `tool_call` SSE events arrived in the same millisecond,
and `Map.set(callId, ...)` silently overwrote earlier entries. DB still persisted all calls, so the missing ones reappeared on reload
### fix 
- capture one `base_ts` per turn and offset each row by its index in milliseconds — distinct timestamps, insertion order preserved. (`backend/src/services/session.rs` -> `.enumerate()` + `base_ts + i ms`)

- in `createApiAdapter` add a monotonic counter to the callId, guaranteeing uniqueness regardless of clock resolution.

## Compatibility notes
**Backend fix**
- changes only the `created_at` values written by one code path. Existing save/read SQL, API contract, and DB schema are unchanged. 
Entries already in the DB are not migrated. only newly persisted tool calls get the staggered timestamps.

-- before
```
INSERT INTO session_tool_calls VALUES (..., '2026-04-20T14:00:12.200Z');  -- call 1
INSERT INTO session_tool_calls VALUES (..., '2026-04-20T14:00:12.200Z');  -- call 2
INSERT INTO session_tool_calls VALUES (..., '2026-04-20T14:00:12.200Z');  -- call 3
```

-- after
```
INSERT INTO session_tool_calls VALUES (..., '2026-04-20T14:00:12.200Z');  -- call 1 (base+0ms)
INSERT INTO session_tool_calls VALUES (..., '2026-04-20T14:00:12.201Z');  -- call 2 (base+1ms)
INSERT INTO session_tool_calls VALUES (..., '2026-04-20T14:00:12.202Z');  -- call 3 (base+2ms)
```
-> The per-entry offset stays within (N−1) ms per turn — harmless for ordering.
If future features need the true wall-clock time, a separate `sequence` column could be a cleaner fix.

**Frontend fix**
- `callId` is an ephemeral, in-memory Map key used only while a single turn is streaming. It is not sent over the network, not persisted, and gone as soon as the page unmounts or reloads. Reload uses the server-generated `SessionToolCall.id`, which this PR does not touch.

### DIFF
--- a/backend/src/repository/sqlite.rs
+++ b/backend/src/repository/sqlite.rs
@@ -1522,7 +1522,7 @@ impl Repository for SqliteRepository {
             SELECT id, message_id, tool_name, tool_args, tool_result, duration_ms, created_at
             FROM session_tool_calls
             WHERE message_id = ?
-            ORDER BY created_at ASC, id ASC;
+            ORDER BY created_at ASC;
             "#,
         )
         .bind(message_id)
@@ -1568,7 +1568,7 @@ impl Repository for SqliteRepository {
             WHERE tc.message_id IN (
                 SELECT id FROM session_messages WHERE session_id = ?
             )
-            ORDER BY tc.created_at ASC, tc.id ASC;
+            ORDER BY tc.created_at ASC;
             "#,
         )
         .bind(session_id.to_string())
--- a/backend/src/services/session.rs
+++ b/backend/src/services/session.rs
@@ -301,17 +301,22 @@ pub async fn send_message_streaming(
                 {
                     Ok(Some(msg)) => {
                         // Save tool calls
+                        // Stagger created_at by index so same-turn rows sort deterministically
+                        // (DB stores ms precision; without offsets ORDER BY falls back to UUID).
                         if !tool_calls_for_db.is_empty() {
+                            let base_ts = chrono::Utc::now();
                             let tool_call_models: Vec<SessionToolCall> = tool_calls_for_db
                                 .iter()
-                                .map(|(name, args, result)| SessionToolCall {
+                                .enumerate()
+                                .map(|(i, (name, args, result))| SessionToolCall {
                                     id: uuid::Uuid::new_v4().to_string(),
                                     message_id: msg.id.clone(),
                                     tool_name: name.clone(),
                                     tool_args: args.clone(),
                                     tool_result: result.clone(),
                                     duration_ms: None,
-                                    created_at: chrono::Utc::now(),
+                                    created_at: base_ts
+                                        + chrono::Duration::milliseconds(i as i64),
                                 })
                                 .collect();
                             let _ = repository.save_tool_calls(&msg.id, &tool_call_models).await;
--- a/frontend/components/chat/api-runtime-provider.tsx
+++ b/frontend/components/chat/api-runtime-provider.tsx
@@ -41,6 +41,8 @@ function createApiAdapter(sessionId: string): ChatModelAdapter {
 
       // Track tool calls across events keyed by insertion order
       const toolCalls: Map<string, { toolName: string; args: unknown; result?: unknown }> = new Map();
+      // Counter prevents callId collisions within the same ms (Date.now() alone collides).
+      let callCounter = 0;
 
       // Build the tool-call content parts array from current state
       function buildToolParts() {
@@ -64,7 +66,7 @@ function createApiAdapter(sessionId: string): ChatModelAdapter {
               };
               break;
             case "tool_call": {
-              const callId = `tc_${Date.now()}_${data.tool ?? "unknown"}`;
+              const callId = `tc_${Date.now()}_${callCounter++}_${data.tool ?? "unknown"}`;
               toolCalls.set(callId, { toolName: data.tool ?? "", args: data.args });
               yield {
                 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- assistant-ui lacks public tool-call content-part type


ljhh-0611 (approved):
🫡

grf53:
@ljhh-0611 
Currently, because proper sorting cannot be achieved using only the information existing in the session tool call rows, do we  need to inject the 'order' manually?

ljhh-0611:
> @ljhh-0611 Currently, because proper sorting cannot be achieved using only the information existing in the session tool call rows, do we need to inject the 'order' manually?

In this step, I think that we dont need to get order so that appropriately, I only managed to get that much done.
I just approved it because I dont get any reason not to apply this PR-
Did you say like this becuase you think it is too much engineering thing?

grf53:
> so ORDER BY created_at ASC, id ASC fell back to random UUID order.

I thought that the point we need to modify is here.

It would be best if proper sorting were possible using only the existing information, but if not, order injection like the current change will be necessary, and then accordingly, sorting by ID don't need to be done here.

ljhh-0611:
> > so ORDER BY created_at ASC, id ASC fell back to random UUID order.
> 
> I thought that the point we need to modify is here.
> 
> It would be best if proper sorting were possible using only the existing information, but if not, order injection like the current change will be necessary, and then accordingly, sorting by ID don't need to be done here.

Indeed, we should also figure out whether using autoincrement instead of random UUID for the ID would be sufficient to maintain ordering.

grf53:
The only one thing I want to confirm was: 'Is it true that proper sorting is impossible based on the information currently held in the rows?'
And if so, I agree that the current change is sufficiently necessary, and the suggestion was to remove sorting by the id field there, which seems to be no longer necessary.

It wasn't any critic or sth for the ID schemes.

ljhh-0611:
> The only one thing I want to confirm was: 'Is it true that proper sorting is impossible based on the information currently held in the rows?' And if so, I agree that the current change is sufficiently necessary, and the suggestion was to remove sorting by the id field there, which seems to be no longer necessary.
> 
> It wasn't any critic or sth for the ID schemes.

Ah, I just rethinked with your opinion! It is right, in case of almost simultaneously call, in current situation, this unordered things can be occured and we couldnt sort properly again.

Created time can be solution about this, but I think that auto-increment id system will be more core solution in spite of DB restructuring

-> UPDATED: Just Do this PR. It is not that much important process. it will be enough to just ordering calls.

ljhh-0611 (approved):

grf53:
@selloriwoo 
Could you check we can modify `ORDER BY created_at ASC, id ASC` into `ORDER BY created_at ASC` in the `get_tool_calls_for_message` and `get_tool_calls_for_session`?
If there is no other issue, then we could exclude `id` from sorting tool calls.

grf53 (approved):
Thank you!

PR #39, opened 2026-04-24
feat: add convert_pdf_to_md by pyinstaller and shell

### LINKED ISSUE #38: Apply PyInstaller on Docling execution environment
## Description
Bundle the docling execution environment and a Python program that converts PDF files to Markdown into onefile (or onedir) using PyInstaller.

## To-do
1. docling을 돌릴 수 있는 uv 기반의 venv 실행 환경 생성
    1. pyproject.toml(+ 필요시 uv.lock)으로서 코드화
3. docling을 사용해서 pdf binary를 받아서(stdin) markdown text을 출력(stdout)하는 프로그램 작성
4. pyinstaller로 해당 프로그램(+venv 실행 환경)을 실행할 수 있는 onefile(불가피한경우 onedir) binary 생성
5. 3을 간편하게 돌려주는 bash 스크립트 작성

## Expected artifacts
- `pyproject.toml`
- (`uv.lock`)
- `run.py` (docling 실행 프로그램)
- `build.bash` (pyinstaller 실행 스크립트)

selloriwoo:
## Ticket
Resolves #38

## Description
Bundle the docling execution environment and a Python program that converts PDF files to Markdown into `onedir` using PyInstaller.

## Remaining Tasks
- [x] Check compatibility on other environments


convert-pdf-to-md/README.md
```diff
@@ -0,0 +1,52 @@
...
+```bash
+uv sync --dev
+```
+
+### CPU-Only version
+To install a CPU-only torch wheel excluding NVIDIA dependencies for environments without a GPU or for reducing PyInstaller binary size, use the following command.
+
+```bash
+uv sync --extra cpu
+```
```
nuri-yoo, line 17: This only shows `uv sync --extra cpu`, but `build.sh` invokes `.venv/bin/pyinstaller`, and pyinstaller is in `[dependency-groups].dev` — not in the `cpu` extra. Following this section verbatim leaves the venv without pyinstaller and the build fails.

```suggestion
uv sync --dev --extra cpu
```

Might also be worth a one-line note that the `dev` group is required whenever you want to run `build.sh`.

nuri-yoo (commented):
The overall structure looks reasonable. Left specific suggestions inline.
Two things to flag at the file level:

- The README CPU-only path and `build.sh:1` both look like leftovers from the `run.py` → `convert_pdf_to_md.py` rename.
- `validate_pdf()` is stricter than what real-world PDFs and major readers accept — see inline.

**Optional**

- Adding `set -euo pipefail` and a shebang to `build.sh` would make partial build failures easier to spot.
- `convert-pdf-to-md/.gitignore` could pick up `__pycache__/` and `.venv/` — the root `.gitignore` is Cargo/Node-focused and doesn't cover them.

convert-pdf-to-md/convert_pdf_to_md.py
```diff
@@ -0,0 +1,73 @@
...
+    return build_converter().convert(stream).document.export_to_markdown()
+
+
+def validate_pdf(data: bytes) -> None:
+    if not data:
+        sys.exit("error: empty input on stdin")
+    if not data.startswith(b"%PDF-"):
+        sys.exit(f"error: not a PDF (header={data[:8]!r})")
+    if b"%%EOF" not in data[-1024:]:
+        sys.exit("error: PDF EOF marker not found")
```
nuri-yoo, line 57: `startswith(b"%PDF-")` is stricter than what real-world PDFs and major readers accept. ISO 32000 doesn't require the header at byte 0 (PDF/A does, but that's a separate profile), and Acrobat / pikepdf / PyPDF / MuPDF all look for `%PDF-` anywhere in the first 1024 bytes — this is the long-standing tolerance for PDFs that pick up extra bytes from mail gateways or other tooling.

```suggestion
    if b"%PDF-" not in data[:1024]:
        sys.exit(f"error: not a PDF (no %PDF- in first 1024 bytes; head={data[:16]!r})")
    if b"%%EOF" not in data[-1024:]:
        sys.exit("error: PDF EOF marker not found")
```

The `%%EOF` check at `data[-1024:]` is fine — that matches what every major reader actually does as a first pass.

convert-pdf-to-md/build.sh
```diff
@@ -0,0 +1,16 @@
+rm -f run.spec
```
nuri-yoo: PyInstaller derives the spec filename from the entry script, so this build generates `convert_pdf_to_md.spec`, not `run.spec`. Looks like a leftover from the `run.py` → `convert_pdf_to_md.py` rename in 8108c83. Either drop the line or:

```suggestion
rm -f convert_pdf_to_md.spec
```

convert-pdf-to-md/convert_pdf_to_md.py
```diff
@@ -0,0 +1,73 @@
+import sys
+from io import BytesIO
+import json
+import logging
+import os
+from pathlib import Path
```
nuri-yoo: `json` and `pathlib.Path` aren't referenced anywhere in this file (verified via AST). Likely leftover from before the script was simplified — safe to drop.

```suggestion
import sys
from io import BytesIO
import logging
import os
```

convert-pdf-to-md/convert_pdf_to_md.py
```diff
@@ -0,0 +1,73 @@
...
+
+def main():
+    data = sys.stdin.buffer.read()
+    validate_pdf(data)
+    markdown = run_docling(BytesIO(data))
+    print(markdown)
+
+
+if __name__ == "__main__":
+    # to prevent chile process to run entry point again
```
nuri-yoo: nit: `chile` → `child`. The comment itself isn't really pulling its weight (`multiprocessing.freeze_support()` is the standard PyInstaller idiom), so deleting the line is also fine.

convert-pdf-to-md/convert_pdf_to_md.py
nuri-yoo: Reference for the 1024-byte tolerance — Adobe PDF Reference 1.4, p.790 (Implementation Notes §3.4.1, "File Header"):

> 12. 13. Acrobat viewers require only that the header appear somewhere within the first 1024 bytes of the file.
> Acrobat viewers will also accept a header of the form `%!PS−Adobe−N.n PDF−M.m`

Source: https://opensource.adobe.com/dc-acrobat-sdk-docs/pdfstandards/pdfreference1.4.pdf

ISO 32000 §7.5.2 itself just says "first line" without a byte offset. Byte 0 is enforced by PDF/A (e.g. veraPDF PDF/A-1 rule: "The % character of the file header shall occur at byte offset 0 of the file"), but that's the PDF/A profile, not general PDF.

convert-pdf-to-md/build.sh
selloriwoo: I confirmed that even if we remove that line, running build.sh overwrites convert_pdf_to_md.spec with a newly generated file. The file’s creation timestamp appears older, but its contents and modification time are updated. Since the creation time isn’t important in this context, I’ll go ahead and remove the line.

nuri-yoo: Thanks for confirming.

convert-pdf-to-md/README.md
selloriwoo: I tested this locally — uv sync --extra cpu still installs pyinstaller since uv includes the dev dependency group by default (unless --no-dev is used). ./build.sh also runs fine after that.

That said, I’ll add --dev explicitly to the command to make the behavior consistent across different environments (e.g., potential version differences or OS-specific setups) rather than relying on defaults.


-test-
`uv sync --extra cpu --no-dev`
Resolved 116 packages in 1ms
Uninstalled 3 packages in 23ms
 - -altgraph==0.17.5
 - -pyinstaller==6.20.0
 - -pyinstaller-hooks-contrib==2026.4

`uv sync --extra cpu`
Resolved 116 packages in 1ms
Installed 3 packages in 29ms
 + +altgraph==0.17.5
 + +pyinstaller==6.20.0
 + +pyinstaller-hooks-contrib==2026.4

`uv sync --dev --extra cpu`
Resolved 116 packages in 1ms
Installed 3 packages in 26ms
 + +altgraph==0.17.5
 + +pyinstaller==6.20.0
 + +pyinstaller-hooks-contrib==2026.4

nuri-yoo: Thanks for the correction. Explicit `--dev` for clarity sounds right.

convert-pdf-to-md/convert_pdf_to_md.py
selloriwoo: I confirmed that these imports are no longer used in this file. I'll remove them.

convert-pdf-to-md/convert_pdf_to_md.py
selloriwoo: Thanks for the suggestion. I checked, and multiprocessing.freeze_support() is indeed a common PyInstaller pattern, so I'll remove the comment.

convert-pdf-to-md/convert_pdf_to_md.py
selloriwoo: Thanks for the clarification. I checked a few libraries and references, and it looks like allowing %PDF- within the first few KB is standard practice. I’ll update this.

nuri-yoo (approved):
LGTM.

grf53 (approved):
:+1:

PR #40, opened 2026-04-26
find: line-based matching with bare-word fallback and query syntax

### LINKED ISSUE #44: Improve query grammar for `find` tool of speedwagon
## Description
Based on discussions following the announcement of Query Relaxation, apply the existing phased Query Relaxation, but redefine the word position expression methods and the syntax of the Query language into a more general format

## Reference
- Slide: [Query Relaxation (0424)](https://docs.google.com/presentation/d/1pyNesnJacA7qKjj9xZ4C3An0yNhoNLR1-6vfql7iJKc/edit?usp=sharing)

nuri-yoo:
### Internal: line/col matching, external: byte offsets

`find_in_document` now splits the document into lines (`compute_line_spans` records `(byte_start, byte_end_exclusive)` for each line so we can translate back to absolute offsets later). All match decisions — clause evaluation, AND/HALF/OR thresholding, one-match-per-line collapse — happen on lines.

The public `FindMatch` keeps the existing `start: usize` / `end: usize` byte offsets and the `context: String` byte-window (defaults to ±256 bytes around the match), so the tool's I/O contract is unchanged. The conversion happens at the very end: `start_byte = line_byte_start + col_offset_in_line`.

This satisfies the "internally line/col, externally byte" requirement: the matching algorithm and fallback live in line space, while the tool's JSON response keeps the same byte-offset interface it had under the simple-regex implementation.

### Bare-word fallback (AND → HALF → OR)

When the input is purely whitespace-separated bare words — no quotes, no `+`/`-`, no `AND`/`OR`/`NOT`, no parens, no `/.../` — we apply a progressive threshold:

1. require ALL keywords on the same line; if no line qualifies,
2. require at least HALF (rounded up); if still none,
3. require ANY single keyword.

The applied level is reported in `FindResult.relaxation` as `"all" | "half" | "any" | null`. `null` means either no fallback was needed (the strict match already returned hits, or there's only a single keyword so "all" is meaningless to report) or the query used explicit operators (which opt out of fallback — see below).

This is meant for LLM tool callers in particular: if a model issues a multi-keyword search and the response comes back with `relaxation: "any"`, the model can tell the result set was loosened and decide whether to refine the query, ask for more context, or trust the fallback.

### Query syntax (structured-query subset)

Replaces the single-regex input with a parser supporting:

| Syntax              | Meaning                                                   |
| ------------------- | --------------------------------------------------------- |
| `term`              | literal substring (case-insensitive)                      |
| `"phrase"`          | exact phrase, treated as a single literal                 |
| `+term`             | required (must appear in the matched line)                |
| `-term`             | excluded                                                  |
| `term1 AND term2`   | conjunction (operator is case-sensitive, must be `AND`)   |
| `term1 OR term2`    | disjunction                                               |
| `NOT term`          | negation (same as `-term`)                                |
| `(group)`           | grouping for precedence                                   |
| `/regex/`           | regex literal, case-insensitive                           |

Any explicit use of `"`, `+`, `-`, `(`, `)`, `/`, or the operators `AND`/`OR`/`NOT` switches the query out of bare-word mode and disables fallback. The query is then evaluated as written. This preserves caller intent: if the user wrote `alpha AND beta`, the tool should return zero results when no line has both — never quietly relax to "any line with alpha or beta."

The parser is a small recursive-descent (`or → and → unary → atom`) that produces an `Expr` AST (`Literal`, `Regex`, `And`, `Or`, `Not`). Adjacent terms with no connector default to AND. `+term` is a no-op semantically (its surrounding AND group already requires it) — it's accepted because it signals intent and disables fallback. A pure-negative AND (e.g. just `-foo`) intentionally returns nothing, since there's no positive evidence to anchor a match on.

Field selectors (`field:value`), boosts (`^N`), fuzziness (`~`), and proximity searches are intentionally not supported. They aren't needed for single-document grep, and adding them would push us toward a much larger parser surface.

### Reported anchor per match

Each matching line emits one `FindMatch`. The anchor (the keyword and position reported) is the earliest positive-evidence term hit in the line. For `Not` branches there is no anchor — they only filter lines, never surface as the matched text. For literals we keep the original-case spelling as the `keyword` field; for regexes we keep the source pattern.

### `relaxation` field on the response

`FindResult` (and the JSON returned by the tool) gains a `relaxation` key:

```json
{
  "next_cursor": null,
  "relaxation": "half",
  "matches": [...]
}
```

Values: `"all"` (multi-keyword strict match), `"half"`, `"any"`, or `null` (no fallback applied or single-keyword query). The tool description was updated so an LLM caller sees what to expect.

## Files touched

- `speedwagon/src/store/document.rs` — full rewrite of `find_in_document` plus tokenizer/parser/AST/evaluator. `Document`, `read_in_document`, the byte-boundary helpers, and `FindMatch` shape are unchanged. `FindResult` gains the `relaxation` field.
- `speedwagon/src/tool/find.rs` — surfaces `relaxation` in the JSON response, and updates the tool description and parameter docs to describe the new query syntax (and the fallback semantics) so an LLM caller knows how to use it.

No changes to `Store::find` signature, no new modules, no new dependencies (uses the existing `regex` crate).

## Tests

`cargo test --lib` on speedwagon: **29 passed, 1 ignored** (the ignored one requires network + docling — unrelated).

The 23 tests in `store::document::tests` cover:

- single-keyword match + byte-offset round-trip
- `"phrase"` quoted phrases (incl. content with regex meta chars)
- `+required` / `-excluded`
- `AND` (must require both — no fallback)
- `OR` operator
- `NOT` (excludes lines)
- `/regex/` literal + case-insensitivity through regex
- `(group)` with `OR` and `+required`
- bare-word fallback at each level: ALL match, HALF match, ANY match (3-keyword case where `half` is unreachable so we fall through to `any`), 2-keyword case where `half == any` (we report `half`, the level we actually used)
- single-keyword queries report `relaxation: null` (nothing to relax)
- explicit `AND` does NOT trigger fallback
- one-match-per-line collapse when a keyword repeats on the same line
- cursor pagination + cursor exhaustion
- UTF-8 boundary safety (Korean text + tight `context_bytes`)
- empty/whitespace-only patterns return no matches without panicking

### DIFF
--- a/speedwagon/src/store/document.rs
+++ b/speedwagon/src/store/document.rs
@@ -37,6 +37,11 @@ impl fmt::Display for Document {
 pub struct FindResult {
     pub id: String,
     pub next_cursor: Option<usize>,
+    /// `Some("all"|"half"|"any")` when bare-word fallback was applied to
+    /// produce these matches; `None` otherwise (no relaxation, or the query
+    /// used explicit operators). Lets the caller (typically an LLM) know
+    /// the match set was loosened from a strict AND-of-keywords intent.
+    pub relaxation: Option<&'static str>,
     pub matches: Vec<FindMatch>,
 }
 
@@ -48,6 +53,30 @@ pub struct FindMatch {
     pub context: String,
 }
 
+/// Search a document for occurrences of `pattern`. Matching is case-insensitive
+/// and line-oriented: at most one match is reported per matching line.
+///
+/// Query syntax (subset of structured query syntax):
+/// - `term`              bare word
+/// - `"phrase"`          exact phrase (single literal)
+/// - `+term` / `-term`   required / excluded
+/// - `term1 AND term2`   conjunction (operators are case-sensitive)
+/// - `term1 OR term2`    disjunction
+/// - `NOT term`          negation (same as `-term`)
+/// - `(group)`           grouping
+/// - `/regex/`           regex literal (case-insensitive)
+///
+/// When the input is purely bare words (no operators at all), a progressive
+/// fallback is applied:
+/// 1. require ALL keywords on the same line; if no hits,
+/// 2. require at least HALF; if still no hits,
+/// 3. require ANY single keyword.
+///
+/// Any explicit operator opts out of fallback so caller intent is preserved.
+///
+/// Internally we work in line/column space (matching is line-oriented and
+/// the fallback threshold is computed per line); only byte offsets are
+/// surfaced in the returned `FindMatch`.
 pub fn find_in_document(
     id: &str,
     content: &str,
@@ -56,48 +85,62 @@ pub fn find_in_document(
     k: usize,
     context_bytes: usize,
 ) -> FindResult {
-    if pattern.is_empty() {
-        return FindResult {
-            id: id.to_string(),
-            next_cursor: None,
-            matches: vec![],
-        };
-    }
-
-    // Always match case-insensitively; (?i) can be overridden inline via (?-i) if needed
-    let Ok(re) = Regex::new(&format!("(?i){pattern}")) else {
-        return FindResult {
-            id: id.to_string(),
-            next_cursor: None,
-            matches: vec![],
-        };
+    let empty_result = || FindResult {
+        id: id.to_string(),
+        next_cursor: None,
+        relaxation: None,
+        matches: vec![],
     };
 
-    let all_matches: Vec<(usize, usize)> = re
-        .find_iter(content)
-        .map(|m| (m.start(), m.end()))
-        .collect();
-    let total_found = all_matches.len();
+    if pattern.trim().is_empty() {
+        return empty_result();
+    }
 
+    let parsed = parse_query(pattern);
+    let Some(expr) = parsed.expr else {
+        return empty_result();
+    };
+
+    let line_spans = compute_line_spans(content);
+    let lines: Vec<&str> = line_spans.iter().map(|&(s, e)| &content[s..e]).collect();
+    let lines_lower: Vec<String> = lines.iter().map(|l| l.to_lowercase()).collect();
+
+    let mut compiler = Compiler::default();
+    compiler.compile(&expr);
+
+    let (matched, relaxation) = if let Some(keywords) = parsed.bare_keywords.as_ref() {
+        select_with_fallback(keywords, &compiler, &lines, &lines_lower)
+    } else {
+        let m: Vec<usize> = (0..lines.len())
+            .filter(|&i| eval_expr(&expr, &compiler, lines[i], &lines_lower[i]))
+            .collect();
+        (m, None)
+    };
+
+    let occurrences =
+        collect_occurrences(&expr, &compiler, &line_spans, &lines, &lines_lower, &matched);
+
+    let total_found = occurrences.len();
     let mut matches: Vec<FindMatch> = Vec::new();
     let mut output_chars: usize = 0;
-    for (i, &(match_start, match_end)) in all_matches.iter().enumerate() {
+    for (i, occ) in occurrences.iter().enumerate() {
         if i < cursor {
             continue;
         }
         if matches.len() >= k || output_chars >= MAX_OUTPUT_CHARS {
             break;
         }
-
-        let ctx_start = floor_char_boundary(content, match_start.saturating_sub(context_bytes));
-        let ctx_end = ceil_char_boundary(content, (match_end + context_bytes).min(content.len()));
+        let ctx_start = floor_char_boundary(content, occ.start_byte.saturating_sub(context_bytes));
+        let ctx_end = ceil_char_boundary(
+            content,
+            (occ.end_byte + context_bytes).min(content.len()),
+        );
         let context = content[ctx_start..ctx_end].to_string();
         output_chars += context.len();
-
         matches.push(FindMatch {
-            keyword: pattern.to_string(),
-            start: match_start,
-            end: match_end,
+            keyword: occ.keyword.clone(),
+            start: occ.start_byte,
+            end: occ.end_byte,
             context,
         });
     }
@@ -106,6 +149,7 @@ pub fn find_in_document(
     FindResult {
         id: id.to_string(),
         next_cursor,
+        relaxation,
         matches,
     }
 }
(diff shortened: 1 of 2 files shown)


nuri-yoo:
@jhlee525 @grf53

jhlee525 (approved):
Wow, we’ve eventually built a DSL.

The query syntax now looks much more Lucene-like, so it feels fairly standard.

I’ve also been thinking about making the progressive fallback deterministic by exposing it as an explicit parameter `p` (so the LLM can adjust `p` over iterations when issuing queries). However, since the current approach is already a validated pattern, I don’t think this needs to be addressed right away.

Also, with this logic included, the document is starting to accumulate quite a lot of content. It might make sense to separate out the text query–related parts, but I’ll handle that later.

Thanks!

speedwagon/src/store/document.rs
```diff
@@ -56,48 +85,62 @@ pub fn find_in_document(
...
-            matches: vec![],
-        };
+    let parsed = parse_query(pattern);
+    let Some(expr) = parsed.expr else {
+        return empty_result();
+    };
+
+    let line_spans = compute_line_spans(content);
+    let lines: Vec<&str> = line_spans.iter().map(|&(s, e)| &content[s..e]).collect();
+    let lines_lower: Vec<String> = lines.iter().map(|l| l.to_lowercase()).collect();
```
grf53, line 106: Performing these lines(processing whole document content) every single `find` call could be quite heavy for big documents.

I reviewed their overhead and noticed that 'converting all content lines to lowercase is quite expensive.'

Based on my brief research, case-insensitive find can be performed without 'pre-converted content lines' for queries consisting of ASCII characters or characters for which cases are not separately considered, using the logic below.
With this, we can reduce big overheads for case-insensitive find, and simplify the signatures of related functions.

(Assuming `needle_lower` here is always lower-cased.)
```rust
fn find_ci_ascii(haystack: &str, needle_lower: &str) -> Option<usize> {
    let h = haystack.as_bytes();
    let n = needle_lower.as_bytes();
    if h.len() < n.len() {
        return None;
    }
    let last = h.len() - n.len();
    let mut i = 0;
    while i <= last {
        let mut j = 0;
        while j < n.len() {
            let hb = h[i + j];
            let folded = if hb.is_ascii_uppercase() { hb | 0x20 } else { hb };
            if folded != n[j] {
                break;
            }
            j += 1;
        }
        if j == n.len() {
            return Some(i);
        }
        i += 1;
    }
    None
}
```

In addition, to cover cases where cases exist that are not simple ASCII, such as äpfel, you can apply the following fallback logic.

```rust
fn find_ci(haystack: &str, needle_lower: &str) -> Option<usize> {
    if needle_lower.is_empty() {
        return None;
    }
    if needle_lower.is_ascii() {
        return find_ci_ascii(haystack, needle_lower);
    }
    haystack.to_lowercase().find(needle_lower)
}
```

grf53: Or, we just can actively reuse lowercase content lines created once.

speedwagon/src/tool/find.rs
```diff
@@ -51,7 +63,7 @@ pub fn build_find_in_document_tool(store: Arc<Store>) -> (ToolDesc, ToolFunc) {
                 },
                 "pattern": {
                     "type": "string",
-                    "description": "Regex pattern to search for"
+                    "description": "Query string. Bare words (e.g. `revenue cost`) get AND→half→any fallback. Use `\"phrase\"`, `+term`, `-term`, `AND`, `OR`, `NOT`, `(group)`, or `/regex/` for explicit semantics (no fallback)."
                 },
```
grf53, line 67: Have you ever had the opportunity to check if the LLM understands the grammar well and uses it differently depending on the situation?

Since the capabilities of modern LLMs are quite good, we could expect them to perform reasonably well; however, as it is a DSL capable of quite a wide variety of expressions, it seems necessary to verify whether the LLM can utilize it without difficulty.

grf53: Or, I think it's possible that the LLM attempts to utilize syntax excessively, thereby failing to leverage automatic query relaxation. While a good result can eventually be achieved through several failures and retries, it may be advisable to reduce inefficiencies arising from methodology bias.

Therefore, I believe it's necessary to briefly verify the behavior patterns of the LLM when this tool is given.

speedwagon/src/store/document.rs
nuri-yoo: Thanks for the suggestion. I ran a benchmark across 24 scenarios covering ASCII / Korean / mixed × small / medium / large × no-hits / many-hits / single / multi keyword / short / long.

Three variants compared:
- **current**: build `lines_lower: Vec<String>` once per call, use `String::contains()`.
- **option1** (the suggested byte-fold approach): no `lines_lower`. ASCII keywords use a byte-level case-fold scan; non-ASCII keywords fall back to `line.to_lowercase().find(...)` per line.
- **option2** (hybrid): if every literal keyword is ASCII, skip `lines_lower` and use the ASCII fast-path everywhere. If any keyword is non-ASCII, build `lines_lower` once and reuse.

| # | Scenario | current | option1 | option2 | Winner |
|---|---|---|---|---|---|
| S1 | en/200K, ASCII kw=2, 0hit | **708** | 1442 (0.49x) | 1461 (0.48x) | current |
| S1b | en/1M, ASCII kw=2, 0hit | **3526** | 7187 (0.49x) | 7252 (0.49x) | current |
| S2 | kr/200K, KR kw=2, 0hit | 3295 | 3276 (1.01x) | 3259 (1.01x) | tie |
| S2b | kr/1M, KR kw=2, 0hit | 16435 | 16315 (1.01x) | 16413 (1.00x) | tie |
| S3 | mix/200K, ASCII+KR kw=2 | 1976 | **1211 (1.63x)** | 2722 (0.73x) | option1 |
| S4 | en/50 lines, ASCII kw=2 | **7.1** | 14.1 (0.51x) | 15.2 (0.47x) | current |
| S5 | en/200K, ASCII kw=2, hit | **730** | 948 (0.77x) | 959 (0.76x) | current |
| S5b | en/1M, ASCII kw=2, hit | **3659** | 4748 (0.77x) | 4791 (0.76x) | current |
| S6 | kr/200K, KR kw=1, hit | 3210 | 3189 (1.01x) | 3191 (1.01x) | tie |
| S7 | en/200K, ASCII kw=1 | **628** | 1187 (0.53x) | 1232 (0.51x) | current |
| S8 | en/1M, ASCII kw=1 | **3119** | 6139 (0.51x) | 6169 (0.51x) | current |
| S9 | kr/1M, KR kw=1 | 16089 | 16028 (1.00x) | 16129 (1.00x) | tie |
| S10 | en/200K, ASCII kw=3 | **711** | 1398 (0.51x) | 1484 (0.48x) | current |
| S11 | en/1M, ASCII kw=5 | **3550** | 7149 (0.50x) | 7252 (0.49x) | current |
| S12 | kr/200K, KR kw=3 | 3266 | 3222 (1.01x) | 3301 (0.99x) | tie |
| S13 | en/1M, ASCII 1-char kw | **2137** | 4088 (0.52x) | 4115 (0.52x) | current |
| S14 | en/1M, ASCII 2-char kw=2 | **4828** | 7135 (0.68x) | 7377 (0.65x) | current |
| S15 | en/1M, 50-char ASCII phrase | 11065 | **4589 (2.41x)** | 4704 (2.35x) | option1 |
| S16 | kr/1M, long KR phrase | 23402 | 22996 (1.02x) | 23445 (1.00x) | tie |
| S17 | en/1M ALL-CAPS, ASCII kw=2 | **5219** | 10794 (0.48x) | 10896 (0.48x) | current |
| S18 | en/5MB, ASCII kw=2 | **17961** | 59156 (0.30x) | 37183 (0.48x) | current |
| S19 | kr/5MB, KR kw=2 | 82470 | 81555 (1.01x) | 82384 (1.00x) | tie |
| S20 | en/1M, KR kw=2 | 3823 | 3704 (1.03x) | 3840 (1.00x) | tie |
| S21 | en/200K, 5x repeated calls | **3151** | 6129 (0.51x) | 6355 (0.50x) | current |

All numbers in µs (median over 50 iterations, 5 warmup, release build, Apple M-series).

current wins in 14, option1 wins in 2 (S3, S15), and the rest are ties (typically Korean documents where the dominant cost is the substring search itself, not the lowercase pass).

### Why each variant performs the way it does

**current — wins on most ASCII workloads**
- `String::contains()` uses memchr's SIMD implementation. On modern hardware that's roughly 16-32 bytes per cycle for short needles, vs. ~1 byte per loop iteration for a hand-rolled scan.
- `String::to_lowercase()` on ASCII does no real character mapping — for already-lowercase bytes it's effectively a memcpy, and the entire `lines_lower` allocation amortizes across all keywords in the AND group.
- Loses on long ASCII phrases (S15) because needle length defeats SIMD throughput, so the `to_lowercase()` pass becomes pure overhead.
- Loses on mixed-language queries (S3) because the pre-computed `lines_lower` is wasted work — the ASCII keyword would have failed out on the very first byte of every Korean line via a byte-fold scan.

**option1 — fast on long phrases and mixed-language fail-fast paths**
- The byte-level fold (`b | 0x20`) is branch-free and inlines tightly, which matters when the substring search itself is the bottleneck (long needle on long haystack).
- Wins S3 by skipping `to_lowercase()` entirely on lines where the ASCII keyword can't possibly match (different byte ranges).
- Loses on short keywords because every byte goes through a manual loop with no SIMD acceleration. The gap is roughly 2x for 1–5 char ASCII needles and grows with input size (S18: 0.30x at 5MB).
- Pays per-line `to_lowercase()` for non-ASCII keywords, so on Korean docs it ends up doing the same amount of work as current — just spread across the loop instead of front-loaded — hence the consistent ~1.0x ties on S2/S6/S9/S12/S16/S19.

**option2 — almost always strictly worse than the better of the other two**
- Same fast path as option1 when all keywords are ASCII, so it inherits option1's losses on short-keyword ASCII workloads.
- Same `lines_lower` build as current when any keyword is non-ASCII, so it inherits current's overhead.
- The hybrid logic adds a per-keyword ASCII check and a branch in the inner loop without giving back enough wins to justify it. The only scenario where it edges current is on Korean docs, where the win is within noise (1.00x–1.01x).

For typical short-keyword queries (1-10 chars), which I expect to dominate LLM tool calls, current is roughly 2x faster on English documents and on par with option1 on Korean documents.

speedwagon/src/tool/find.rs
nuri-yoo: Followup eval — split into two phases to separate "does the model use syntax on its own?" from "does the model understand the syntax when nudged toward it?"

**Setup**
- model: `openai/gpt-5.4-mini` (speedwagon default)
- corpus: 9 FinanceBench 10-K documents (3M filings 2015–2023)
- Phase 1: 15 natural-language questions
- Phase 2: 8 "directed" questions where an explicit operator is the objectively right choice (exact phrase, exclusion, OR alternatives, regex pattern, strict AND, etc.) — wording nudges toward a specific operator without naming it. The last question hit a TPM rate limit so we have data for 7/8 directed questions.

### Phase 1 (Natural) — 30 calls

| Metric | Value |
|---|---|
| bare-word | 30 / 30 (100%) |
| any explicit operator | 0 / 30 (0%) |
| zero-hit responses | 1 / 30 (3.3%) |

`relaxation`: all=19 (63%), half=5 (17%), null=6 (20%).

Confirms the first eval — on natural questions the model never reaches for the structured-query syntax.

### Phase 2 (Directed) — 21 calls

| Metric | Value |
|---|---|
| bare-word | 12 / 21 (57%) |
| any explicit operator | 9 / 21 (43%) |
| zero-hit responses | 2 / 21 (9.5%) |

Operators actually used:
- `"phrase"`: 4 calls
- `/regex/`: 4 calls (one regex repeated 3x for pagination)
- `OR`: 1 call (combined with phrase)
- `NOT`: 1 call
- `+`, `-`, `AND`, `(group)`: 0 calls

`relaxation`: null=13, all=7, half=1.

### Every Phase 2 call (in order, per question)

| # | pattern | hits | relax | operators |
|---|---|---|---|---|
| 1 | `"cost of goods sold"` | 0 | null | phrase |
| 2 | `cost of goods sold` | 20 | half | bare |
| 3 | `/cost of goods sold/i` | 0 | null | regex |
| 4 | `goods sold` | 2 | all | bare |
| 5 | `revenue NOT deferred` | 19 | null | NOT |
| 6 | `revenue` | 19 | null | bare |
| 7 | `EBITDA` | 3 | null | bare |
| 8 | `"operating income"` | 32 | null | phrase |
| 9 | `/\b201[0-9]\b/` | 49 | null | regex |
| 10 | `/\b201[0-9]\b/` | 49 | null | regex (pagination) |
| 11 | `/\b201[0-9]\b/` | 49 | null | regex (pagination) |
| 12 | `total liabilities` | 12 | all | bare |
| 13 | `Risk Factors` | 11 | all | bare |
| 14 | `## Item 1A. Risk Factor s.` | 1 | all | bare |
| 15 | `"goodwill impairment" OR "impairment charge"` | 8 | null | phrase + OR |
| 16 | `"goodwill impairment"` | 5 | null | phrase |
| 17 | `acquisition divestiture` | 20 | all | bare |
| 18 | `divest` | 28 | null | bare |
| 19 | `2015 acquisition` | 20 | all | bare |
| 20 | `2015 acquisition activity` | 3 | all | bare |
| 21 | `acquisition` | 32 | null | bare |

### What this tells us

**The model does understand the DSL.** Given a directed question where an operator is the right tool, it picks the right operator 43% of the time (9/21 calls). The four operators it actually uses are exactly the ones that change behavior (`"phrase"`, `/regex/`, `OR`, `NOT`); the ones it never uses (`+`, `-`, `AND`, `(group)`) are either redundant (`+`/`AND` are the default in bare-word mode) or rare in natural usage.

**Specific evidence of comprehension:**
- Q1 ("exact phrase 'cost of goods sold'"): wraps in `"..."` correctly.
- Q3 ("either EBITDA or 'operating income'"): combines `"phrase"` with `OR`, picking the right syntax for each side.
- Q4 ("4-digit year between 2010 and 2019"): writes `/\b201[0-9]\b/` — correct word-boundary regex, not a brute-force enumeration of 10 separate searches.
- Q2 ("revenue but NOT deferred revenue"): correctly forms `revenue NOT deferred` rather than two separate queries.

**The two-layer recovery worked as intended.** On Q1 the model started with `"cost of goods sold"` and `/cost of goods sold/i`, both of which returned 0 hits because the corpus splits that phrase across bullet-point lines — a real characteristic of this document, not a model error. The model then naturally relaxed its own query to bare-word `cost of goods sold` (20 hits via `half` relaxation) and `goods sold` (2 hits via `all`). This is the recovery path the design was aiming for: when the model commits to an explicit-syntax query and the corpus doesn't cooperate, the bare-word + fallback path is right there as a safety net. The model didn't need any retry budget pressure or external prompting to find it — it self-corrected on the second exchange.

### Reading both phases together

The original concern was that adding the DSL would either (a) confuse the model, or (b) make it use the syntax aggressively and miss bare-word fallback. The data says neither happens in this run:

- (a) doesn't happen because the model demonstrably understands `"phrase"`, `/regex/`, `OR`, `NOT` and uses them correctly when the question calls for them.
- (b) doesn't happen on natural questions (0% explicit operators); on directed questions the model does reach for syntax (43%), and when an explicit query under-matches it relaxes its own query — at which point the bare-word fallback path is there to catch it.

The structured-query operators are essentially a power-user feature that the model only reaches for when explicitly prompted toward them, while staying on the bare-word + fallback path for ordinary lookups. That's the behavior pattern the design was hoping for — happy to revisit if a follow-up eval on a stronger model shows different behavior.

speedwagon/src/store/document.rs
grf53: Thanks for pointing out that it can be better to accept preprocessing costs like lowercasing to take advantage of SIMD. I appreciate you validating methods with actual measurements including my opinion given without a deep examination. It is good enough to select the current approach for now.

Some other optimization methods could be considered, but it doesn't seem to be needed here.

speedwagon/src/tool/find.rs
grf53: Thank you for the detailed evalution with clear implication!

grf53 (approved):
:+1:

PR #41, opened 2026-04-26
Add purpose metadata field to Store ingest

### LINKED ISSUE #37: Add `purpose` as a metadata field to `Store`
## Description
There is [`Store::ingest()`](https://github.com/brekkylab/agent-k/blob/refactoring-applied/speedwagon/src/store/mod.rs#L53) function in `refactoring-applied` branch.
There is only a metadata `title` which you can get with `parser::get_title()`.

We need to add `purpose` which is [tested](https://chat.google.com/room/AAQAHkViqQc/Bw7uLcxP2aA/Bw7uLcxP2aA?cls=10) before.

## Related Task
- https://github.com/orgs/brekkylab/projects/4/views/1?pane=issue&itemId=177474756

nuri-yoo:
## Summary

Adds an LLM-generated `purpose` metadata field to the speedwagon `Store` ingest pipeline. This was the best-performing variant in the metadata ablation: **+15.3pp BM25 hit@5 on FinanceBench** (32.0% → 47.3%) over the title-only baseline, beating the alternative `intent + summary` two-field schema (+10.0pp) at lower generation cost.

Closes #37.

### Schema
Tantivy schema gains a `purpose` field (`TEXT | STORED`) alongside `title` and `content`. `QueryParser` default fields are now `[title, purpose, content]` so purpose terms contribute to BM25 scoring at search time.

### Generation (`parser.rs`)
- New `PurposeAgent` mirrors the existing `TitleAgent` pattern (Ailoy `openai/gpt-5.4-mini`).
- Prompt and preview strategy reproduce the ablation v1 winner: first 3000 chars of content, JSON output (`{"purpose": "..."}`), one-sentence 80–150 char framing optimized for BM25 retrieval.
- Response parsing handles plain JSON and JSON wrapped in surrounding text. Empty/invalid responses fall back to `""` with a `log::warn!` so ingest still succeeds; transport-level errors propagate via `?`.

### Ingest (`mod.rs`)
`Store::ingest` and `Store::ingest_many` now run `get_title` and `get_purpose` in parallel via `tokio::try_join!` to avoid serializing the two LLM calls.

### Index compatibility (`indexer.rs`)
`open_or_create` detects schema mismatch on existing index directories and **automatically rebuilds the index** (`fs::remove_dir_all` + recreate). The corpus directory is preserved, so users only re-pay the LLM cost on first run after upgrade. A `log::info!` line announces the rebuild so the cost is visible. The `add_document` / `add_documents` / `delete_document` / `get_document` / `get_documents` / `list_documents` functions all round-trip the new field.

### Document struct (`document.rs`)
`Document` gains `pub purpose: String`. The `Display` impl is updated accordingly.

## Tests

13 new unit tests (parser response parsing, schema build, schema-mismatch rebuild path, purpose round-trip through every indexer function, search hits via purpose-only matches). Existing tests updated for the new `(id, title, purpose, content)` tuple shape.

```
cargo test --lib
test result: ok. 27 passed; 0 failed; 1 ignored
```

`cargo check --features internal` and `cargo clippy --features internal --tests` are clean for changed code.

## Notes

- The schema-rebuild policy is intentionally simple: delete and re-create. There is no migration helper. This is fine while `refactoring-applied` is pre-production; if migration becomes a concern later, a `.schema_hash` cache file can short-circuit unchanged-schema reopens.
- `get_purpose` always calls the LLM (no frontmatter fallback like `get_title`). The empty-string fallback on bad LLM output keeps the index at baseline quality rather than failing the whole ingest.

### DIFF
--- a/speedwagon/src/store/document.rs
+++ b/speedwagon/src/store/document.rs
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 pub struct Document {
     pub id: String,
     pub title: String,
+    pub purpose: String,
     pub content: Option<String>,
     pub len: usize,
 }
@@ -27,8 +28,8 @@ impl fmt::Display for Document {
         };
         write!(
             f,
-            "{{ id: {}, title: {}, len: {}, content: {} }}",
-            self.id, self.title, self.len, content
+            "{{ id: {}, title: {}, purpose: {}, len: {}, content: {} }}",
+            self.id, self.title, self.purpose, self.len, content
         )
     }
 }
--- a/speedwagon/src/store/indexer.rs
+++ b/speedwagon/src/store/indexer.rs
@@ -5,39 +5,75 @@ use tantivy::{
     Index, IndexWriter, TantivyDocument, Term,
     collector::TopDocs,
     query::{AllQuery, BooleanQuery, Occur, Query, TermQuery},
-    schema::{IndexRecordOption, OwnedValue, STORED, STRING, TEXT},
+    schema::{IndexRecordOption, OwnedValue, STORED, STRING, Schema, TEXT},
 };
 
 use super::Document;
 
+fn build_schema() -> Schema {
+    let mut sb = Schema::builder();
+    sb.add_text_field("id", STRING | STORED);
+    sb.add_text_field("title", TEXT | STORED);
+    sb.add_text_field("purpose", TEXT | STORED);
+    sb.add_text_field("content", TEXT | STORED);
+    sb.build()
+}
+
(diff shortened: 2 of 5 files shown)


nuri-yoo:
@grf53 @jhlee525

jhlee525 (approved):
👍

grf53 (approved):
It appears that changes are needed to ensure the API provider is always retrieved from the environment variable `OPENAI_API_KEY` and to address the issue where other providers cannot be utilized; however, this is considered a problem that should be handled as a separate issue, and the changes in this PR are deemed sufficient.

PR #46, opened 2026-04-28
Add HTML support for Store

### LINKED ISSUE #43: Add HTML support for `Store`
## Description
Adding a feature to convert HTML documents into Markdown format.

Since HTML is already a text format, there is no need to convert it separately for the sole purpose of understanding LLM. This can be interpreted as adopting Markdown as the 'normalized form' of the documents used in the project.

grf53:
## Ticket
resolves #43 

## Summary
This PR unifies file type handling around `FileType` and applies the same mapping rules to CLI ingest.

## Changes
This PR is split into two areas:  
1. `FileType` / `translator` structure refactor  
2. HTML translator addition

### `FileType` / `translator` structure refactor
- Centralized extension handling around `FileType`.
  - Added `from_extension`, `from_path`, `canonical_extension`, and `supported_extensions`.
- Switched translator dispatch from string-based extension checks to enum-based routing.
  - `translate(filetype, origin_path, corpus_path)`
  - Handles `PDF` and `HTML`; explicitly rejects `MD` in the translator path.
- Updated `Store::ingest` / `ingest_many` to pass the existing `FileType` directly to translator.
- Standardized origin filename extension generation using `canonical_extension`.
- Updated CLI `/ingest` to use the same shared `FileType` mapping, keeping behavior consistent for `pdf/html/htm/md`.

### HTML translator addition (chrome-strip body extraction)
- Added an HTML-to-Markdown translator path with YAML frontmatter output.                                                                                                                                                                                       
  - Uses dom_smoothie for metadata only (title, byline, excerpt, dates, language, image, favicon, …) — article.content is intentionally not used for the body.
  - Body extraction goes through a unified dom_query chrome-strip pass; output is then converted via html-to-markdown-rs.                                                                                                                                         
  - Strip rules cover both generic and framework-specific noise:                                                                                                                                                                                                  
    - Generic HTML5 chrome: `<script>` / `<style>` / `<noscript>` / `<nav>` / `<header>` / `<footer>` / `<aside>`, aria landmarks (`[role=navigation|banner|contentinfo|complementary]`), and hidden nodes (`[hidden]`, `[aria-hidden=true]`).                                        
    - MediaWiki / Wikipedia / Fandom in-content noise: `.mw-editsection`, `.mw-jump-link`, `.mw-indicators`, `.mw-empty-elt`, `.printfooter`, `.catlinks`, `.sistersitebox`, `.hatnote`, `.mbox-text-span`, `.navbox`, `.navbox-inner`.                                                 
  - Returns `<body>` inner HTML after stripping (or the whole document when there's no `<body>`); empirically preserves heading structure as well as Readability-style scoring while avoiding the sibling-section pruning that Readability commonly causes.

## Validation
`cargo test -p speedwagon --no-default-features --lib store::translator`


speedwagon/src/store/translator/html.rs
```diff
@@ -0,0 +1,180 @@
...
+        readable_min_score: 10.0,
+        readable_min_content_length: 70,
+        n_top_candidates: 10,
+        ..Default::default()
+    };
+    let mut reader = dom_smoothie::Readability::new(html.as_str(), None, Some(cfg))
+        .map_err(|e| anyhow::anyhow!("dom_smoothie::new: {e}"))?;
+    let article = reader
+        .parse()
+        .map_err(|e| anyhow::anyhow!("dom_smoothie::parse: {e}"))?;
```
nuri-yoo: parse() failures fail the whole ingest. I verified 4 degenerate inputs all hit "failed to grab the article": empty string, whitespace only, just `<!doctype html>`, and `<html><body></body></html>`.

If dom_smoothie is for metadata only, take it as Option and let the body path proceed when metadata extraction fails.

```rust
let article = reader.parse().ok();
// then article.as_ref().and_then(...)
```

speedwagon/src/store/translator/html.rs
```diff
@@ -0,0 +1,180 @@
...
+    if let Some(s) = article.dir.as_deref() {
+        add_field(&mut out, "dir", s);
+    }
+    if article.length > 0 {
+        out.push_str(&format!("text_length: {}\n", article.length));
+    }
+    out.push_str("---\n\n");
+
+    // 6. Optional H1 prepend.
+    if !article.title.is_empty() && !body_starts_with_h1(&md_body) {
```
nuri-yoo: After chrome-strip some pages still keep layout-level H1s. Repro on rust-lang.org book ch1 yields 3 H1s in the output:

```
# Getting Started - The Rust Programming Language
# The Rust Programming Language
# [Getting Started](#getting-started)
```

body_starts_with_h1 only checks the first H1. Tighten the condition to "no H1 anywhere in the body".

```rust
let has_h1 = md_body.lines()
    .any(|l| l.trim_start().starts_with("# "));
if !article.title.is_empty() && !has_h1 { ... }
```

speedwagon/src/store/translator/html.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+fn body_starts_with_h1(body: &str) -> bool {
+    body.lines()
+        .find(|l| !l.trim().is_empty())
+        .map(|l| l.trim_start().starts_with("# "))
+        .unwrap_or(false)
+}
+
+fn yaml_str(s: &str) -> String {
```
nuri-yoo, line 110: yaml_str escapes `\` and `"`, but parser::parse_title at store/parser.rs:8 just does `.trim_matches('"').trim_matches('\'')` and doesn't unescape. The round-trip breaks:

```
input  : title: "Hello \"World\""
parsed : Some("Hello \\\"World\\")
```

Titles with quotes (e.g. `<title>That "Smart" Move</title>`) get indexed under a corrupted title. Fix parse_title alongside this, or split it into a separate issue.

Switching to single-quoted scalars narrows the escape surface (`'` → `''`).

speedwagon/src/store/translator/html.rs
grf53: Makes a lot of sense.
`dom_smoothie` was an essential part of parsing, but it's not anymore.
And right, it is just for metadata, so let me apply that.

speedwagon/src/store/translator/html.rs
grf53: Good point.
If there is already an H1 tag in the body, there is no need to add it additionally.
Let me apply this.

As a side note, `# The Rust Programming Language` is included in `<div class="menu-bar">` corresponding to `<header>`, so ideally it should be removed and only `# [Getting Started](#getting-started)` remain, but this cannot be covered by the current logic.

speedwagon/src/store/translator/html.rs
grf53: And also, `dom_smoothie` is quite big to use it only for metadata extraction.
Since its original purpose has been replaced by sth else, it seems like a good direction to replace the metadata extraction part and remove the `dom_smoothie` dependency.

grf53: I found that using `exclude_selectors` in `html_to_markdown_rs::ConversionOptions` eliminates the need for stripping via `dom_query`. Therefore, I remove `dom_query` along with `dom_smoothie`.

nuri-yoo (approved):
Verified all three comments resolved. LGTM.

speedwagon/src/store/translator/html.rs
```diff
@@ -0,0 +1,191 @@
...
+        .content
+        .clone()
+        .context("html-to-markdown-rs returned no content")?;
+    let doc_meta = &result.metadata.document;
+    let title = doc_meta.title.clone().unwrap_or_default();
+
+    let mut out = String::from("---\n");
+    out.push_str("converter: html-to-markdown-rs\n");
+    add_field(&mut out, "title", &title);
+    if let Some(s) = doc_meta.author.as_deref() {
```
selloriwoo, line 33: This pattern repeats for nearly every field — making add_field accept Option<&str> would let each call collapse to a single line.   
``` 
   add_field(&mut out, "title", doc_meta.title.as_deref());
    add_field(&mut out, "author", doc_meta.author.as_deref());
    add_field(&mut out, "excerpt", doc_meta.description.as_deref());
    add_field(&mut out, "site_name", doc_meta.open_graph.get("site_name").map(String::as_str));
    add_field(&mut out, "language", doc_meta.language.as_deref());
    add_field(&mut out, "published_time", doc_meta.meta_tags.get("article-published_time").map(String::as_str));
    add_field(&mut out, "modified_time", doc_meta.meta_tags.get("article-modified_time").map(String::as_str));
```

```
fn add_field(out: &mut String, key: &str, value: Option<&str>) {
        if let Some(s) = value
            && !s.is_empty()
        {
            out.push_str(&format!("{key}: {}\n", yaml_str(s)));
        }
    }
```

selloriwoo: Looks like my review was left in a pending state, sorry about that 😅

PR #52, opened 2026-05-06
feat: introduce reflect-agent crate (single lead via AgentBuilder)

nuri-yoo:
## Summary

New workspace member `reflect-agent`. Single main agent built through `ailoy::agent::AgentBuilder` with three default builtin tools (`bash`, `python_repl`, `web_search`). Local runenv by default; sandbox feature gate is plumbed for a follow-up. No subagents, no verify gate yet.

This PR is the foundation for the deterministic verify gate work that lands as a separate stacked PR.

## Stacked on PR #51

Base = `fix/ailoy-bump-391`. Once #51 merges, GitHub will retarget this PR's base to `refactoring-applied` automatically. The two ailoy API changes that this PR depends on land in #51:

- `AgentBuilder` (post-#391 ailoy)
- `register_provider_from_env` pattern (PR #51's `speedwagon::provider`, mirrored here at `reflect_agent::provider` — kept duplicated to avoid pulling speedwagon's full dependency tree for a 12-line wrapper).

## Why a new crate

Three options were considered: extend `chat-agent`, extend `speedwagon`, or create a new crate. New crate won because:

- `chat-agent` is not on the active development path of this branch.
- `speedwagon` is a worker (KB-bound search agent), not a lead. Mixing lead-side gate logic into speedwagon would conflate the two roles — `reflect-agent` keeps that separation explicit.
- A new crate gives the gate work a stable, well-scoped home that survives future architectural shifts (a future orchestrator-workers shape can layer on top without renaming).

The name `reflect-agent` reflects the crate's primary identity: the home for self-check passes (verify gate, future reflect gate).

## Module layout

| File | Purpose |
|---|---|
| `lib.rs` | Re-exports of `build_agent`, `DEFAULT_MODEL`, `register_provider_from_env`. |
| `agent.rs` | `build_agent(model)` — clones the global provider, attaches `bash` / `python_repl` / `web_search` to `provider.tools`, then `AgentBuilder::new(model).provider(...).tool("bash").tool("python_repl").tool("web_search").build()`. |
| `provider.rs` | `register_provider_from_env(&mut AgentProvider)` — reads `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GEMINI_API_KEY` and registers the same glob patterns ailoy's removed convenience constructors used (`openai/*`, `anthropic/claude-*`, `google/gemini-*`). |
| `main.rs` | CLI: REPL by default, or `--query "..."` for a single non-interactive turn. |

[...]

PR #51, opened 2026-05-06
fix(deps): bump ailoy to post-#391 (AgentBuilder + sandbox sharing)

nuri-yoo:
## Summary

Bumps ailoy from `c42231e1` (2026-04-21) to `098a8289` ([brekkylab/ailoy#391](https://github.com/brekkylab/ailoy/pull/391), 2026-05-04). Pulls in 27 commits worth of breaking changes; this PR only patches `speedwagon` to compile and pass tests against the new API.

## Why

Speedwagon's pinned ailoy was three substantive refactors behind: `ToolFactory` introduction (#389), `RunEnv` trait + sandbox feature split (#390), and the registry overhaul that removed `ToolSet` / `model_openai` helpers / `Agent::try_with_tools` (#391). Each is a hard breaking change. Without this bump, speedwagon's API surface diverges from current ailoy, blocking any future work that wants to use both.

## Breaking changes absorbed

| ailoy change | speedwagon-side adjustment |
|---|---|
| `ToolSet` removed; `ToolFactory` introduced | `tool/mod.rs`: `build_toolset(store) -> ToolSet` renamed to `build_tool_provider(store) -> ToolProvider`, using `ToolProvider::new().custom(ToolFactory::simple(desc, func))` per tool. |
| `Agent::try_with_tools(spec, provider, toolset)` removed; tools live on `provider.tools` | `main.rs::build_agent`: clones the global provider, overrides `provider.tools` with the store-bound tool registry, then `Agent::try_with_provider(spec, &provider)`. |
| `AgentProvider::get_model(name)` removed; replaced by `AgentProvider.models.get(name)` | `store/helper.rs:88`: same call rewritten. Same semantics, new accessor location. |
| `AgentProvider::model_openai/model_claude/model_gemini` helper constructors removed | New module `speedwagon::provider` with `register_provider_from_env(&mut AgentProvider)` that reads `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` / `GEMINI_API_KEY` and registers the same glob patterns the removed helpers used (`openai/*`, `anthropic/claude-*`, `google/gemini-*`). `main.rs` and the `description.rs` integration test both call this — keeps the env-key → glob-pattern mapping in one place, preserving #49's invariant that helper modules never read env directly. |
| `RunEnv` trait + sandbox feature split | No direct touch — speedwagon never constructs sandboxes. The default `Local` runenv gets used implicitly when an `Agent` is built without `.runenv()`. |

## Out of scope

[...]

PR #53, opened 2026-05-06
feat(reflect-agent): deterministic verify gate (detection layer)

nuri-yoo:
## Summary

This PR is a **proof of concept** for a deterministic (no-LLM) verify gate on top of `reflect-agent`. After every agent turn, the new `run_with_verify` pass inspects the slice of history that turn just produced and flags four classes of issue without making a single LLM call. When at least one issue is found, the report goes to stderr; on a clean run, stderr stays silent.

The signal logic is the part that earns its keep: `verify_run` is a pure function over a `&[Message]` slice, so the same checks remain reusable from any cut of history if a future PR wants to act on them earlier than turn-end.

## Stacked on PR #52

Base = `feat/reflect-agent`. Once #52 merges, GitHub will retarget this PR's base to `fix/ailoy-bump-391`, and once #51 merges that retargets again to `refactoring-applied`. Nothing else needs to move.

## What this layer does and doesn't do

| | Detection (this PR) | Possible future direction |
|---|---|---|
| **Timing** | post-hoc, turn-end | per tool-result / per final-message |
| **Granularity** | tool calls of a single lead | subagent results once a worker pool exists |
| **Action** | report to stderr only | replace value / stop turn / reject final |

Whether to build the right column — and exactly how — is a question for after this layer has run on real workloads and we know which signals fire often enough to justify intercepting.

## Signals and where they came from

Each signal is a deterministic check on the history slice produced by one agent turn. The patterns are inspired by what production coding agents already encode in their published system prompts; this PR adds the deterministic counterpart that runs without invoking the model.

| Signal | What it detects | Inspired by |
|---|---|---|
| `EmptyResult` | Tool returned an empty value: `null`, an empty string / array / object, or a fixed-schema object whose every content field (string / array / nested object) is empty. Numeric / boolean metadata fields like `exit_code: 0` are ignored. | Devin's "gather information before concluding" guidance ([Prompt.txt](https://raw.githubusercontent.com/x1xhlol/system-prompts-and-models-of-ai-tools/main/Devin%20AI/Prompt.txt) L12); Cowork's generalized verification step in TodoList ([claude-cowork.md](https://raw.githubusercontent.com/asgeirtj/system_prompts_leaks/main/Anthropic/claude-cowork.md)) |
| `LoopDetected` | Same `(tool_name, args)` invoked at least `loop_threshold` times (default 3). Threshold is tunable via `VerifyConfig`. | Devin's three-attempt CI rule: *"ask the user for help if CI does not pass after the third attempt"* ([Prompt.txt](https://raw.githubusercontent.com/x1xhlol/system-prompts-and-models-of-ai-tools/main/Devin%20AI/Prompt.txt) L402) |
| `UnverifiedCitation` | Final assistant text mentions a source (HTTP/HTTPS URL, file path, or ISO-8601 timestamp) that doesn't appear in the tool log under any of a small set of representational variants (see *Citation matching* below). | Cowork's `<citation_requirements>` block (*"Claude MUST include a Sources: section"*) ([claude-cowork.md](https://raw.githubusercontent.com/asgeirtj/system_prompts_leaks/main/Anthropic/claude-cowork.md)); Devin's *"Don't assume content of links without visiting them"* ([Prompt.txt](https://raw.githubusercontent.com/x1xhlol/system-prompts-and-models-of-ai-tools/main/Devin%20AI/Prompt.txt) L26). Cowork enforces *citing*; this check enforces that the cited source actually came from a tool. |
| `BashFailure` | The `bash` tool result reports a non-zero exit code, a timeout, or a validation error (missing required parameter). | Standard Unix convention; ailoy's bash tool already separates `stdout` / `stderr` / `exit_code` / `timed_out` fields, so the check is a direct shape match. |

## Citation matching

The check tries cheap probes first and only reaches for the variant probes on miss:

1. **Exact substring** — the citation appears verbatim somewhere in the tool log (the original PR-#53 behaviour).
2. **URL variants** — trailing slash on/off, and `http://` ↔ `https://` swap.
3. **Path normalisation** — strip leading `./` / `~/` and trailing `/` from the citation before the substring check. Absolute-vs-bare paths are *not* normalised: the fuzzy step only strips suffixes/leads, never invents a prefix.
4. **Timestamp prefixes** — for ISO-style timestamps, walk the citation back along its natural separators (zone → `:` → `T` → date) so a precise instant is licensed by a date-only haystack. The walk is asymmetric: vague tool output can satisfy a precise quote, but a precise tool output already covers a vague quote via step 1.

Failures still surface as `UnverifiedCitation`; the matcher only changes which citations make it past the probes.

## What's in this PR

- `verify.rs` — `VerifyConfig`, `VerifyReport`, `Issue`, `verify_run`. Pure functions; no async, no LLM, no I/O.
- `verify.rs::appears_in_haystack` — the citation matcher described above. Probes are layered cheapest-first so the common case stays a single substring check.
- `agent.rs::run_with_verify(agent, query, config)` — drives `Agent::run` to completion, then calls `verify_run` on the new history slice.
- `main.rs::run_query` — uses `run_with_verify` and prints the report to stderr only when issues were found. Clean runs leave stderr empty.
- `tests/demo_task.rs` — three integration tests for the demo scenario (unstructured experiment log → structured `(timestamp, metric, value)` table → time-vs-metric plot):
  - `demo_task_full_failure_trajectory_fires_all_four_signals` — fixture-based, deterministic. Hand-built history that hits every signal at once.
  - `demo_task_clean_run_produces_no_issues` — fixture-based, deterministic. Happy path; report is empty.
  - `demo_task_end_to_end_pipeline` — `#[ignore]`d. Spins up a real agent against the configured provider, parses a small log fixture, asserts the verify pipeline produces a well-formed report. Run with `cargo test --test demo_task -- --ignored`.

## Design notes

- **Empty-result detection generalises beyond `null`/`{}`/`[]`.** Tools with a fixed schema (ailoy's `bash` returns `{stdout, stderr, exit_code, timed_out}` even for a no-output run) would otherwise never trip the check. `is_empty_value` therefore looks at content fields only and ignores numeric / boolean metadata. The full failure trajectory test pins this behaviour.
- **Citation matching is fuzzy by enumeration, not heuristic.** The matcher only knows the variant set above (URL slash/scheme, path leads, timestamp prefixes) — there's no Levenshtein, no token-ngram, no language model. If a future paraphrase pattern shows up in real runs, we add it as another deterministic probe rather than loosening the existing ones.
- **`loop_threshold` is the only tunable today.** The `VerifyConfig` struct exists so future signals (e.g. a stderr-dominance ratio) have a single home rather than scattering knobs across function signatures.

## Verification

```
cargo check --workspace --tests --all-features         # clean
cargo test  -p reflect-agent --lib                      # 29 passed; 0 failed
cargo test  -p reflect-agent --test demo_task           # 2 passed; 0 failed; 1 ignored
cargo test  -p speedwagon  --lib --all-features         # 71 passed; 0 failed; 2 ignored (no regression)
cargo run   -p reflect-agent -- --model openai/gpt-4o-mini \
            --query "Use bash to run \`exit 1\` and report what you see."
# stdout: agent's response describing exit code 1
# stderr: ─── verify gate findings ─── / - bash exited with code 1
```

## Test plan

- [x] `cargo check --workspace --tests --all-features` is clean
- [x] `cargo test -p reflect-agent --lib` passes 29/29 (22 originals + 7 fuzzy-match cases)
- [x] `cargo test -p reflect-agent --test demo_task` passes 2/2 (one ignored)
- [x] No regression on `speedwagon`: 71/0/2 still green
- [x] Smoke run with a forced bash failure (`exit 1`) produces exactly one `BashFailure NonZeroExit 1` issue on stderr
- [x] Citation matcher pins all four variant directions: timestamp precise→date, date→precise, fabricated stays flagged, URL trailing-slash + http/https swap, `~/` path normalisation, absolute-vs-bare stays flagged


nuri-yoo:
@grf53 @jhlee525

jhlee525 (commented):
My understanding of “reflection” is that it verifies the final output produced after the LLM has used tools and completed its actions.

The current implementation, however, seems to validate a single tool call instead.

Thus, I expected there to be a **verify tool**, the one that rely on the LLM, and the agent's system prompt would instruct to use the "verify tool" to perform reflection before emitting the result to user.

Please feel free to let me know if there something I misunderstand.

nuri-yoo:
Two separate verification shapes are in play, and the body wasn't clear enough about the split.

This PR is the **deterministic** half: Rust-only checks that run once at turn-end against the new history slice (every tool call + the final assistant message). No LLM call, no `verify` tool. The PR stack tells the story — #51 lines up the ailoy version, #52 routes the agent through `AgentBuilder`, and #53 puts the signal logic in `verify.rs` with tests, written so the same functions can move to wherever they're called from later. The intent of #53 is to land the logic and review it.

The shape you're describing — a `verify` tool the agent calls before emitting, prompted to use it — is the **LLM-driven** half. That has more moving parts to think through, so I started with the lower-impact piece (the deterministic gate) first.

The crate name covers both layers; this PR is only the first.

reflect-agent/src/verify.rs
```diff
@@ -0,0 +1,1076 @@
...
+/// - HTTP(S) URLs
+/// - Absolute or `./` / `~/` file paths
+/// - ISO-8601 timestamps (the demo task's main citation form: experiment
+///   logs carry per-event timestamps that should round-trip to the plot)
+fn extract_citations(text: &str) -> Vec<String> {
+    let mut out = Vec::new();
+    out.extend(scan_pattern(text, is_url_char, "http://"));
+    out.extend(scan_pattern(text, is_url_char, "https://"));
+    out.extend(scan_paths(text));
+    out.extend(scan_iso_timestamps(text));
```
grf53: All these `scan_*()` functions are implemented directly without using libraries or sth else.
You might use `regex` or `chrono` for some functions, which is not chosen here.
Is this intended? If so, please share the reason for this.

grf53: Currently, the haystack is traversed once for each scan target, resulting in a total of 4 traversals.

Since the representations of each scan target cannot overlap, applying multiple scan methods simultaneously would allow the same scan to be performed in a single traversal, which could be more efficient.

This approach may require caution, as it could excessively increase code complexity compared to the actual efficiency improvement. Since the current code is easy to understand the intentions, so maintaining the current implementation might be worthwhile.

nuri-yoo: It was intended. Purpose was to minimize dependencies in first PR (which is PoC) and thought main function of regex is to cover complex expression. Considering scalability and other comments below, implementing regex and chrono would be more robust and simple. Will fix together.

nuri-yoo: Didn't consider traversal issues. Regarding complexity, local testing showed `regex` can replace                                            
 the four scans with a single pass with same patterns, ~25 lines instead of ~95,  and only one walk over the input.  Pushing shortly.

reflect-agent/src/verify.rs
```diff
@@ -0,0 +1,1076 @@
...
+            } else {
+                break;
+            }
+        }
+        // Trim a trailing punctuation we don't want to be part of the citation.
+        let candidate = trim_trailing_punct(&text[start..end]);
+        if candidate.len() > prefix.len() {
+            out.push(candidate.to_string());
+        }
+        i = end.max(start + prefix.len());
```
grf53:
```suggestion
        i = end;
```

`end` is initialized with immutable `start + prefix.len()`, and `end` can only be increased (by adding `usize` variable). So `.max()` is not needed.

grf53: This part is also outdated now. I will handle it with resolve.

nuri-yoo: Thanks. The whole `scan_pattern` helper is gone in f9299ab. The single pass regex now doesn't track an index,  this `.max()` no longer exists.

reflect-agent/src/verify.rs
```diff
@@ -0,0 +1,1058 @@
...
+        for issue in &self.issues {
+            out.push_str("- ");
+            out.push_str(&format_issue(issue));
+            out.push('\n');
+        }
+        out
+    }
+}
+
+fn format_issue(issue: &Issue) -> String {
```
grf53: You could consider implementing `fmt::Display` for `Issue` instead of this.

reflect-agent/src/agent.rs
```diff
@@ -48,6 +54,32 @@ fn attach_default_tools(provider: &mut AgentProvider) {
...
 }
 
+/// Stream one user turn, collect every [`MessageOutput`] the agent emits,
+/// and run the post-hoc verify pass on the slice of history that this turn
+/// just appended.
+///
+/// This is the Stage A entry point for the verify gate: the agent has
+/// already finished and produced a response by the time the report is
+/// built. Stage B (a future PR) will move signal evaluation between turns
+/// via an ailoy-side hook so the same checks can intercept tool results.
```
grf53: (This is just a comment not related to the features.)

I believe that comments regarding future implementation plans are necessary only when the current implementation might be difficult to understand or appear inappropriate without the context that it was designed with future plans in mind.

(Meanwhile, I think 'TODO' comments for small phrases that need improvement are acceptable as bookmarks or reminders to help quickly identify issues in the near future.)

The 'Stage A' and 'Stage B' mentioned here were not used in the prior discussion and were not explained in the PR description, making it somewhat difficult to find their context.

I believe this PR(or every PR we do) is distinguishable from others because the changes in it are meaningful as itself; therefore, I do not think it is good practice to leave such comments in the code.

It would likely have been better if the explanation for Stage B had been in the PR or a separate document.

grf53: Does the mentioned `ailoy-side hook` require functional changes to the current ailoy? If so, it seems likely that a comprehensive assessment will be required to determine whether it is truly necessary for the ailoy.

nuri-yoo: I had a personal assessment and concluded that to really 'block' or 'intercept' as a gate.  Current PRs works itself without ailoy change. (which is not actual block/intercept, after turn)

nuri-yoo: Stage A/B was the original  framing which I had in mind for first time. 
Split into separate PRs that wording went stale, other comment as well.
Thanks for the feedback.

reflect-agent/src/verify.rs
nuri-yoo: applied in latest commit 17c551e

reflect-agent/src/verify.rs
grf53: Maybe we need to update `Cargo.lock` to add `"chrono"` and `"regex"` in `dependencies`.

nuri-yoo: Applied in f9299ab. Cargo now has the correct dependencies.

grf53 (approved):

jhlee525 (approved):
👍 

We'll later revise it, but for now, it looks good

PR #54, opened 2026-05-06
feat: add auth/user APIs

### LINKED ISSUE #16: Implement user system: auth + per-user management

khj809:
### New Features (auth/user APIs)
- **JWT authentication**: `POST /auth/signup`, `POST /auth/login` — issues HS256 tokens (`AGENT_K_JWT_SECRET` env var; falls back to an insecure default if unset, `JWT_EXPIRY_SECONDS` defaults to 7 days)
- **Current user endpoints**: `GET /me`, `PATCH /me` — requires JWT; password change validates current password
- **Admin user management**: `GET/POST /admin/users`, `GET/PATCH/DELETE /admin/users/{id}` — admin role only
- **Admin bootstrap**: on server start, auto-creates admin from `AGENT_K_ADMIN_USERNAME/PASSWORD` env vars if no admin exists; `create-admin` CLI subcommand also available
- **Middleware**: `auth_required` (JWT validation + `is_active` check), `admin_required` (role check)
- **Password**: argon2id hashing, minimum 8-character validation

### Structural Refactoring
- **Handler split**: 650-line `router.rs` → `handlers/{auth,user,session}.rs` with a pure routing `router.rs`
- **Auth module cleanup**: `validate_password` → `auth/password.rs`, `bootstrap_admin_if_needed` → `auth/bootstrap.rs`, CLI logic → `cli.rs`
- **sqlx migrations**: replaced inline DDL with versioned `migrations/0001_create_sessions.sql` and `migrations/0002_create_users.sql`

### Bug Fixes & Improvements
- **`create_user_admin` two-phase commit removed**: added `is_active` field to `NewUser` so the user is created with the correct state in a single INSERT (previously: created active then immediately updated — response body always returned `is_active: true` even when `false` was requested)
- **`update_user` dynamic SQL**: replaced manual binding-order synchronisation with `sqlx::QueryBuilder`
- **`UserListQuery`**: added missing `JsonSchema` derive

## Test Plan
- [x] `cargo test --lib` — 8 unit tests (repository, sqlite)
- [x] `cargo test --test auth_test` — 22 integration tests (signup, login, /me, admin CRUD, bootstrap)
- [x] Existing session/message integration tests pass without regression

## Out of Scope (follow-up)
- Add `user_id` FK to `sessions` / `session_messages` (requires separate migration strategy)
- Token refresh / revocation
- Rate limiting / login failure counter


jhlee525 (approved):

backend-v2/src/router.rs
ljhh-0611, line 1: [chore] The code mixes two styles: some places use axum::routing:: prefix, while others use use statements to avoid the prefix.

backend-v2/src/model/mod.rs
```diff
@@ -1,6 +1,6 @@
 mod message;
 mod session;
-pub mod user;
+mod user;
```
ljhh-0611, line 3: [Question] Could we consider making the subfolders under model and handlers pub? I'm worried that without proper categorization, it will become harder to navigate as the project grows.

backend-v2/src/router.rs
khj809: Thanks! I unified those usages into `aide::axum::routing`.

backend-v2/src/model/mod.rs
khj809: The purpose of making these mods private is to limit the depth of module import path as 3 (e.g. `crate::model::SignupRequest`).
However, as you mentioned, there would be conflicting names as the size of project grows, so I'm also open to making them as `pub(crate)`. Let me apply the changes in the upcoming PR.

ljhh-0611 (approved):
🫡

PR #55, opened 2026-05-06
feat: introducing docling-sys crate

jhlee525:
## Summary
- Adds a new workspace member `docling-sys` that wraps the existing `convert-pdf-to-md` PyInstaller bundle behind a Rust crate API.
- `build.rs` does what `build.sh` did (`uv sync --extra cpu` + `pyinstaller`), writing the bundle into cargo's `OUT_DIR` with content-hash caching so re-builds are no-ops.
- `lib.rs` exposes `convert_pdf_to_md(&[u8]) -> String` and `convert_pdf_file(path) -> String` (async). Internally spawns the bundled binary, pipes PDF bytes via stdin, reads markdown from stdout.
- Shipping path unchanged for now: `bundle_dir()` points into `target/`. For packaged builds, downstream copies the directory next to the final binary and calls `override_bundle_dir()`. `convert-pdf-to-md/` is left in place; we'll delete it once chat-agent is migrated off `BuiltinToolProvider::ConvertPdfToMd`.

## Why
Today the docling pipeline is reached two ways:
1. ailoy's `BuiltinToolProvider::ConvertPdfToMd` runs `uv sync` + Python on the **end user's** machine on first invocation.
2. `convert-pdf-to-md/build.sh` produces a pre-built bundle that no Rust crate currently consumes.

Neither is a clean Rust dependency. With `docling-sys`, any crate in the workspace can `docling-sys = { path = "../docling-sys" }` and call an async Rust function — no runtime `uv`, no out-of-band bundle wiring. The slow build is paid once at compile time and cached.

## Notes
- First build is several minutes (downloads CPU torch wheels + runs pyinstaller); subsequent builds skip via `OUT_DIR/.bundle-stamp`.
- `--features skip-bundle` or `DOCLING_SYS_SKIP_BUNDLE=1` short-circuits the bundle build for unrelated workspace work; the runtime API then returns an error.
- Cross-compilation isn't supported (pyinstaller produces host-arch binaries).
- After it approved, we can replace `convert-pdf-to-md` with this crate.


docling-sys/build.rs
```diff
@@ -0,0 +1,186 @@
...
+        fail("`uv` not found in PATH. Install via https://docs.astral.sh/uv/#installation");
+    }
+
+    run(
+        Command::new("uv")
+            .arg("sync")
+            .arg("--project")
+            .arg(&python_dir)
+            .arg("--extra")
+            .arg("cpu"),
```
nuri-yoo: The `--extra cpu` is hardcoded, which I assume is to avoid pulling the ~2GB CUDA wheels into the bundle. That tradeoff makes sense for Linux, but it's costing us on macOS for no real benefit: the default torch wheel on macOS already ships with MPS support and is roughly the same size as the CPU wheel. So a mac host building this gets CPU-only inference with no size win.

What about branching on host OS, with an env override for the rare cases someone wants to force a specific variant?

```rust
let extra = match env::var("DOCLING_SYS_TORCH").ok().as_deref() {
    Some("cpu") => Some("cpu"),
    Some("default") => None,
    _ if cfg!(target_os = "macos") => None, // default wheel = MPS, similar size
    _ => Some("cpu"),                        // linux/windows: keep CPU wheel
};
```

jhlee525: I changed it to optional, we'll use the one with accelerator default, uses CPU if "cpu" feature is enabled in `docling-sys` crate

nuri-yoo: LGTM

nuri-yoo (approved):

selloriwoo (approved):

grf53 (approved):

PR #56, opened 2026-05-07
feat(reflect-agent): LLM-driven reflect mode with cross-model escalation

nuri-yoo:
## Summary

This PR layers an LLM-driven reflect gate on top of the deterministic verify layer in #53. After running the calibration experiment summarised below, the design simplifies to two modes:

- `off` (default) — only the deterministic verifier from #53 runs at turn-end.
- `forced` — after every turn a Haiku reflect call examines the draft. When the first-pass verdict is a `Stop` whose self-reported confidence is below 0.7, the wrapper escalates to a stronger reflect model (Sonnet 4.6 by default) and uses the stronger model's verdict as final.

Earlier revisions of this PR shipped four modes (`off`, `self`, `forced`, `hybrid`); the calibration data described below removed the case for keeping `self` and `hybrid`, so they were dropped in 3ca3d37.

## Stacked on PR #53

Base = `feat/verify-gate`. Once #53 merges, GitHub will retarget this PR's base to `feat/reflect-agent`, then to `fix/ailoy-bump-391`, then to `refactoring-applied` as the upstream stack lands. Nothing on this branch needs to move.

## Background — the paper this is built on

Pacchiardi et al., "Trust or Escalate: LLM Judges with Provable Guarantees for Human Agreement" (ICLR 2025 Oral) — https://arxiv.org/abs/2407.18370 — argues that an LLM judge is useful when (a) it emits a verdict together with a verbalised confidence, (b) low-confidence cases route to a stronger model, and (c) the threshold is calibrated against held-out data. The verdict + confidence schema this PR uses is the same dual-signal shape Codex's `review_prompt.md` evaluator enforces; the threshold value (0.7) matches the OpenAI agentic-governance cookbook's `confidence_threshold` default for tunable LLM-based guardrails and LangGraph's `GroundingConfig` default for ReAct reliability.

## What each mode does

| Mode | Source of verdict | Action on low-confidence Stop | Retries |
|---|---|---|---|
| `off` (default) | n/a — only the deterministic verifier from #53 runs at turn-end | n/a | none |
| `forced` | Haiku reflect call after every turn | escalate to a stronger model (default Sonnet 4.6); accept its `Stop`, ignore its `Retry` | first-pass `Retry` honoured up to `RETRY_BUDGET = 1` |

The escalation model's `Retry` verdict is intentionally ignored: in the calibration data, every same-model retry that fired left the answer unchanged or made it worse. Only the stronger model's `Stop` (i.e. confirmation of the original draft) is acted on; same-model self-retry is not on the public surface.

## Calibration experiment

Held off-tree under `reflect-agent/research/` (untracked working material): a paired-trial study on a 13-task hard suite (off-by-one traps, subtle constraints, format gotchas, multi-step traps), n=10 trials per task per mode = 390 trials, agent + first-pass reflect = `claude-haiku-4-5-20251001`, escalation = `claude-sonnet-4-6`.

| Mode | Accuracy | 95% CI (Wilson) | vs. forced |
|---|---:|---:|---:|
| forced (Haiku-only reflect, no escalation) | 46.2% (60/130) | [37.8, 54.7] | baseline |
| hybrid (forced + verify-finding hints + same-model retry on low conf) | 46.2% (60/130) | [37.8, 54.7] | 0/130 paired (McNemar p = 1.0) |
| escalate (forced + Sonnet on low confidence) | 81.5% (106/130) | [74.0, 87.3] | +46/130 paired, 0 regressions (McNemar p < 0.0001) |

`hybrid` and `forced` produced an identical answer on every single (task, trial) pair — the dual mechanisms hybrid added (verify-finding hint injection + low-confidence promotion to retry) had zero detectable effect at this sample size. A separate paired ablation against escalate (n=5 = 65 trials) confirmed the verify-finding hint contributes ≤ 1 trial out of 65 (McNemar p = 1.0), so the escalation path doesn't need the hint either; Sonnet's raw strength on the same draft carries the gain.

The full report — methodology, calibration plot, threshold sweep, qualitative case examples, conclusion — is preserved as `reflect-agent/research/calibration-report.html` (also archived locally outside the repo). It is not part of this commit's diff but is the source of the four design decisions in this PR:

1. drop `Self_` (self-discipline triggered verify tool) — peer-reviewed work on agentic instruction following (e.g. AgentIF) shows ≤ 30% full-compliance even with explicit "MUST" prompts; without measurement the mode is too unpredictable to keep.
2. drop `Hybrid` — null effect in our data.
3. keep escalation but drop the verify-finding hints into the reflect call — the n=65 paired ablation found no marginal effect.
4. ignore the escalation model's `Retry` verdict — when it fired in our data, the resulting same-model retry never recovered the answer.

## Public surface (after this PR)

- `ReflectMode = Off | Forced` (was `Off | Self_ | Forced | Hybrid`)
- `run_with_forced_reflect(agent, query, verify_config, provider, reflect_model, escalate_model)` — `escalate_model` is the new parameter
- `ForcedReflectOutcome { outputs, verify_report, reflect_verdicts, retry_count, escalations }` — `escalations` is new (count of times the stronger model was invoked); `low_confidence_promotions` is gone
- `LOW_CONFIDENCE_THRESHOLD: f32 = 0.7` — replaces `HYBRID_LOW_CONFIDENCE_THRESHOLD`
- `DEFAULT_ESCALATE_MODEL: &str = "anthropic/claude-sonnet-4-6"` — new
- `DEFAULT_REFLECT_MODEL`, `RETRY_BUDGET`, `reflect_call`, `ReflectVerdict` — unchanged
- gone: `ReflectMode::Self_`, `ReflectMode::Hybrid`, `run_with_hybrid`, `HybridReflectOutcome`, `HYBRID_LOW_CONFIDENCE_THRESHOLD`, `render_verify_hints`, the `verify` tool factory, the self-mode system prompt

## CLI

```
reflect-agent [OPTIONS]

  --model <MODEL>            agent model           [default: anthropic/claude-haiku-4-5-20251001]
  --query <QUERY>            single-shot, exit
  --reflect-mode <MODE>      off | forced          [default: off]
  --reflect-model <MODEL>    first-pass reflect    [default: anthropic/claude-haiku-4-5-20251001]
  --escalate-model <MODEL>   stronger reflect      [default: anthropic/claude-sonnet-4-6]
  --verbose                  show verify report in forced mode
```

CLI breaking change: `--reflect-mode self` and `--reflect-mode hybrid` no longer parse. The two replaced flows have been measured against the surviving one and the data does not justify them; downstream invocations that hard-coded those values will need to switch to `forced` (or `off`).

## Reflect-call system prompt

Unchanged from the previous revision — both Haiku and Sonnet reflect calls receive the same prompt:

> You are a verification assistant. Read the draft answer below and decide whether it can be sent to the user as-is.
>
> Output a single JSON object with these fields:
>   - verdict:    "stop" or "retry"
>   - rationale:  one short sentence explaining the choice
>   - next_query: only when verdict is "retry"; a re-posed task that fixes the issue
>   - confidence: a number in [0.0, 1.0] expressing how sure you are of the verdict
>
> Choose "stop" when the draft answers the user's question correctly and is self-consistent with the cited evidence. Choose "retry" when the draft contains errors, leaves the user's question unanswered, or cites sources that do not exist in the trajectory. Use a confidence below 0.7 only when you genuinely cannot tell whether the draft is right. Output JSON only — no prose around it.

## Verdict format

```json
{ "verdict": "stop" | "retry",
  "rationale": "...",
  "next_query": "...",   // only when retry
  "confidence": 0.92 }    // optional, [0.0, 1.0]
```

`parse_verdict` is fail-open: missing fields, surrounding prose, mixed casing, retry-without-next_query, and out-of-range confidence all coerce sensibly (mostly to a synthetic `Stop`) so a malformed verifier response never aborts a real user turn.

## Verification

```
cargo check --workspace --tests --all-features         # clean
cargo test  -p reflect-agent --lib                      # passes
cargo test  -p reflect-agent --test demo_task           # 2 passed; 1 ignored
cargo test  -p reflect-agent --test reflect_modes       # 2 ignored (need API key)
cargo test  -p speedwagon  --lib --all-features         # no regression
```

## Test plan

- [x] `cargo check --workspace --tests --all-features` is clean
- [x] `cargo test -p reflect-agent --lib` passes
- [x] `cargo test -p reflect-agent --test demo_task` 2/0/1 (one ignored, needs API key)
- [x] `cargo test -p reflect-agent --test reflect_modes` compiles two `#[ignore]`d e2e tests (one per mode)
- [x] No regression on `speedwagon`: 71/0/2 still green
- [x] Manual smoke: `--reflect-mode forced` on a Haiku-easy query (`What is 17 * 23?`) — Haiku reflect returns Stop@0.99, no escalation
- [x] Manual smoke: `--reflect-mode forced` on a Haiku-uncertain query (Fibonacci F(15)) — Haiku Stop@0.5, escalates to Sonnet, Sonnet Stop@0.85, draft accepted

### DIFF
--- a/reflect-agent/src/reflect.rs
+++ b/reflect-agent/src/reflect.rs
@@ -1,0 +1,490 @@
+//! LLM-driven reflect gate (Phase 2 of the reflect-agent design).
+//!
+//! Two modes are exposed:
+//!
+//! - **`Self_`** — the agent's `bash` / `python_repl` / `web_search` tool set
+//!   gains a fourth tool, `verify`, and the system prompt instructs the LLM
+//!   to call it before emitting the final answer. The LLM decides when to
+//!   invoke the verify pass; if it doesn't, the gate is bypassed.
+//!
+//! - **`Forced`** — `Agent::run` is wrapped on the outside. After the agent
+//!   finishes a turn, the wrapper unconditionally runs a separate LLM call
+//!   on the draft answer and gets a [`ReflectVerdict`] back. On `Retry`,
+//!   the wrapper re-invokes the agent with `next_query`. The retry budget
+//!   is fixed at one re-attempt per user turn.
+//!
+//! Both modes share [`reflect_call`] — the JSON-shaped LLM critique that
+//! emits a `stop` / `retry` verdict — and [`ReflectVerdict`] parsing.
+
+use ailoy::{
+    agent::AgentProvider,
+    lang_model::LangModel,
+    message::{Message, Part, Role},
+};
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+
+/// Default LLM used by the reflect call. Cheap enough that one extra call
+/// per turn is acceptable; same model family as the main agent so that the
+/// global provider already covers it.
+pub const DEFAULT_REFLECT_MODEL: &str = "anthropic/claude-haiku-4-5-20251001";
+
+/// Maximum number of `Retry` verdicts honoured per user turn. The second
+/// `Retry` is silently coerced to `Stop`. Mirrors the budget the rest of
+/// the verify-gate work uses.
+pub const RETRY_BUDGET: usize = 1;
+
+/// Which reflect strategy to apply to a turn.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub enum ReflectMode {
+    /// No LLM-driven reflect. The deterministic [`crate::verify_run`] pass
+    /// from PR #53 still runs; this mode is what the CLI defaults to.
+    #[default]
+    Off,
+    /// `verify` tool registered on the agent + system prompt instructs the
+    /// LLM to call it before final emission. LLM-self-discipline driven.
+    Self_,
+    /// Wrapper around `Agent::run` that unconditionally runs [`reflect_call`]
+    /// on the draft answer after the turn finishes. Strict but blind to
+    /// what the deterministic verify pass already noticed.
+    Forced,
+    /// Forced behaviour plus the deterministic verifier's findings get
+    /// passed into the reflect call as hints, and a low-confidence `Stop`
+    /// (below [`HYBRID_LOW_CONFIDENCE_THRESHOLD`]) is treated as a
+    /// retry within the standard budget. The two layers (deterministic
+    /// verify + LLM reflect) work together rather than in parallel.
+    Hybrid,
+}
+
+impl ReflectMode {
+    /// Lower-case CLI label.
+    pub fn as_str(self) -> &'static str {
+        match self {
+            ReflectMode::Off => "off",
+            ReflectMode::Self_ => "self",
+            ReflectMode::Forced => "forced",
+            ReflectMode::Hybrid => "hybrid",
+        }
+    }
+
+    /// Parse the `--reflect-mode` CLI value.
+    pub fn parse(s: &str) -> Result<Self> {
+        match s {
+            "off" => Ok(ReflectMode::Off),
+            "self" => Ok(ReflectMode::Self_),
+            "forced" => Ok(ReflectMode::Forced),
+            "hybrid" => Ok(ReflectMode::Hybrid),
+            other => anyhow::bail!(
+                "unknown reflect mode: {other:?} (expected off|self|forced|hybrid)"
+            ),
+        }
+    }
+}
+
+/// Outcome of one reflect call. `Stop` means the draft can be returned to
+/// the user; `Retry` carries a re-posed query for the agent to run again.
+///
+/// `confidence` is an `Option<f32>` in `[0.0, 1.0]`. The verifier LLM is
+/// asked to emit it but is allowed to omit it — older verdicts produced
+/// before the field was introduced still parse cleanly. Hybrid mode reads
+/// it to decide whether a `Stop` verdict is strong enough to honour;
+/// other modes treat low-confidence stops the same as high-confidence
+/// ones.
+#[derive(Clone, Debug, PartialEq)]
+pub enum ReflectVerdict {
+    Stop {
+        rationale: String,
+        confidence: Option<f32>,
+    },
+    Retry {
+        rationale: String,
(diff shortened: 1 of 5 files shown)


reflect-agent/src/reflect.rs
```diff
@@ -0,0 +1,490 @@
...
+        }
+    }
+}
+
+/// Threshold under which Hybrid mode treats a `Stop` verdict as "not
+/// confident enough" and forces a retry (within the standard retry
+/// budget). Picked at 0.7 to match the loose convention in Codex's
+/// `review_prompt.md` evaluator: low-confidence assessments shouldn't
+/// be allowed to silently end the turn. Other modes ignore this.
+pub const HYBRID_LOW_CONFIDENCE_THRESHOLD: f32 = 0.7;
```
grf53: I found this comment as the basis for determining the threshold.

It does use a confidence_score between 0 and 1 in review_prompt.md. Is the codex using a value of 0.7? I couldn't find any clear justification for this.

(I am not demanding a clear rationale for the threshold value. I simply want to understand the utility of the hybrid methodology more clearly.)

nuri-yoo: where 0.7 came from : answered in comment below. 
Hybrid Utility:  It uses very findings injected as reflect hints. Deterministic
verify runs in every mode, but hybrid is the only mode where  those findings reach the reflect LLM as explicit hints. 

For now, all mode are available for further decision.

grf53:
It seems that the introduction of escalation enables the goal to be achieved more successfully.

The 'Trust-or-Escalate' paper appears to have theoretically placed no limit on the number of tier levels, and since it seems to have used a total of three tiers experimentally, we could consider more levels when actually applying it to a service. (I think It's up to your decision.)

In the current form, there are only two tiers for escalation, and this alone appears to be sufficient. If we intend to maintain this, the current implementation seems adequate.

grf53 (approved):
Look good enough to be merged. Thanks!

jhlee525 (approved):

PR #57, opened 2026-05-07
ADD: batch document ingest + bulk purge with partial success

ljhh-0611:
## Summary

This PR adds batch-oriented document materialization APIs on top of the Speedwagon store, with partial-success semantics for both ingest and purge. The goal is to let clients upload or delete multiple documents in one request while still preserving successful work when individual files or ids fail.

It builds on the Speedwagon-backed session/message flow from #48 by exposing the document corpus through HTTP endpoints and covering the multi-document path in backend integration tests.

### Speedwagon Store
- Add `Store::ingest_many` returning `IngestResult { succeeded, failed }`.
  - Valid documents are accumulated and written to the index in one batch.
  - Per-file failures are recorded with the original item index.
  - Newly written intermediate corpus/origin files are cleaned up on failure where possible.
  - Duplicate documents are treated as successful and return the existing deterministic UUID.
- Add `Store::purge_many` returning `PurgeResult { purged, failed }`.
  - Missing ids are reported as per-item failures instead of failing the whole request.
- Update the preset corpus path to use the new batch ingest result shape.

### Backend Document API
- Add document DTOs in `backend-v2/src/model/document.rs`:
  - `DocumentResponse`
  - `BatchIngestResponse`
  - `BatchPurgeResponse`
  - `FailedItem`
  - `BulkPurgeRequest`
- Extend backend routing with document endpoints:
  - `GET /documents` — list indexed documents
  - `POST /documents` — multipart multi-file ingest
  - `GET /documents/{id}` — fetch one document by id
  - `DELETE /documents/{id}` — purge one document by id
  - `DELETE /documents` — bulk purge via JSON body
- Add per-file multipart validation and file-type mapping for `pdf`, `md`, `markdown`, and `txt`.
- Return `201 Created` when a batch ingest has no failures, and `200 OK` when the request completed with partial failures.
- Reject unsupported file types and empty bulk purge requests with client-facing errors instead of store-level panics.

### Tests
- Add a dedicated `document_test` integration suite covering:
  - empty document lists
  - single ingest/list/get/purge
  - duplicate ingest behavior
  - multi-file ingest
  - mixed success/failure ingest
  - unsupported file handling
  - bulk purge success and partial failure
  - missing document behavior
- Expand `tests/common` with reusable document API helpers.
- Rewrite the ignored E2E flow to ingest two documents over HTTP, ask the agent about both facts, bulk purge them, verify the list is empty, then confirm the session still responds after purge.

### Batch ingest

```http
POST /documents
Content-Type: multipart/form-data
```

```json
{
  "succeeded": [
    { "id": "...", "title": "...", "len": 123 }
  ],
  "failed": [
    { "name": "bad.exe", "error": "unsupported file type '.exe' — supported: pdf, md, txt" }
  ]
}
```

### Bulk purge

```http
DELETE /documents
Content-Type: application/json
```

```json
{ "ids": ["..."] }
```

```json
{
  "purged": ["..."],
  "failed": [
    { "name": "...", "error": "document not found" }
  ]
}
```

## Validation

- [x] `cargo fmt --check -p agent-k-backend`
- [x] `cargo check -p agent-k-backend`
- [x] `cargo test -p agent-k-backend --test document_test` — 14 passed
- [x] `cargo test -p agent-k-backend --test e2e_test -- --ignored --list` — confirms the ignored live E2E test is present

## Notes / Out of Scope

- The live E2E test remains `#[ignore]` because it requires `OPENAI_API_KEY` and performs an actual model-backed RAG round trip.
- **Temporary API / Store contract warning:** `ingest_many` currently follows the existing single-document `Store` behavior as closely as possible, but this is an initial API shape rather than a finalized long-term contract. As the Store is hardened (metadata generation, indexing strategy, failure recovery, corpus ownership, and migration behavior), the batch ingest API and its response/error semantics are expected to evolve with it.
- This PR keeps the current single shared Speedwagon store design from #48; multi-tenant or per-user corpus ownership should be handled in a later API/auth integration pass.
- Multipart upload currently accepts the supported document extensions only (`pdf`, `md`, `markdown`, `txt`). Additional formats should be added through the shared Speedwagon `FileType`/translator path.

### DIFF
--- a/backend-v2/src/handlers/document.rs
+++ b/backend-v2/src/handlers/document.rs
@@ -1,0 +1,187 @@
+use std::sync::Arc;
+
+use axum::{
+    Json,
+    extract::{Multipart, Path, Query, State},
+    http::StatusCode,
+};
+use serde::Deserialize;
+use speedwagon::FileType;
+use uuid::Uuid;
+
+use crate::{
+    error::{ApiResult, AppError},
+    model::{
+        BatchIngestResponse, BatchPurgeResponse, BulkPurgeRequest, DocumentResponse, FailedItem,
+    },
+    state::AppState,
+};
+
+#[derive(Debug, Deserialize)]
+pub struct ListDocumentsQuery {
+    #[serde(default)]
+    pub page: Option<u32>,
+    #[serde(default)]
+    pub page_size: Option<u32>,
+}
+
+pub async fn list_documents(
+    State(state): State<Arc<AppState>>,
+    Query(query): Query<ListDocumentsQuery>,
+) -> ApiResult<Json<Vec<DocumentResponse>>> {
+    let page = query.page.unwrap_or(0);
+    let page_size = query.page_size.unwrap_or(50);
+
+    let store = state.store.read().await;
+    let docs = store
+        .list(false, page, page_size)
+        .map_err(|e| AppError::internal(e.to_string()))?;
+
+    Ok(Json(docs.into_iter().map(DocumentResponse::from).collect()))
+}
+
+pub async fn get_document(
+    State(state): State<Arc<AppState>>,
+    Path(id): Path<Uuid>,
+) -> ApiResult<Json<DocumentResponse>> {
+    let store = state.store.read().await;
+    match store.get(id) {
+        Some(doc) => Ok(Json(DocumentResponse::from(doc))),
+        None => Err(AppError::not_found("document not found")),
+    }
+}
+
+fn parse_filetype(filename: &str) -> Result<FileType, String> {
+    let ext = filename.rsplit('.').next().unwrap_or("").to_lowercase();
+    match ext.as_str() {
+        "pdf" => Ok(FileType::PDF),
+        "md" | "markdown" | "txt" => Ok(FileType::MD),
+        _ => Err(format!(
+            "unsupported file type '.{ext}' — supported: pdf, md, txt"
+        )),
+    }
+}
+
+pub async fn ingest_document(
+    State(state): State<Arc<AppState>>,
+    mut multipart: Multipart,
+) -> ApiResult<(StatusCode, Json<BatchIngestResponse>)> {
+    let mut valid_items: Vec<(Vec<u8>, FileType)> = Vec::new();
+    let mut filenames: Vec<String> = Vec::new();
+    let mut failed: Vec<FailedItem> = Vec::new();
+
+    while let Ok(Some(field)) = multipart.next_field().await {
+        if field.name() != Some("file") {
+            continue;
+        }
+        let filename = field.file_name().unwrap_or("upload").to_string();
+        let bytes = field
+            .bytes()
+            .await
+            .map_err(|e| AppError::internal(format!("failed to read upload: {e}")))?;
+
+        match parse_filetype(&filename) {
+            Ok(filetype) => {
+                valid_items.push((bytes.to_vec(), filetype));
+                filenames.push(filename);
+            }
+            Err(e) => {
+                failed.push(FailedItem {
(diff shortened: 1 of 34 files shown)


speedwagon/src/store/mod.rs
```diff
@@ -92,64 +124,117 @@ impl Store {
...
+            match indexer::document_exists(&self.index, &id.to_string()) {
+                Ok(true) => {
+                    succeeded.push(id);
+                }
+                Ok(false) => match fs::read_to_string(&corpus_path) {
+                    Ok(content) => {
+                        to_index.push((idx, id, content));
+                    }
+                    Err(e) => {
+                        remove_ingest_artifact(&corpus_path);
```
khj809: This line may remove previously existing corpus file if `fs::read_to_string` fails. Suggest to add a flag to check if corpus file has been newly created, and call `remove_ingest_artifact` only if the flag is set.
```rust
let corpus_was_new = !corpus_path.exists();
// ...
if corpu_was_new {
    remove_ingest_artifact(&corpus_path);
}
```

speedwagon/src/store/mod.rs
```diff
@@ -92,64 +124,117 @@ impl Store {
...
         let mut docs: Vec<(String, String, String)> = Vec::with_capacity(to_index.len());
-        for (id, content) in to_index {
-            let title = parser::get_title(&content).await?;
-            docs.push((id.to_string(), title, content));
+        for (idx, id, content) in to_index {
+            match parser::get_title(&content).await {
+                Ok(title) => docs.push((id.to_string(), title, content)),
+                Err(e) => {
+                    let corpus_path = self.root.join("corpus").join(format!("{id}.md"));
+                    remove_ingest_artifact(&corpus_path);
```
khj809: Similar as before if `parser::get_title` fails.

backend-v2/src/model/document.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub purged: Vec<String>,
+    pub failed: Vec<FailedItem>,
+}
+
+// --- Request DTOs ---
+
+#[derive(Clone, Debug, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct BulkPurgeRequest {
+    pub ids: Vec<Uuid>,
```
khj809: The `DocumentResponse.id` is `String` but here `ids` is `Vec<Uuid>`. Would be better to unify them to either `Uuid` or `String`.

backend-v2/src/handlers/document.rs
```diff
@@ -0,0 +1,189 @@
...
+
+pub async fn ingest_document(
+    State(state): State<Arc<AppState>>,
+    NoApi(mut multipart): NoApi<Multipart>,
+) -> ApiResult<(StatusCode, Json<BatchIngestResponse>)> {
+    let mut valid_items: Vec<(Vec<u8>, FileType)> = Vec::new();
+    let mut filenames: Vec<String> = Vec::new();
+    let mut failed: Vec<FailedItem> = Vec::new();
+
+    while let Ok(Some(field)) = multipart.next_field().await {
```
khj809: This loop may be silently terminated if `multipart.next_field().await` returned `Err`. Suggest to fix it as below, so API can explicitly respond with 400 error:
```suggestion
    while let Some(field) = multipart.next_field().await
        .map_err(|e| AppError::bad_request(format!("multipart error: {e}")))? {
```

ljhh-0611: @khj809  I modified all of your suggestion! Thank you for your great insight!

khj809 (approved):
LGTM!

PR #62, opened 2026-05-07
feat: add Project workspace and session sharing model

khj809:
## Summary

Introduces a **Project workspace** concept to a backend that previously had no ownership or authentication on sessions. Sessions are now scoped under projects, with per-session sharing controls and JWT auth enforced across all routes.

### Design Decisions

| Item | Decision |
|---|---|
| Project access | Owner + Member (single tier). Only the owner can invite/remove members, edit metadata, or delete the project |
| Session sharing | Per-session `share_mode`: `private` / `shared_readonly` / `shared_chat`. Only the session creator can change it |
| Chat lock | SSE stream automatically holds `tokio::sync::Mutex::try_lock_owned()` for its lifetime → 423 Locked on contention |
| Default project | `"Personal"` project auto-created on signup and admin bootstrap |
| Session visibility | Other members' `private` sessions return 404 (existence hidden, not just forbidden) |
| Existing data | Dev stage — `sessions` / `session_messages` dropped and recreated |

### Changes

**DB (`migrations/0003_projects_and_sessions.sql`)**
- Drop and recreate `sessions`, `session_messages`
- New `projects` table (`owner_id` FK → users, ON DELETE RESTRICT)
- New `project_members` table (composite PK)
- Rebuilt `sessions` with `project_id`, `creator_id`, `share_mode` columns

**Repository (`src/repository/`)**
- New types: `ShareMode`, `DbProject`, `DbProjectMember`, `SessionAccess`
- `DbSession` expanded to 6 fields
- `get_session_with_authz()` — SQL CASE expression computes access level via project membership JOIN
- `list_sessions_in_project()` — visibility-filtered list
- `update_session_share_mode()`, full project/member CRUD
- `create_user_with_personal_project()` — creates user then Personal project sequentially
- `delete_user()` — pre-deletes owned projects/sessions to satisfy RESTRICT FK before user deletion

**Handlers / Router**
- New `src/handlers/project.rs` — 8 handlers for `/projects/*`
- Full rewrite of `src/handlers/session.rs` — JWT `Extension(AuthUser)`, `try_lock_owned()` lock, `get_session_with_authz`-based authorization
- `auth_required` middleware applied to all session and project routes
- `send_message_stream` truncates in-memory history on error to keep DB consistent
- `clear_message_history` acquires agent lock before DB + memory clear to prevent race

**Auth**
- `signup` calls `create_user_with_personal_project`
- `bootstrap.rs` admin creation calls `create_user_with_personal_project`

**Tests**
- All existing integration tests updated to use JWT tokens and project-scoped session creation
- New `tests/project_test.rs` (6 tests) — Personal project on signup, member invite/remove, owner leave blocked, cascade delete
- New `tests/session_authz_test.rs` (5 tests) — private session hiding, read-only enforcement, creator-only share_mode, project session list visibility

### Permission Matrix

| Action | Non-member | ReadOnly member | Chat member | Creator |
|---|---|---|---|---|
| GET session / messages | 404 | 200 | 200 | 200 |
| POST messages | 404 | 403 | 200 / 423 if locked | 200 / 423 if locked |
| PATCH share_mode | 404 | 403 | 403 | 200 |
| DELETE session / messages | 404 | 403 | 403 | 200 |

## Test plan

- [x] `cargo build` — clean, no warnings
- [x] `cargo test` — 51 passed, 0 failed (7 ignored: require AI provider / microsandbox)
  - lib unit tests: 8 passed
  - auth_test: 22 passed (includes Personal project auto-creation check)
  - message_history_persistence: 9 passed
  - project_test: 6 passed
  - session_authz_test: 5 passed
  - sandbox_per_session: 1 passed


backend-v2/src/repository/session.rs
```diff
@@ -0,0 +1,296 @@
...
+             FROM sessions s
+             JOIN projects p ON p.id = s.project_id
+             LEFT JOIN project_members pm
+                   ON pm.project_id = s.project_id AND pm.user_id = ?1
+             WHERE s.id = ?2",
+        )
+        .bind(&uid)
+        .bind(&sid)
+        .fetch_optional(&self.pool)
+        .await?;
```
ljhh-0611, line 155: Sessions are project-dependent, so should we use admin instead of creator?

Also, in get_session_with_authz, only the creator gets special treatment. We might need to clarify whether the project owner should have the same permissions as regular users for sessions case.

backend-v2/src/handlers/project.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    state
+        .repository
+        .delete_project(project_id)
+        .await
+        .map_err(|e| AppError::internal(e.to_string()))?;
+
+    tracing::info!(id = %project_id, "project deleted");
+    Ok(StatusCode::NO_CONTENT)
+}
```
ljhh-0611, line 124: When a project is deleted, sessions are cascaded and removed from the database.
However, the cleanup logic in the existing session handler—agent deletion and sandbox cleanup—won't be executed.
We should make sure these cleanups are also triggered in the cascade.

backend-v2/src/handlers/session.rs
```diff
@@ -126,184 +137,302 @@ pub async fn create_session(
     let sandbox = Sandbox::new(cfg)
         .await
         .map_err(|e| AppError::internal(e.to_string()))?;
-
     let agent = build_agent(sandbox)
         .await
         .map_err(|e| AppError::internal(e))?;
+    state.insert_agent(session.id, agent);
```
ljhh-0611, line 152: This is a pre-existing issue, but I think we need to add atomicity to both create and delete session operations.

All three components—agent, sandbox, and session_db—need to be created/deleted together without partial failures. What would be the best approach?

backend-v2/src/repository/session.rs
ljhh-0611: One more thing: there's an authorization inconsistency where creators retain direct session access even after being removed from project membership.

It's probably not a practical issue since the project won't be visible on the real app, but still, it's worth reviewing.

khj809: I understood your concern. If project owner can see my private sessions, there's no meaning of "private". It would be more proper to limit the permission of project owner to project's attribute level (e.g. name, description, invited users, etc.), and apply the same session sharing policy even for project owner. Would it make you agreed?

backend-v2/src/handlers/session.rs
khj809: That makes sense. I'll fix to handle them atomically.

backend-v2/src/repository/session.rs
ljhh-0611: > I understood your concern. If project owner can see my private sessions, there's no meaning of "private". It would be more proper to limit the permission of project owner to project's attribute level (e.g. name, description, invited users, etc.), and apply the same session sharing policy even for project owner. Would it make you agreed?

Oh, I forgot about "private" session :) In case of private session, the access level is reasonable.

I just wonder whether project owner can administrate "shared_sessions" or not

khj809: > One more thing: there's an authorization inconsistency where creators retain direct session access even after being removed from project membership.
> 
> It's probably not a practical issue since the project won't be visible on the real app, but still, it's worth reviewing.

Oh I couldn't see this comment.. 
After I see this edge case, I've changed my mind to let owner be able to see every sessions including private one. Otherwise, the left user's private sessions will become invisible orphans, and there's no way to cleanup them until the project is deleted. At least we should give owners a last resort to remove such orphans.

In this perspective,
> I just wonder whether project owner can administrate "shared_sessions" or not

I think owner should be able to administrate shared (and even private) sessions in project.
How do you think about it?

ljhh-0611: For first version, I agree that it's enough to get administrate permission for all of sessions in project to owner 👍

khj809: @ljhh-0611 I've changed the permission `SessionAccess::Creator` to `SessionAccess::Admin` which includes project owners too, and admin can see every sessions in project.
Also, the members removed from project cannot see their sessions anymore.

ljhh-0611 (approved):
🙌

The below content is a just note future consideration from ai review

> As a follow-up, please also review admin self/last-admin lockout guards. Self-delete is blocked, but self-demotion/self-deactivation and last-active-admin demotion/deactivation do not appear to be guarded. This may be pre-existing rather than introduced by this PR, but the new user/project/session ownership model increases the impact of user lifecycle changes.

PR #66, opened 2026-05-11
fix: guard admin self-demotion, self-deactivation, and last-admin removal

khj809:
## Summary

- **Self-demotion blocked**: `PATCH /admin/users/:id` with `role: "user"` (or any non-admin role) on the caller's own account now returns `400 Bad Request`.
- **Self-deactivation blocked**: `PATCH /admin/users/:id` with `is_active: false` on the caller's own account now returns `400 Bad Request`.
- **Last active admin protected**: Demoting or deactivating a user whose removal would leave zero active admins is now blocked with `400 Bad Request`.

Previously, `update_user_admin` did not capture `auth` (the extension was `_auth`), so no self-checks were possible. The handler now captures `auth` and enforces all three guards before touching the database.

## What was vulnerable

```rust
// before: auth user discarded, no guards
pub async fn update_user_admin(
    Extension(_auth): Extension<AuthUser>,
    ...
) -> ApiResult<...> {
    // straight through to DB update
}
```

## Tests added (`tests/auth_test.rs`)

| Test | Scenario | Expected |
|------|----------|----------|
| `admin_cannot_demote_self` | Admin PATCHes own role → `user` | `400` |
| `admin_cannot_deactivate_self` | Admin PATCHes own `is_active` → `false` | `400` |
| `last_admin_cannot_be_demoted_by_another_admin` | Sole active admin tries to demote themselves | `400` |

## Test plan

- [x] `cargo test -p agent-k-backend --test auth_test` — all 25 tests pass
- [x] Existing `admin_can_update_user_role` and `admin_can_deactivate_user` still pass (non-self operations unaffected)
- [x] `admin_cannot_delete_self` still passes (pre-existing guard untouched)

### DIFF
--- a/backend-v2/src/handlers/user.rs
+++ b/backend-v2/src/handlers/user.rs
@@ -148,10 +148,39 @@ pub async fn get_user_admin(
 
 pub async fn update_user_admin(
     State(state): State<Arc<AppState>>,
-    Extension(_auth): Extension<AuthUser>,
+    Extension(auth): Extension<AuthUser>,
     Path(id): Path<Uuid>,
     Json(payload): Json<AdminUpdateUserRequest>,
 ) -> ApiResult<Json<UserResponse>> {
+    let removes_admin_access = matches!(&payload.role, Some(r) if *r != Role::Admin)
+        || matches!(payload.is_active, Some(false));
+
+    if auth.id == id {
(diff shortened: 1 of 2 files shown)


ljhh-0611 (approved):
🫡

PR #67, opened 2026-05-11
feat: project-scoped file upload (dirent API) + sandbox readonly mount

khj809:
## Summary

- **New dirent API** — project-scoped file upload/download/list/delete at `/projects/{project_id}/dirents`. Files are stored at `data/projects/<pid>/uploads/` and preserve the uploaded directory hierarchy.
- **Sandbox mount** — every session's agent sandbox gets the project's `uploads/` directory bind-mounted at `/workspace/.uploads` (readonly). `/workspace` itself remains read-write so the agent can freely create files for intermediate work.
- **Project cleanup** — `DELETE /projects/{id}` now removes the project's `data/projects/<pid>/` directory after the DB delete succeeds.
- **Legacy preserved** — existing `/documents*` routes, `speedwagon::Store`, and tantivy indexing are untouched.

## New Endpoints

| Method | Path | Description |
|--------|------|-------------|
| `POST` | `/projects/{project_id}/dirents` | multipart upload; relative paths in `filename`; atomic write; 50 MB per-file limit (env `AGENT_K_MAX_UPLOAD_BYTES`) |
| `GET` | `/projects/{project_id}/dirents` | BFS listing with optional `?prefix=` and `?recursive=` |
| `GET` | `/projects/{project_id}/dirents/{*path}` | single file download with `Content-Type` via `mime_guess` |
| `DELETE` | `/projects/{project_id}/dirents/{*path}` | delete file or recursive directory |

All endpoints require `auth_required` + project membership (`user_in_project`).

## Security

- `safe_join` rejects empty paths, absolute paths, `..` segments, NUL bytes, and Windows drive prefixes.
- `symlink_metadata` used in `get_file` / `delete_path` — symlinks are rejected (404) to prevent escape from `uploads_root`.
- BFS `list` uses `entry.file_type()` (no symlink follow) and skips symlinks.
- Upload reads multipart fields chunk-by-chunk; size limit (`AGENT_K_MAX_UPLOAD_BYTES`, default 50 MB) is enforced during streaming — server never buffers more than one chunk past the limit. Oversized files land in `failed[]` and remaining fields continue processing.
- `DefaultBodyLimit::disable()` scoped to POST upload only; GET list retains Axum's default.
- Atomic writes: `.tmp.<uuid>` → `rename`; write/rename failures are per-file (pushed to `failed[]`, other files in the same request continue). Orphaned temp files are cleaned up with a warn log on cleanup failure.
- `?prefix=` filter uses path-component boundary matching (`prefix=src` matches `src/` but not `src2/`).
- `AGENT_K_MAX_UPLOAD_BYTES` is read once at `AppState` initialization, not on every request.

## Agent System Prompt

Updated to document:
- `/workspace` — read-write working directory for the agent
- `/workspace/.uploads` — readonly view of project-uploaded files; new uploads appear immediately (live bind mount)

## Disk Layout

```
data/
└── projects/
    └── <project_uuid>/
        └── uploads/
            ├── README.md
            └── src/
                └── foo.md
```

## Test Plan

**Automated (tests/dirent_test.rs):**
- [x] `upload_and_list_files` — multi-file upload → `succeeded`/`failed` split; both files appear in listing
- [x] `path_traversal_in_upload_goes_to_failed` — `../etc/passwd`-style filename → appears in `failed`, no disk trace
- [x] `upload_over_size_limit_goes_to_failed` — oversized file → `failed[]`
- [x] `get_file_returns_bytes_and_content_type` — downloaded bytes match uploaded content
- [x] `delete_file_removes_it` — DELETE 204; subsequent GET 404
- [x] `delete_directory_is_recursive` — deleting `src/` removes nested files; sibling files unaffected
- [x] `non_member_gets_403_on_all_endpoints` — all 4 endpoints return 403 for non-member
- [x] `list_prefix_filter_uses_path_component_boundary` — `?prefix=src` matches `src/` but not `src2/`

**Manual:**
- [x] Session agent can `ls /workspace/.uploads` and read uploaded files
- [x] Files uploaded after session creation appear in sandbox immediately (live bind mount)
- [x] `touch /workspace/.uploads/x` in sandbox → permission denied
- [x] `touch /workspace/scratch.txt` in sandbox → succeeds
- [x] `DELETE /projects/{id}` removes `data/projects/{id}/` directory


backend-v2/src/handlers/dirent.rs
```diff
@@ -0,0 +1,432 @@
...
+        let mut over_limit = false;
+        loop {
+            match field
+                .chunk()
+                .await
+                .map_err(|e| AppError::bad_request(format!("multipart error: {e}")))?
+            {
+                Some(chunk) => {
+                    data.extend_from_slice(&chunk);
+                    if data.len() > max_bytes {
```
ljhh-0611: The current upload path still buffers each accepted file into data: Vec<u8> while checking data.len().
That means a valid near-limit file is fully resident in memory before it is written to disk.
Could we stream each chunk directly into the temp file while only tracking byte_count for the size limit?

backend-v2/src/handlers/project.rs
```diff
@@ -119,6 +119,17 @@ pub async fn delete_project(
...
+    // Best-effort cleanup after DB delete; ignore NotFound (project may never have had uploads)
+    let project_dir = state
+        .data_root
+        .join("projects")
+        .join(project_id.to_string());
+    if let Err(e) = tokio::fs::remove_dir_all(&project_dir).await {
+        if e.kind() != std::io::ErrorKind::NotFound {
+            tracing::warn!(id = %project_id, "failed to remove project dir: {e}");
+        }
+    }
```
ljhh-0611, line 131: Is it okay to best-effort procedure after db deleted?
How about before 116 line..?

backend-v2/src/handlers/dirent.rs
khj809: Fixed. Replaced the Vec<u8> buffer with streaming: each chunk is now written directly to the temp file via AsyncWriteExt::write_all, and only byte_count: u64 is tracked for the size limit. The over-limit path still drains remaining chunks to keep the multipart stream parseable, then removes the partial temp file before continuing to the next field.

backend-v2/src/handlers/project.rs
khj809: The current order is intentional. FS cleanup is genuinely best-effort — it can fail silently without breaking correctness. If we move it before the DB delete and the DB delete subsequently fails, we end up with the DB record intact but the files already gone. That state is harder to reason about (project exists in the API but has no files). With the current order, the DB delete is the authoritative tombstone: once it succeeds the project is gone from the API regardless of what happens to the files. A failed FS cleanup leaves orphaned files that are invisible to the API, which is a recoverable leak rather than a consistency issue.

ljhh-0611 (approved):

PR #69, opened 2026-05-12
refactor: replace inline docling Python with docling-sys

selloriwoo:
## Summary
- Replace the inline Python docling script (with on-demand `uv` venv) in `speedwagon/src/store/translator/pdf.rs` with a call into the `docling-sys` crate.
- Make `translate()` / `translate_pdf()` async; add `.await` at the two callers in `store/mod.rs` (`ingest`, `ingest_many`).
- Add a FinanceBench-based round-trip test for the PDF translator path. 

## Tests
- Convert the two existing HTML/MD translator tests from `#[test]` to `#[tokio::test]` (now that `translate` is async).
- New `translate_pdf_from_financebench`:
  - Gated behind the `internal` feature, matching the existing `test_ingest_financebench_samples` pattern in `store/mod.rs`.
  - Marked `#[ignore = "requires docling bundle & network access"]`.
  - Pulls FinanceBench[0], writes it under a `tempdir`, runs `translate(FileType::PDF, ...)`, and asserts the produced markdown is non-empty.
 
## Build prerequisites
- `uv` must be on PATH for `docling-sys/build.rs` to assemble the PyInstaller bundle (`brew install uv`).
- The first build takes a few minutes while the bundle is produced; subsequent builds are cached.
- At deploy time, ship the `run_docling/` directory alongside the executable (see `docling-sys/src/lib.rs:28-36`).
- For builds that don't need docling (CI lean builds, unrelated crate checks), set `DOCLING_SYS_SKIP_BUNDLE=1` to skip bundling. Runtime calls will then return an error.

## Test plan
- [x] `DOCLING_SYS_SKIP_BUNDLE=1 cargo check -p speedwagon --tests` — type-check passes.
- [x] `cargo test -p speedwagon` — non-PDF tests pass (uv-less environment OK with `DOCLING_SYS_SKIP_BUNDLE=1`).
- [x] `cargo test -p speedwagon translator::tests::translate_pdf_from_financebench -- --ignored --nocapture` — real docling round-trip (requires `uv` + network).

Resolves #60


jhlee525 (approved):
👍

grf53 (approved):

PR #73, opened 2026-05-13
feat(router,session): multi-intent step Plan with sequential dispatch

nuri-yoo:
## Summary

Replace the single-agent router with a multi-step plan, so a request that mixes intents can be dispatched to several sub-agents in order.

- `router.rs`: `RouterDecision` → `Plan { steps: Vec<Step> }`. One LLM call, returns an ordered list of `(agent, input, reason)`. Adds a `## Splitting` section to the instruction; `## Agents`, `## Rules`, and the second line of `## Response format` are kept verbatim. The parser validates `agent` against the `speedwagon | vegapunk | minerva` enum on every step and rejects empty `steps` arrays, with the existing 2-retry loop covering both. Final fallback returns a single-minerva Plan over the original input.
- `session.rs`: dispatch each step in order. Prior step outputs are prepended to the next step's prompt so dependent intents (e.g. "do X, then explain it") can see the earlier result. `Session` enum, slash commands, the streaming minerva UX, and the artifact dir cleanup all remain unchanged.

speedwagon and vegapunk dispatch are still stubs; multi-step plans that include them log the TODO message and append a placeholder so dependent steps still have context.

## Why

Evaluated on 35 multi-intent cases (24 multi + 11 single) and a 50-case single-intent set:

- **Multi-intent (24 cases):** the current router matches 0/24 by construction (single agent). The multi-step plan matches **20/24 (83.3%)**, coverage 91.7%. Elapsed 8.8 s → 16.8 s due to N sub-agent calls.
- **Single-intent:** the multi-step plan matches 9/11 in the small set and **44/50 (88.0%)** in the dedicated set. Over-split is 3/50 (6%), only on listy / "정리해서" / "with example" prompts.
- **JSON-escape probe (100 router-only calls × 5 meta/identity prompts):** ~2% set `agent: "fallback"` instead of an enum value. The parser's enum check + retry loop catches these. Structured Outputs lands in a follow-up to bring this to 0%.

## Test plan

- [x] `cargo build -p agent-k --bin session` succeeds
- [x] Single-intent QA → 1-step `speedwagon` plan
- [x] Multi-intent ("계산 + mutex 설명") → 2-step `minerva` → `speedwagon`, step 2 prompt receives step-1 output
- [x] artifact-wins preserved → "explain + tiny code" stays 1 minerva step
- [x] Meta/identity prompt × 10 runs → valid enum agent every time

## Out of scope (follow-up PRs)

- LLM-based stitcher that integrates multi-step outputs into one answer instead of streaming sequentially
- Structured Outputs on the router call to remove the ~2% enum escape deterministically

### DIFF
--- a/agent-k/src/agents/router.rs
+++ b/agent-k/src/agents/router.rs
@@ -22,7 +22,7 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
 - One step per distinct intent, in the order the user wrote them. A single intent — even if listy, long, or about multiple items — stays one step.
 - Each step.input is the corresponding part of the user's request, kept close to the original wording. The dispatcher passes step.input to the agent, with prior step outputs prepended as context, so phrase step.input as if those prior outputs are already in view.
 - An explanation paired with a tiny inline example is one minerva step (artifact wins) — the general split rule above does not apply when the example is inline.
-- Honor negations and self-corrections: only emit steps for what the user actually wants done.
+- Honor negations and self-corrections: only emit steps for the user's latest stated intent.
 
 ## Response format
 {"steps": [{"agent": "<agent name>", "input": "<sub-request>", "reason": "<one short sentence>"}]}
--- a/agent-k/src/agents/router.rs
+++ b/agent-k/src/agents/router.rs
@@ -46,7 +46,7 @@ pub struct Plan {
 const ROUTER_MAX_RETRIES: usize = 2;
 
 pub async fn run_gpt_router_agent(user_input: impl Into<String>) -> anyhow::Result<Plan> {
-    run_router_agent("openai/gpt-4o-mini", user_input).await
+    run_router_agent("openai/gpt-5.4-nano", user_input).await
 }
 
 pub async fn run_claude_router_agent(user_input: impl Into<String>) -> anyhow::Result<Plan> {
--- a/agent-k/src/agents/router.rs
+++ b/agent-k/src/agents/router.rs
@@ -8,22 +8,20 @@ use tokio_stream::StreamExt;
 const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request and produce a plan of one or more steps, each routed to a single agent.
 
 ## Agents
-- "speedwagon": RAG Q&A. Factual or knowledge questions answerable from a static document corpus.
+- "speedwagon": Q&A. Factual or knowledge questions, regardless of how the answer is sourced.
 - "vegapunk": Deep research. Multi-source investigation: literature review, topic survey, option comparison, or a long-form research report.
-- "minerva": General-purpose execution. Running commands, exploring or editing files/code, orchestrating multi-step work, fetching live information, producing code or runnable artifacts.
+- "minerva": General-purpose execution. Running commands, exploring or editing files/code, orchestrating multi-step work, producing code or runnable artifacts.
 
 ## Rules
-- Live information (today's weather, current stock price, today's news, anything that needs to be fetched right now) must route to minerva. speedwagon only covers static corpus knowledge. "As of <past date>" is a static fact, not live — route those to speedwagon.
-- If the request asks for both analysis/comparison and a concrete artifact (code, config, script, runnable example), the artifact intent wins — route to minerva.
-- If the request is primarily a question but also asks for an example, snippet, or code, treat the artifact intent as decisive and route to minerva.
+- A request that mixes Q&A/research with execution should split — speedwagon (or vegapunk) for the Q&A or research part, minerva for the execution part.
 - Requests to translate text from one specific language to another (e.g. "translate this Korean to English") are execution — route to minerva. Just writing in a non-English language is not a translation request.
 - If the request does not fit any agent well (greetings, identity questions about yourself, pure noise, ambiguous fragments, refusals to act), still pick the closest agent but prefix the "reason" field with "fallback: ".
 - Write "reason" in the dominant language of the user's request — the language carrying the semantic content, not short carrier phrases like "please" or "tell me".
 
 ## Splitting
 - One step per distinct intent, in the order the user wrote them. A single intent — even if listy, long, or about multiple items — stays one step.
-- Each step.input is a self-contained rewriting of that slice. The dispatcher passes step.input to the agent, with prior step outputs prepended as context, so phrase step.input as if those prior outputs are already in view.
-- An explanation paired with a code example is one minerva step (artifact wins).
+- Each step.input is the corresponding part of the user's request, kept close to the original wording. The dispatcher passes step.input to the agent, with prior step outputs prepended as context, so phrase step.input as if those prior outputs are already in view.
+- An explanation paired with a tiny inline example is one minerva step (artifact wins) — the general split rule above does not apply when the example is inline.
 - Honor negations and self-corrections: only emit steps for what the user actually wants done.
 
 ## Response format
--- a/agent-k/src/agents/router.rs
+++ b/agent-k/src/agents/router.rs
@@ -5,7 +5,7 @@ use ailoy::{
 use serde::Deserialize;
 use tokio_stream::StreamExt;
 
-const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request, split it into one or more sub-requests, and choose which agent should handle each.
+const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request and produce a plan of one or more steps, each routed to a single agent.
 
 ## Agents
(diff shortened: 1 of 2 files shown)


agent-k/src/agents/router.rs
```diff
@@ -20,40 +20,54 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
...
+- Each step.input must be a SELF-CONTAINED rewriting of that slice of the user's request — the dispatcher will pass step.input to the chosen agent. If a later step depends on the previous step's result, write step.input so that it can be answered given "the previous result" (which the dispatcher will provide as context).
+- Do NOT split a single research/survey/comparison task into per-item steps just because it lists multiple items.
+- A request that asks for an explanation WITH a code example is ONE minerva step (artifact wins).
+- Negations or self-corrections must be honored: only emit steps for what the user actually wants done.
+
 ## Response format
-{"agent": "<agent name>", "reason": "<one short sentence>"}
+{"steps": [{"agent": "<agent name>", "input": "<sub-request>", "reason": "<one short sentence>"}]}
 Respond with EXACTLY one JSON object, and nothing else (no prose, no markdown, no code fence). The "agent" field must be exactly one of the available agents.
 "#;
```
jhlee525: 1. There are quite a lot of uppercase words. Are all of them really necessary for emphasis?

2. If there is only a single step, what about simply redirecting the user’s original question as-is? (to preserve the user’s original phrasing and intent)

3. I think the response format could simply be: `
[{"agent": "<agent name>", "input": "<sub-request>", "reason": "<one short sentence>"}]
`
(If you are intentionally leaving room for future extension with fields beyond `steps`, it's ok.)

4. Also, if the intended flow is something like `query 1 -> based on that, query 2`, then I think the plan for how the result of query 1 gets passed into query 2 should be made explicit as well.

nuri-yoo: 1. Most are decorative. Will drop them later.
2.  Good point. When `plan.steps.len() == 1`, ignore `step.input` and forward the raw `user_input` is current plan. 
3. The wrapper was for future  use. For example,  requires_clarification  (the router asks the user to before dispatch which is Claude-style), explicit dependencies between steps. An object wrapper has future room for those without a schema break.
4. "if the intended flow is something like query 1 -> based on that, query 2" => It was intended. 
The dispatcher always prepends prior step  outputs before passing step.input to the sub agent (dependent or not — keeps dispatch simple). I'll make this explicit.

agent-k/src/agents/router.rs
```diff
@@ -16,44 +16,56 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
 - Live information (today's weather, current stock price, today's news, anything that needs to be fetched right now) must route to minerva. speedwagon only covers static corpus knowledge. "As of <past date>" is a static fact, not live — route those to speedwagon.
 - If the request asks for both analysis/comparison and a concrete artifact (code, config, script, runnable example), the artifact intent wins — route to minerva.
 - If the request is primarily a question but also asks for an example, snippet, or code, treat the artifact intent as decisive and route to minerva.
-- Requests to translate text from one specific language to another (e.g. "translate this Korean to English") are execution — route to minerva. Just writing in a non-English language is NOT a translation request.
+- Requests to translate text from one specific language to another (e.g. "translate this Korean to English") are execution — route to minerva. Just writing in a non-English language is not a translation request.
```
jhlee525, line 17: We may also need to consider introducing a translator subagent.

agent-k/src/agents/router.rs
```diff
@@ -16,44 +16,56 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
...
 - If the request does not fit any agent well (greetings, identity questions about yourself, pure noise, ambiguous fragments, refusals to act), still pick the closest agent but prefix the "reason" field with "fallback: ".
```
jhlee525, line 18: Not related to this PR, this might not be needed. (There is no use-cases when it get's fallback for `reason`).

Our interests only related to which agent is being routed.

agent-k/src/agents/router.rs
```diff
@@ -16,44 +16,56 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
 - Live information (today's weather, current stock price, today's news, anything that needs to be fetched right now) must route to minerva. speedwagon only covers static corpus knowledge. "As of <past date>" is a static fact, not live — route those to speedwagon.
```
jhlee525: As we discussed, currently there is no way to figuring out which information can be retrieved from speedwagon or minerva.

1 Using speedwagon for all Q&A tasks or 2 Upstreaming which can be retrieved via speedwagon, is planned todo right?

agent-k/src/agents/router.rs
```diff
@@ -16,44 +16,56 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
...
 - If the request asks for both analysis/comparison and a concrete artifact (code, config, script, runnable example), the artifact intent wins — route to minerva.
 - If the request is primarily a question but also asks for an example, snippet, or code, treat the artifact intent as decisive and route to minerva.
```
jhlee525: ditto

agent-k/src/agents/router.rs
```diff
@@ -16,44 +16,56 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
...
+- Requests to translate text from one specific language to another (e.g. "translate this Korean to English") are execution — route to minerva. Just writing in a non-English language is not a translation request.
 - If the request does not fit any agent well (greetings, identity questions about yourself, pure noise, ambiguous fragments, refusals to act), still pick the closest agent but prefix the "reason" field with "fallback: ".
 - Write "reason" in the dominant language of the user's request — the language carrying the semantic content, not short carrier phrases like "please" or "tell me".
 
+## Splitting
+- One step per distinct intent, in the order the user wrote them. A single intent — even if listy, long, or about multiple items — stays one step.
+- Each step.input is a self-contained rewriting of that slice. The dispatcher passes step.input to the agent, with prior step outputs prepended as context, so phrase step.input as if those prior outputs are already in view.
+- An explanation paired with a code example is one minerva step (artifact wins).
+- Honor negations and self-corrections: only emit steps for what the user actually wants done.
```
jhlee525: It encourages "query correction". If our policy is preserving user's original question, it is good to modify this line.

agent-k/src/agents/router.rs
```diff
@@ -16,44 +16,56 @@ const ROUTER_INSTRUCTION: &str = r#"You are a router. Read the user's request an
 - Live information (today's weather, current stock price, today's news, anything that needs to be fetched right now) must route to minerva. speedwagon only covers static corpus knowledge. "As of <past date>" is a static fact, not live — route those to speedwagon.
 - If the request asks for both analysis/comparison and a concrete artifact (code, config, script, runnable example), the artifact intent wins — route to minerva.
```
jhlee525: In this case, we are willing to select speedwagon + minerva right?

agent-k/src/agents/router.rs
nuri-yoo: Default sub agent (for general use) and translator sub agent could be considered.

agent-k/src/agents/router.rs
nuri-yoo: "Just writing in a non-English language is not a translation request." 
this message is a guardrail, and  necessary at least somewhere to prevent language confusion. 
Language confusion issue was found while testing cases. 

For " Requests to translate text from one specific language to another (e.g. "translate this Korean to English") are execution — route to minerva.", this should be removed along with implementation of somewhat translator sub agent.

agent-k/src/agents/router.rs
nuri-yoo: Opened to either approach and is todo.  As previous test was based on current instruction, implementation of 2 changes cause difficulty isolating the effect.

agent-k/src/agents/router.rs
nuri-yoo: Intentionally kept, as above.

agent-k/src/agents/router.rs
nuri-yoo: Will fix shortly.

agent-k/src/agents/router.rs
nuri-yoo: When an input mixes live info with corpus-style facts, the plan should split (minerva for the live     
  slice, speedwagon for the rest).  Will fix shortly.

jhlee525 (approved):
👍

grf53 (approved):

PR #74, opened 2026-05-13
feat(session): integrate multi-step outputs with a stitcher LLM

nuri-yoo:
## Summary

Stacks on top of #73. For multi-step plans only, integrate the per-step sub-agent outputs into one final answer using a small stitcher LLM (`gpt-5.4-mini`) before printing. Single-step plans keep the existing streaming UX and skip the stitcher entirely.

- For multi-step plans, sub-agent output is buffered (no stdout streaming during dispatch), then handed to the stitcher with the user's original request. The stitcher returns one concise answer in the user's language, which is printed at the end.
- Single-step plans stream as before — no extra LLM call, no behavior change for the common case.
- Stitch failure falls back to the prior `[step N • agent]` concat so the user still sees something useful.
- The helper is a private function inside `session.rs`. When backend-v2 adopts the same flow it can move into a library module.

## Why

The raw `[step N • agent] ...` concat is noisy when several sub-agents handle different slices of a request — the user has to mentally stitch the pieces back together. The stitcher turns N raw outputs into one answer with no step markers or meta-language.

Measured against the multi-step subset of the 35-case multi-intent set:
- Multi-step coverage: 91.7% → 90.0% (within noise; routing-time metrics unchanged because the stitcher runs after dispatch).
- Elapsed: 14.4 s → 14.9 s on average (~+3%). Stitch adds ~1.4 s on multi-step cases and 0 s on single-step (call is skipped).
- On the 50-case single-intent set, only 2/50 cases routed multi-step at all, so the stitcher fires rarely.

Qualitatively, multi-step answers no longer leak step headers and read as one coherent reply. Spot-checks (3 examples):
- "HTTP 418 + README line count" → one paragraph covering both, no `[step 1 ...]` headers.
- 4-intent plan (definition + comparison table + save + tree) → 745 chars stitched vs 2098 chars raw concat, with the comparison table preserved.
- "Compare LLM eval harnesses, separately compare Rust GPU libs" → 1566 chars stitched vs 7089 chars raw concat, both sections preserved with a closing one-line takeaway.

## Test plan

- [x] `cargo build -p agent-k --bin session` succeeds
- [x] Single-intent ("Python 한 줄 합산 코드") → 1-step `minerva`, streamed verbatim, no `[stitch]` log
- [x] Multi-intent ("1234 * 567 계산 + mutex 설명") → 2-step `minerva` → `speedwagon`, sub-agent outputs not streamed, `[stitch] integrating 2 step outputs` log, then one integrated answer
- [x] Stitch failure path reachable via the inline fallback (raw `[step N • agent]` concat)

## Out of scope (follow-up PRs)

- Wiring stitch into `backend-v2` so the FE can render either the integrated answer or per-step events.
- Structured Outputs on the router call to remove the ~2% enum escape deterministically.

### DIFF
--- a/agent-k/src/bin/session.rs
+++ b/agent-k/src/bin/session.rs
@@ -2,8 +2,11 @@
 //!
 //! Reads a user request from argv (or stdin if none given), asks a small router
 //! agent for a Plan (a sequence of (agent, input) steps), then dispatches each
-//! step in order. Step outputs are appended verbatim to stdout (streaming for
-//! minerva). Dependent steps see prior step outputs in their prompt.
+//! step in order. Single-step plans stream the sub-agent output directly to
+//! stdout (existing UX). Multi-step plans buffer each sub-agent's output and
+//! then call a small stitcher LLM (gpt-5.4-mini) to integrate the outputs into
+//! one final answer that is printed at the end. Dependent steps see prior step
+//! outputs in their prompt either way.
 //!
 //! echo "주간 보고 메일 초안 써줘" | cargo run -p agent-k --bin session
 //! cargo run -p agent-k --bin session -- "세계 날씨를 확인할 수 있는 간단한 HTML 페이지 만들어주세요"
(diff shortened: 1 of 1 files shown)


jhlee525 (approved):
👍

grf53 (approved):

PR #83, opened 2026-05-18
feat(files): batch move/copy via PATCH /dirents, unified rename

ljhh-0611:
## Summary

- Add **move** and **copy** for project dirents (files and folders) with sidebar drag-drop and dropdown "이동…" / "복사…" entry points.
- **Unify rename:** drop the dedicated `PATCH /dirents/{*path}` handler. Rename now flows through a single collection-level `PATCH /dirents` endpoint with a tagged op (`move` | `copy`); the rename UX is expressed as a 1-source move that carries `new_name`.
- One response shape — `DirentBatchResult` — is shared by upload, move, and copy, so clients consume a single contract everywhere.

## API changes

`POST /projects/{id}/dirents` (multipart upload) now returns `DirentBatchResult { project_id, succeeded: Dirent[], failed: FailedFile[] }`.

`PATCH /projects/{id}/dirents` is new. Body is a tagged union:

```jsonc
{ "op": "move", "sources": ["a.txt"], "destination": "Drafts",         "new_name": null }
{ "op": "move", "sources": ["a.txt"], "destination": "",               "new_name": "b.txt" }  // rename
{ "op": "copy", "sources": ["note.txt", "spec.md"], "destination": "Drafts" }
```

Response is `DirentBatchResult`. The endpoint follows a **partial-success contract**: per-source failures accumulate in `failed[]` without aborting the batch. Batch-wide errors (empty `sources`, `destination` not a directory, `new_name` with multiple sources) return 4xx and short-circuit.

Removed: `PATCH /projects/{id}/dirents/{*path}` (`rename_path`), `RenameRequest`, `UploadResponse`, `UploadedFile`.

## Behaviour highlights

- **Folder copy is recursive** (symlinks rejected throughout).
- **Finder-style auto-suffix on copy:** `note.txt` → `note copy.txt` → `note copy 2.txt`. Hidden files (`.keep`) get no extension split.
- **Self-into-descendant rejection:** moving or copying a folder into its own subtree fails per-source with a clear message.
- **Path safety:** every source and destination passes through `safe_join` (rejects `..`, absolute paths, NUL); `prepare_dest_dir` re-confirms the resolved path is a directory.

## Frontend UX

- Sidebar tree folders + the project-root row are drop targets with a highlighted outline. `Alt`/`Ctrl`/`Cmd` during a drop switches `move` → `copy`.
- Compact custom drag image (a small pill showing the filename or item count) replaces the browser default so the drop area stays visible.
- Body drop is a **silent cancel** — sidebar folders are the only valid intra-app destinations, so an aborted drag does not produce a spurious "already exists" toast or API call.
- A custom MIME (`application/x-cowork-dirent-paths`) lets the existing external-file upload drop zone coexist with intra-app drag.
- New `FolderPickerDialog` (tree picker reusing `buildFolderTree`) drives menu/toolbar-based "이동…" / "복사…", with source folders + their descendants disabled to prevent self-into-self.
- The existing `RenameDialog` is unchanged at the UX level; it now drives the unified `moveMutation` via a `renamingRef` flag so the success toast remains "이름이 변경되었습니다" rather than "이동되었습니다".

## Implementation notes

- Backend handler decomposes into focused helpers — `prepare_dest_dir` (batch-wide validation, returns `ApiError`), `load_source` (per-source validation, returns `String`), `move_one` / `copy_one`, `copy_dir_recursive`, `find_available_name`, `build_dirent`. The split between `ApiError` and `String` encodes the partial-success contract at the type level.
- Upload handler now also populates `modified_at` on each succeeded `Dirent` (single extra `metadata()` call per file).
- Drag-end cleanup is wired globally on `window` so any aborted drag clears the drop-target highlight; `dragLeave` uses `relatedTarget` containment check to avoid child-element flicker.

## Test plan

- [x] `cd backend && cargo build` — clean
- [x] `cd backend && cargo test --test dirent_test` — **12 / 12 pass**, including 4 new batch_op regressions:
  - `batch_op_rename_via_move_with_new_name`
  - `batch_op_copy_applies_finder_style_suffix`
  - `batch_op_move_folder_into_descendant_fails`
  - `batch_op_partial_success_on_name_conflict`
- [x] `cd app && npx tsc --noEmit` — clean
- [x] Manual UI smoke (Playwright + curl against the cowork demo):
  - Upload returns the new `DirentBatchResult` shape with `kind` / `bytes` / `modified_at`.
  - F2 and dropdown Rename produce `PATCH /dirents` with `{op:"move", ..., new_name}`; toast says "이름이 변경되었습니다".
  - "이동…" opens `FolderPickerDialog`; moving into `Drafts` updates the sidebar count immediately.
  - Drag a card onto the sidebar Drafts row → moves; drag back onto the file body → silently cancelled (no toast, no network call).
  - Same-folder copy creates `… copy.txt`, then `… copy 2.txt`.
  - Folder copy recursively brings descendants along.
  - Dragging a folder onto its own descendant is rejected.
  - External file upload (drop onto file body) still works.

### DIFF
--- a/backend/src/handlers/dirent.rs
+++ b/backend/src/handlers/dirent.rs
@@ -536,8 +536,16 @@ async fn build_dirent(uploads_root: &Path, host: &Path) -> Result<Dirent, String
         .strip_prefix(uploads_root)
         .map(|p| p.to_string_lossy().replace('\\', "/"))
         .unwrap_or_default();
-    let kind = if meta.is_dir() { DirentKind::Dir } else { DirentKind::File };
-    let bytes = if meta.is_dir() { None } else { Some(meta.len()) };
+    let kind = if meta.is_dir() {
+        DirentKind::Dir
+    } else {
+        DirentKind::File
+    };
+    let bytes = if meta.is_dir() {
+        None
+    } else {
+        Some(meta.len())
+    };
     let modified_at = meta.modified().ok().map(DateTime::<Utc>::from);
     Ok(Dirent {
         path: rel,
@@ -577,10 +585,7 @@ async fn find_available_name(parent: &Path, base_name: &str) -> PathBuf {
         }
     }
     // Fallback (effectively unreachable)
-    parent.join(format!(
-        "{stem} copy {}{ext}",
-        Uuid::new_v4().simple()
-    ))
+    parent.join(format!("{stem} copy {}{ext}", Uuid::new_v4().simple()))
 }
 
 /// Recursive folder copy. Symlinks are skipped to prevent escape.
@@ -650,11 +655,7 @@ async fn move_one(
 }
 
 /// Single copy (folder is recursive; auto " copy" suffix on conflict).
-async fn copy_one(
-    uploads_root: &Path,
-    src_rel: &str,
-    dest_dir: &Path,
-) -> Result<Dirent, String> {
+async fn copy_one(uploads_root: &Path, src_rel: &str, dest_dir: &Path) -> Result<Dirent, String> {
     let (src_host, src_meta) = load_source(uploads_root, src_rel, dest_dir).await?;
 
     let base_name = src_host
@@ -728,7 +729,10 @@ pub async fn batch_op(
             for src in sources {
                 match move_one(&uploads_root, &src, &dest_dir, new_name.as_deref()).await {
                     Ok(d) => succeeded.push(d),
-                    Err(e) => failed.push(FailedFile { path: src, error: e }),
+                    Err(e) => failed.push(FailedFile {
+                        path: src,
+                        error: e,
+                    }),
                 }
             }
             tracing::info!(
@@ -749,7 +753,10 @@ pub async fn batch_op(
             for src in sources {
                 match copy_one(&uploads_root, &src, &dest_dir).await {
                     Ok(d) => succeeded.push(d),
-                    Err(e) => failed.push(FailedFile { path: src, error: e }),
+                    Err(e) => failed.push(FailedFile {
+                        path: src,
+                        error: e,
+                    }),
                 }
             }
             tracing::info!(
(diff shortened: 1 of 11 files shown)


backend/src/handlers/dirent.rs
```diff
@@ -439,3 +446,307 @@ pub async fn delete_path(
...
+        if !cand.exists() {
+            return cand;
+        }
+    }
+    // Fallback (effectively unreachable)
+    parent.join(format!(
+        "{stem} copy {}{ext}",
+        Uuid::new_v4().simple()
+    ))
+}
```
khj809: This can call `Path::exists()` synchronously in maximum 1002 rounds, which may acquire a Tokio worker thread for a long time.
Suggest to use `tokio::fs::try_exists` and make this function async.

backend/src/handlers/dirent.rs
```diff
@@ -439,3 +446,307 @@ pub async fn delete_path(
...
+        } => {
+            if sources.is_empty() {
+                return Err(AppError::bad_request("sources must not be empty"));
+            }
+            if new_name.is_some() && sources.len() != 1 {
+                return Err(AppError::bad_request(
+                    "new_name is only valid for a single source",
+                ));
+            }
+            let dest_dir = prepare_dest_dir(&uploads_root, &destination).await?;
```
khj809, line 728: The batch ops always create the destination dir even it fails on every sources.
Would be better to guarantee that the dest dir is created after the first source's success, or delete the dest dir on the batch op failed for every sources.

backend/src/handlers/dirent.rs
ljhh-0611: `find_available_name` is now async and uses `tokio::fs::try_exists`.

backend/src/handlers/dirent.rs
ljhh-0611: `prepare_dest_dir` now only validates, not real creation.
`move_one`/`copy_one` each call `ensure_dest_dir` helper right before rename/copy.
Added a regression test that asserts the dest dir is absent on disk when every source fails.

khj809 (approved):
LGTM!

PR #85, opened 2026-05-18
feat: add message sender considering both user and agent

khj809:
### Backend: message sender tracking
- Added migration `0005_session_message_sender` — adds `sender_kind`, `sender_name`, `sender_user_id` columns to `session_messages`
- `classify_senders()` assigns each message in the agent history to either a user or the top-level agent (`agent-k`)
- `GET /sessions/:id/messages` response changed from a flat `AiloyMessage[]` to `SessionMessageListResponse`, where each item carries the message body alongside its sender
- Both `send_message` and `send_message_stream` persist sender info via `NewSessionMessage`

### Frontend: message sender model
- `Message.senderId: UserId` replaced with `Message.sender: MessageSender` — a tagged union `{ kind: 'user', userId }` | `{ kind: 'agent', name }`
- `ToolCallInvocation` type added (`id`, `name`, `arguments`, `result`)
- `AiloyToolCall`, `BackendMessageSender`, `SessionMessageItem`, `SessionMessageList` types added to `backend-types.ts`
- `toMessage()` replaced by `toMessageItem()` (handles sender info) + `collapseToolMessages()` (folds `role: tool` messages into the parent assistant message's `toolCalls`, attaching the result text)
- `listMessages()` now consumes the new `SessionMessageList` response shape

### Frontend: real-time tool call display
- `StreamUpdate.toolCalls` changed from `string[]` to `StreamToolCall[]` (`id`, `name`, `arguments`, `result`)
- `streamMessage()` now yields on both `role: assistant` and `role: tool` SSE events — the tool result is attached to the matching tool call as soon as it arrives
- `SessionPage` updates the live message's `toolCalls` on each yield, so `🔧 tool · running…` appears the moment a tool is invoked and the result fills in before the final answer

### Frontend: other
- `formatMessageTime` / `formatMessageTimeFull` utilities added; message timestamps show a full datetime tooltip on hover
- Sidebar session row layout fixed: title gets `flex: 1; min-width: 0` so it truncates properly, menu button wrapper gets `flex-shrink: 0; margin-left: auto` so it's never hidden by a long title

## Test plan

- [x] Messages in history show the correct sender (user vs agent)
- [x] During an agent response that uses tools, `🔧 tool · running…` appears immediately when the tool is invoked, the result appears when it finishes, then the final answer follows
- [x] Tool call results in message history are correctly attached to their parent assistant message
- [x] Long session titles in the sidebar truncate with ellipsis and the `…` menu button remains visible


backend/src/repository/session.rs
```diff
@@ -35,6 +35,46 @@ impl ShareMode {
...
+    pub created_at: DateTime<Utc>,
+}
+
+#[derive(Debug, Clone)]
+pub struct NewSessionMessage {
+    pub message: Message,
+    pub sender_kind: DbSenderKind,
+    pub sender_name: Option<String>,
+    pub sender_user_id: Option<Uuid>,
+}
```
ljhh-0611, line 76: ~I saw that sender_name and sender_user_id are mutually exclusive data, so how can we just merge this two data columns into one with String type..? It is just an opinion, not the strong one at all :)~
~Because id is Uuid and name is just String, so I know that we have to have a function for Uuid serde.~

~Or dont we have to have sender_kind in NewSessionMessge? If is_some(sender_name) equals to Agent and !is_some(sender_name) equals to User..?~

ljhh-0611: Just for now, I agree that simple is the best <^^
What I said will cause the function confusion with unneccesary long code.

backend/src/handlers/session.rs
```diff
@@ -400,27 +425,45 @@ pub async fn get_message_history(
...
+        .map(|r| SessionMessageResponse {
+            message: r.message,
+            sender: match r.sender_kind {
+                DbSenderKind::User => MessageSender::User {
+                    user_id: r.sender_user_id.unwrap_or(session.creator_id),
+                },
+                DbSenderKind::Agent => MessageSender::Agent {
+                    name: r
+                        .sender_name
+                        .unwrap_or_else(|| TOP_LEVEL_AGENT_NAME.to_string()),
```
ljhh-0611: I think the unwrap here should return an error instead!
It looks like error cases are being masked/hidden.

app/src/api/messages.ts
```diff
@@ -45,14 +52,22 @@ export async function* streamMessage(
...
+          toolCalls.push({ id: call.id, name: call.function.name, arguments: call.function.arguments });
+        }
+      }
+      yield { text: accumulated, toolCalls: [...toolCalls], status: 'streaming' };
+    } else if (msg.role === 'tool' && msg.id) {
+      const tc = toolCalls.find((t) => t.id === msg.id);
+      if (tc) {
+        tc.result = aiMessageText(msg.contents as AiloyPart[] | undefined) || '[done]';
+        yield { text: accumulated, toolCalls: [...toolCalls], status: 'streaming' };
       }
```
ljhh-0611:
```suggestion
      const msg = output.message;


      if (msg.role === 'assistant') {
        accumulated = aiMessageText(msg.contents as AiloyPart[] | undefined);
        for (const call of (msg.tool_calls ?? []) as AiloyToolCall[]) {
          if (!call.id || !call.function?.name) continue;
          const existing = toolCalls.find((tc) => tc.id === call.id);
          if (existing) {
            // Stub-backfill: a tool result arrived first and created an entry
            // with name === '(pending)'. Fill in the real name/arguments now.
            if (existing.name === '(pending)') existing.name = call.function.name;
            if (existing.arguments === undefined) existing.arguments = call.function.arguments;
          } else {
            toolCalls.push({
              id: call.id,
              name: call.function.name,
              arguments: call.function.arguments,
            });
          }
        }
        yield { text: accumulated, toolCalls: [...toolCalls], status: 'streaming' };
      } else if (msg.role === 'tool') {
        if (!msg.id) {
          console.warn('[streamMessage] tool message without id; cannot attach', msg);
          continue;
        }
        const resultText = aiMessageText(msg.contents as AiloyPart[] | undefined) || '[done]';
        let tc = toolCalls.find((t) => t.id === msg.id);
        if (!tc) {
          console.warn(
            `[streamMessage] tool result id=${msg.id} arrived without matching tool_call; rendering as stub`,
          );
          tc = { id: msg.id, name: '(pending)' };
          toolCalls.push(tc);
        }
        tc.result = resultText;
        yield { text: accumulated, toolCalls: [...toolCalls], status: 'streaming' };
      }
  ```

How about add warning and backfill in case of undefined?

File: app/src/lib/formatMessageTime.ts [hunk 4]
```diff
@@ -0,0 +1,18 @@
+import dayjs from 'dayjs';
+import 'dayjs/locale/ko';
+dayjs.locale('ko');
```
ljhh-0611:
```suggestion
  import dayjs from 'dayjs';
  ```
I suggest dropping 2~3 lines. If we ever add a locale-dependent token here, prefer instance-level `dayjs(iso).locale('ko')` over a module-level global set so the choice stays local.

Later on, we might consider adding a global language setting separately.

File: app/src/styles/chat.css [hunk 5]
```diff
@@ -55,31 +55,37 @@
...
+  list-style: none;
+  user-select: none;
 }
-.cw-toolcall .dot {
-  width: 5px;
-  height: 5px;
-  border-radius: 999px;
-  background: var(--cw-accent);
-  animation: cw-ping 1.1s steps(2, start) infinite;
+.cw-toolcall > summary::-webkit-details-marker { display: none; }
```
ljhh-0611, line 74: Tiny one — Firefox renders the disclosure triangle via standard `::marker`, not `::-webkit-details-marker`, so the ▶ still peeks through on FF next to the 🔧 emoji. One extra line covers it:

  ```suggestion
  .cw-toolcall > summary::-webkit-details-marker { display: none; }
  .cw-toolcall > summary::marker { display: none; }
  ```

ps) This comment is made by AI totally :)

ljhh-0611 (approved):

PR #87, opened 2026-05-20
Introduce Automation

grf53:
## Ticket
resolves #75 

## Implemation details
https://brekkylab.getoutline.com/doc/v1-I7bOIiyMVA

## What this PR contains

**Migration `0004_automations.sql`** introduces four tables:
- `automations`
- `automation_triggers`
- `automation_runs`
- `automation_run_events`

plus a `sessions.origin` column distinguishing user vs automation sessions.

The migration is edited in place across phases as the design firms up (`idempotency_key` column + UNIQUE partial index; removal of the now-derived `automation_run_events.attempt` column). Fresh DBs only; the file has not yet shipped.

**Three background task groups** spawned at boot:
- `spawn_workers(N)` — claim-poll loop, default `AGENT_K_WORKER_COUNT=2`
- `spawn_housekeeper` — `tokio::select!` of expired-lease reaper (60s) + idempotency-key cleanup (1h)
- `spawn_cron_ticker` — due-cron scan (15s)

**Routes**

| Method | Path | Notes |
|---|---|---|
| POST/GET | `/automations` | top-level, `project_id` in body/query |
| GET/PATCH/DELETE | `/automations/{id}` | |
| POST/GET/PATCH/DELETE | `/automations/{aid}/triggers[/{tid}]` | webhook trigger returns plaintext token once |
| GET | `/automations/{aid}/runs[/{rid}[/events]]` | |
| POST | `/automations/{aid}/runs` | manual queue |
| POST | `/webhooks/automations` | JWT-exempt; Bearer token + optional `Idempotency-Key` |
| POST/GET | `/sessions` | flattened in Phase 3 (`project_id` in body / query) |

**New env vars**:
- `AGENT_K_DEFAULT_TZ` — cron timezone fallback (defaults to UTC)
- `AGENT_K_WORKER_COUNT` — worker pool size

**New dependencies**:
- `tokio-util` — CancellationToken / DropGuard
- `croner 2.x` — 5-field POSIX cron
- `chrono-tz 0.10` — IANA timezone DB
- `sha2` — webhook token hashing

## Phased implementation
<details>
  <summary>Phase Details</summary>

### Phase 1 - DB Schema
3f96fee **feat: add automation DB schema (definitions, triggers, runs, events)**

Four new tables plus a session origin tag:
- `automations`: prompt sequence + project scoping
- `automation_triggers`: cron/webhook with hashed webhook token
- `automation_runs`: 1 run = 1 session, lease-based worker pickup
- `automation_run_events`: append-only audit log
- `sessions.origin`: distinguishes user vs automation sessions

### Phase 2 - Basic Repository methods
7ad9b09 **feat: add automation model and repository layer**

Typed Rust layer over the automation tables:
- `model/`: AutomationResponse, TriggerKind/TriggerSpec, RunStatus, EventKind + payload helpers, request DTOs
- `repository/automation.rs`: 22 methods (CRUD, race-safe `claim_due_run`, lease renew, expired-lease reaper, append-only event log)
- sessions: `SessionOrigin` enum + origin column, `create_session_with_origin`
- 8 unit tests including concurrent claim race and lease reap

### Phase 3 - CRUD HTTP API
da91ebc **feat: add automation CRUD HTTP API and flatten session routes**

automation endpoints:
- POST/GET `/automations`, GET/PATCH/DELETE `/automations/{id}`
- POST/GET `/automations/{aid}/triggers`, GET/PATCH/DELETE `…/{tid}`
- GET `/automations/{aid}/runs`, `/runs/{rid}`, `/runs/{rid}/events`
- webhook trigger returns plaintext token once; DB keeps only SHA-256 hash

session routes flattened: POST `/sessions` (project_id in body), GET `/sessions?project_id=…`

`sha2` added as a direct dependency. 9 new integration tests.

### Phase 4 - Basic Run (only manual)
a320912 **feat: add manual run endpoint and in-process worker**

- POST `/automations/{aid}/runs` queues a run atomically — session + run + triggered/queued events in a single transaction (`create_automation_run_with_session`)
- New worker module spawns N tokio tasks at boot: `claim_due_run` → sandbox+agent → iterate `prompts_json` as user messages → finalize. Lease/reaper deferred to Phase 5
- `handlers/session.rs` builders (build_sandbox/build_agent/sandbox_name_for) become `pub(crate)` so the worker reuses the same agent construction

### Phase 5 - Crash handling
62a9dd8 **feat: add automation worker lease + heartbeat + reaper**

- `spawn_heartbeat`: per-run 30s `renew_lease`. On `Ok(false)` (row reaped out from under us) it self-cancels and signals the agent. `CancellationToken` drop_guard catches panics during worker execution and fires cancel automatically on stack unwinding
- Outer `tokio::select!` races `execute_run` against the heartbeat JoinHandle — if HB ends first for any reason (panic, lease-loss self-cancel, normal cancel-ack), the agent future is dropped immediately and finalize is skipped; the reaper owns the row
- `spawn_reaper`: boot-time `reap_all_running` clears orphaned running rows from prior process incarnations; 60s tick reaps rows whose leases have expired. Each reap is a per-row transaction that atomically requeues the row and emits `lease_lost`
- `repository::finalize_run` packages terminal event INSERT + status UPDATE + lease NULL into one transaction, guarded by `status='running'` so it cannot clobber a row already requeued by the reaper

`LEASE_MINUTES=3`, `HEARTBEAT_INTERVAL=30s`, `REAP_INTERVAL=60s`. Adds `tokio-util` for `CancellationToken` + `DropGuard`.

### Phase 6 - Cron trigger
733077c **feat: add cron trigger to automation worker**

- New `backend::cron` module with `next_fire_after(expr, tz, now)` and a `default_tz_name()` helper that reads `AGENT_K_DEFAULT_TZ` at first call (OnceLock-cached) and falls back to UTC for missing or invalid values
- Only 5-field POSIX cron accepted; 6-field / sub-minute rejected at parse. Uses `croner` for expression parsing — chosen over the `cron` crate for stricter Linux-cron semantics that align with the one-minute-minimum policy
- `handlers::automation` computes `next_fire_at` for Cron triggers on create and on update; invalid expressions or timezones return 400
- `repository::fire_cron_trigger` atomically inserts the session, run, and initial triggered+queued events and advances the trigger's `next_fire_at` in a single transaction, so a half-applied firing can neither leave the trigger pointed at a past time nor silently drop the firing
- `worker::spawn_cron_ticker` polls every 15s and calls `cron_tick_once(state, Utc::now())` — passing `now` as an explicit argument keeps the tick function trivially testable without introducing a Clock trait

Adds `croner 2.x` and `chrono-tz 0.10`. 5 new cron unit tests + 1 repo test.

### Phase 7 - Webhook trigger
17d68c3 **feat: add webhook trigger with idempotency-key dedupe**

POST `/webhooks/automations` on a JWT-exempt route. Bearer token SHA-256 hash both identifies and authenticates against the UNIQUE partial index on `automation_triggers.webhook_token_hash` — a single DB lookup replaces application-level constant-time comparison.

Optional `Idempotency-Key` header dedupes per-trigger retries via a new `automation_runs.idempotency_key` column + UNIQUE partial index `(trigger_id, idempotency_key) WHERE NOT NULL`. Pre-check lookup falls back to re-lookup on UNIQUE violation, so concurrent retries with the same key converge on the same run. The Phase 5 reaper is extended into `spawn_housekeeper` with an hourly cleanup that NULLs keys past 24h, releasing the partial-index slot for caller-side reuse.

### Phase 8 - Retry
ac82064 **feat: add retry + backoff for failed automation runs**

Failed runs are automatically chained back into the queue: `schedule_retry` atomically inserts a fresh session + new run (`previous_run_id`, `scheduled_for=now+backoff`) + a `queued` event + a `retry_scheduled` event on the previous run. The picker query is unchanged — the future `scheduled_for` gates the retry. `MAX_ATTEMPTS=3` with backoffs `[30s, 2m]`.

`attempt` is a derived property, not a persisted column: the worker calls `compute_run_attempt` once per claim (recursive CTE over `previous_run_id`, bounded by MAX) and caches it locally. Reaper-race safety: retry only triggers on `finalize_run` `Ok(true)` — `Ok(false)` means reaper already requeued the row, so re-claim resumes the same attempt instead of inflating the chain.

Schema cleanup in an in-place 0004 edit: `automation_run_events.attempt` is dropped (redundant — all events of a run share the same attempt by construction), and five unused typed payload helper structs are removed (dead since Phase 1/2). One new unit test (`schedule_retry_chains_new_run_and_emits_retry_event`).
</details>

## Test plan

- [x] `cargo test -p agent-k-backend` — full suite (28 lib tests + integration files) all green
- [x] Each phase doc (`docs/automation-phaseN-*.html`, untracked here) includes a "개발자 수동 테스트 시나리오" section with reproducible recipes

## Known limitations (deferred)

- **Non-atomic finalize + schedule_retry** — ms-scale window; can be tightened by merging into one repo method (`finalize_run_with_retry`). v1 monitors via the "stuck chain" query in Phase 8 docs.
- **At-least-once side effects** — prompt re-execution possible on reaper-race or worker crash; users responsible for prompt idempotency.
- **No error classification** — all errors retry with the same schedule; permanent failures waste up to 2.5 min.
- **No per-automation retry policy** — `MAX_ATTEMPTS` / `RETRY_BACKOFFS` are code constants.
- **No worker-level e2e tests** — `schedule_retry` is unit-tested; full "fail → retry → succeed" path is manual-only until a fake repo or LLM stub is introduced.


grf53:
[Currently working on frontend app with this.]

jhlee525:
Since it contains a large changes, it would be better to merge first, and revise continuously.

jhlee525 (approved):
👍  Thank you for huge works!

backend/src/model/mod.rs
```diff
@@ -1,9 +1,17 @@
+mod automation;
+mod automation_run;
+mod automation_run_event;
+mod automation_trigger;
```
khj809: It seems not necessary to split automation-related models into several mods. Why don't we just use a consolidated `automation` mod?

backend/src/main.rs
```diff
@@ -118,6 +118,15 @@ async fn run_server() -> std::io::Result<()> {
...
     let app_state = Arc::new(AppState::new(repo, store, jwt, data_root));
+
+    let worker_count = std::env::var("AGENT_K_WORKER_COUNT")
+        .ok()
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(2usize);
+    worker::spawn_workers(app_state.clone(), worker_count);
+    worker::spawn_housekeeper(app_state.clone());
+    worker::spawn_cron_ticker(app_state.clone());
+
```
khj809: The current main entrypoint always run both API server and worker threads. If someone just want to test workers only, there's no way to do that in current main. At least we should provide CLI options for that, or separate the bin entirely for running worker-only process.

grf53: I added `--mode` flag for `agent-k-backend serve`


```
# cargo run --bin agent-k-backend -- serve --help
...
Run the HTTP server and/or worker loops (default when no subcommand is given)

Usage: agent-k-backend serve [OPTIONS]

Options:
      --mode <MODE>
          Which components to run in this process

          Possible values:
          - api:    HTTP API only — no worker / housekeeper / cron ticker
          - worker: Worker + housekeeper + cron ticker only — no HTTP listener
          - all:    Both API and worker loops in the same process (default)
          
          [default: all]

  -h, --help
          Print help (see a summary with '-h')
```

khj809 (approved):

PR #88, opened 2026-05-20
Coworker agent

jhlee525:
- Add `coworker` agent (previously `minerva`)
- Make new binary `run` and `test_case` instead of `session`.
- Add some test cases


nuri-yoo (approved):
Tested 8 cases in 3 vendor. LGTM.

agent-k/src/agents/coworker.rs
```diff
@@ -0,0 +1,91 @@
...
+    config.cpus = 8;
+    config.memory_mib = 1024;
+    config.workdir = "/workspace".into();
+    config.env.insert("HOME".into(), "/workspace".into());
+    config.volumes.push(VolumeMount::Bind {
+        host: artifacts_dir.as_ref().into(),
+        guest: "/workspace/artifacts".into(),
+        readonly: false,
+    });
+    config.persist = true;
```
khj809: The reason why the sandbox disappears immediately is that the sandbox `Machine`'s `shutdown()` removes the msb instance if `persist` is false. However, the sandbox machine can be booted/shutdowned multiple times across agent's lifetime whenever the tool call happens.

Therefore, we need to apply the new policy about how to persist the sandbox:
- First of all, we should fix `shutdown()` not to remove the msb instance ever, since it can be reused in next tool executions.
- Either remove sandbox on `RunEnv` is dropped and `persist` is false, or add an explicit `remove` API to `RunEnv` and delegate the responsibility to users.

khj809: Confirmed it works without `persist: true` on https://github.com/brekkylab/ailoy/pull/415.
But we need to change the main function to exit normally on ctrl+c in order to let `Sandbox` to be `Drop`ped properly.

jhlee525: Can we merge it after https://github.com/brekkylab/ailoy/pull/415 is being merged?

khj809: Sure. I'll push the working version.

PR #112, opened 2026-05-22
[App] Refine layout shell and sidebar interactions

ljhh-0611:
**Recommend to just try sidebar handling yourselves and give feedback**

## Summary

- Add persisted app layout state so the shell can drive sidebar mode and width.
- Refine the left sidebar into isolated project/session sections with desktop reveal, mobile drawer behavior, and resizer affordances.
- Tune hidden/reveal thresholds: hide below 120px, return to expanded near 185px, keep reveal usable within about 230px, and prevent the sidebar from disappearing while the resizer is held.

## Issues

Relates to #103, #104, #105

## Verification

- `pnpm -C app lint`
- `pnpm -C app test:unit -- src/stores/layout.test.ts`
- `git diff --check -- app/src/components/layout/Sidebar.tsx app/src/routes/_app.tsx app/src/styles/cowork-design-system.css app/src/styles/globals.css app/src/stores/layout.ts app/src/stores/layout.test.ts`
- Browser smoke on `http://127.0.0.1:4110/projects` with no console errors observed

### DIFF
--- a/app/src/components/Icon.tsx
+++ b/app/src/components/Icon.tsx
@@ -6,13 +6,13 @@ export type IconName =
   | 'send' | 'sparkles' | 'check' | 'chevron' | 'zap' | 'calendar' | 'artifact'
   | 'shield' | 'x' | 'arrow-left' | 'analysis' | 'brainstorm' | 'writing'
   | 'recap' | 'general' | 'more' | 'upload' | 'trash' | 'download'
-  | 'list' | 'grid' | 'chevron-right'
+  | 'list' | 'grid' | 'chevron-right' | 'chevron-left'
   | 'file-pdf' | 'file-code' | 'file-archive' | 'file-video' | 'file-audio';
 
 type PreviewIconName =
   | 'lock' | 'eye' | 'message-square' | 'messages-square' | 'clipboard-list'
   | 'lightbulb' | 'pencil' | 'home' | 'folder' | 'folder-open' | 'users'
-  | 'settings' | 'plus' | 'chevron-down' | 'chevron-right' | 'more-horizontal' | 'check'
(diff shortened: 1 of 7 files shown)


ljhh-0611:
@grf53 Thanks for the detailed comparison — pushed 29070ac to address all four points. Recap below, matched 1:1 to your table.

### 1) Expand-with-drag no longer auto-pins

`getSidebarModeForDrag` is now one-way (shrink-only). Dragging in the hidden state freezes the mode and only adjusts the floating panel width — pinning is reachable solely via the explicit toggle. Matches Outline.

### 2) Explicit Hidden / Pinned toggle button

Added an in-sidebar header toggle that is mode-aware:
- `expanded` → chevron-left, collapses to hidden
- `hidden + revealed` (floating) → chevron-right, pins back to expanded

The chevron pair keeps the button visually consistent across states. Desktop hamburger is gone — Outline-style entry on desktop, mobile hamburger preserved as a drawer toggle.

### 3) Pinned → Hidden now produces a clear visual change

The button-toggled hide path keeps `revealed = true` while the cursor is still inside the sidebar, and the panel morphs into a **floating card** (8px top/bottom inset, 6px left, 12px radius, larger soft shadow) — Notion-style cue that this is a temporary state. When the cursor leaves the panel horizontally, it slides out. Drag-to-hide funnels into the same floating state, so both entry paths look identical.

### 4) Hidden → Pinned now animates content shift

The sidebar is now `position: fixed` in both modes, and `--cw-sidebar-w` is registered via `@property` so the grid column width interpolates. The result is a single 220ms transition graph:

- expanded ↔ hidden eases all spatial properties (top/height/border-radius/box-shadow/border-right-width)
- hidden + revealed ↔ hidden + not revealed is **transform-only**, so reveal open/close has zero vertical motion — only horizontal slide

grf53 (approved):
I appreciate you accepting my feedback a lot.

PR #114, opened 2026-05-22
feat: revise dirents API to cover artifacts and attached files

khj809:
## Summary

This PR migrates the agent sandbox from a single upload directory to a three-directory layout and wires up the corresponding UI: a per-session artifacts panel, per-message file attachments with thumbnail previews, a copy-to-shared flow, and artifact chips that appear below agent messages with bidirectional deletion sync.

### Dependencies

- **ailoy updated to microsandbox 0.5.0** (`rev = 38d32c4`) — the workspace `ailoy` dependency is pinned to a new revision that targets microsandbox 0.5.0.

### Backend

**3-directory bind mount migration**
- `agent-k`: Added `CoworkerSandboxOptions` and `get_coworker_agent_with_opts` to allow callers to specify sandbox name and `persist` flag. Existing CLI callers are unchanged via a thin wrapper.
- Added `pub const` path values in `coworker.rs` (`GUEST_ATTACHED_DIR`, `GUEST_SHARED_DIR`, `GUEST_ARTIFACTS_DIR`) so the guest mount paths are defined in a single place and shared across the agent config, backend session handler, and tests.
- `backend/session.rs`: Replaced the old `build_sandbox` + `build_agent` pair with `build_session_agent`, which prepares three host directories per session and mounts them into the sandbox:
  - `projects/{pid}/shared/` → `/workspace/shared` (read-only, project-scoped)
  - `projects/{pid}/sessions/{sid}/inputs/` → `/workspace/attached` (read-only, accumulates per session)
  - `projects/{pid}/sessions/{sid}/artifacts/` → `/workspace/artifacts` (read-write)
- `cleanup_session_resources` now removes the session host directory on deletion.

**Global `/dirents` API**
- Removed per-project dirent routes (`/projects/{pid}/dirents/…`). Replaced with two global routes: `GET|POST|PATCH /dirents` (collection ops via `?path=` query) and `GET|DELETE /dirents/{*path}` (single-file ops via wildcard).
- `parse_dirent_path`: single entry point for scope determination. Validates UUID segments, rejects `..` and unknown prefixes (400), cross-project access (404).
- `enforce_scope_access`: unified authorization across all three scopes.
  - `Inputs` writes require `SessionAccess::Admin` (session owner only — prevents ChatMembers from overwriting session inputs)
  - `Artifacts` writes require at least `ChatMember`
- `dirent_batch_op`: `copy` across scopes is allowed within the same project (enables artifacts → shared copy); `move` remains same-scope only.

**Message attachments — store-clean / inject-at-LLM-time design**
- Migration `0007`: adds `attachments TEXT NOT NULL DEFAULT '[]'` and `artifacts TEXT NOT NULL DEFAULT '[]'` columns to `session_messages`.
- User message content is stored clean in `message_json`; the `[Attached files …]` hint is reconstructed from the `attachments` column at LLM-context time via `build_attachment_note` / `inject_attachment_note`. The note never appears in the raw message returned to the frontend.
- Path resolution per scope: `Inputs` → `GUEST_ATTACHED_DIR/<basename>` (flat mount); `Shared` → `GUEST_SHARED_DIR/<full-tail>` (directory-preserving mount).
- `validate_attachments`: scope must be `inputs` for the current session or `shared` for the same project; file must exist on disk and must not be a symlink (prevents sandbox escape via symlink traversal).

**Artifact-message linkage with bidirectional deletion sync**
- `send_message` and `send_message_stream`: snapshot the artifacts directory before `agent.run()`, diff with `HashSet::difference` after, and attach new relative paths to the last agent message's `artifacts` column before persisting.
- `remove_artifact_from_messages`: when `dirent_delete` removes an artifacts-scoped file, the path is removed from the `artifacts` JSON array on every matching message row (uses `seq` PK — avoids SQLx `ColumnNotFound` panic from SQLite's non-standard `rowid`).
- `SessionMessageResponse` exposes `artifacts: Vec<String>` so the frontend can read it directly.

**Lazy sandbox initialization**
- `create_session` is now a pure DB operation. The sandbox is created lazily on the first `send_message` via the existing `resolve_agent_for` path.

### Frontend

- **`api/dirents.ts`**: Full rewrite around a `DirentScope` discriminated union (`shared | inputs | artifacts`). `scopeRoot` and `stripScopePrefix` helpers keep path construction in one place. `copyDirents` accepts src/dst scopes for cross-scope copy.
- **Files page**: Migrated to `{ kind: 'shared', projectId }` scope. File icons in both list and grid views now use `FileTypeIcon`; folders retain the existing pocket icon.
- **`FileTypeIcon` component**: New pure-SVG file badge component — a folded-corner document shape filled with an extension-specific brand color and a short label. Covers 60+ formats.
- **ArtifactsPanel**: Lists agent-generated files for the session. Features a collapsible header with file count badge, per-row `···` menu (download, copy-to-shared, delete), and a bulk `···` menu that appears on the select-all row. Clicking anywhere on a row (except the `···` area) toggles the checkbox. After deletion, `['messages', sessionId]` query is also invalidated so artifact chips update immediately. Selection is automatically filtered when entries change (e.g. item deleted via chip menu), so the "N개 선택됨" count always reflects currently-visible items.
- **CopyToSharedDialog**: Folder picker that mirrors `FolderPickerDialog`; calls cross-scope `copyDirents` on confirm. Rendered via `createPortal` to escape the grid stacking context and cover the full viewport including the sidebar.
- **Artifact chips below agent messages**: Agent messages that created artifacts display file chips directly below the message body. Chips read `message.artifacts` from the API (no client-side snapshot/diff). Clicking a chip opens a dropdown menu (download / copy to shared / delete) — the `···` button is removed; the chip itself is the trigger. `overflow: visible` on `.cw-artifact-chip` prevents the dropdown from being clipped by the parent `overflow: hidden`.
- **Bidirectional chip sync**: deleting an artifact via any surface (chip menu, ArtifactsPanel, or API) removes it from both the filesystem and the message's `artifacts` DB column; the frontend re-fetches messages so the chip disappears without page reload.
- **Per-message attachments**: Files can be attached in the composer (uploaded to `inputs/` in the background). Message bubbles render image thumbnails or download chips for other file types.
- **`seed_cowork_demo.py`**: Includes `artifacts` column in the message INSERT; seeds the GTM report session with artifact paths on the agent message so chips are visible out of the box.

### Tests

- `session_messages_test`: 8 pure-function unit tests for `build_attachment_note` / `inject_attachment_note`; 3 security tests (`deny_unknown_fields`, symlink rejection, clean stored body)
- `dirent_test`: `chat_member_cannot_upload_to_session_inputs`, `non_member_gets_403_on_patch_endpoint`
- All `NewSessionMessage` struct literals updated with `artifacts: vec![]`

## Test plan

- [x] `cargo test -p agent-k-backend` — all tests pass
- [x] `pnpm tsc --noEmit` in `app/` — no TypeScript errors
- [x] Attachment persisted in message history with correct `attachments` field; note injected for agent, not exposed to frontend
- [x] Cross-scope copy `artifacts → shared` succeeds; file appears in shared listing
- [x] Delete session → `sessions/{sid}/` removed from disk; `shared/` and its files retained
- [x] Agent writes file to `/workspace/artifacts/` → ArtifactsPanel auto-refreshes; download works
- [x] Copy artifact to shared via row menu and bulk selection → file appears on Files page
- [x] FileTypeIcon renders correct brand colors in Files page list+grid views and ArtifactsPanel rows
- [x] Seed data: GTM report session shows artifact chips (GTM_summary_report.md, ICP_comparison_table.csv) below agent message
- [x] Clicking artifact chip opens dropdown menu (download / copy to shared / delete)
- [x] Deleting artifact via chip → file removed from disk, chip disappears, ArtifactsPanel updates
- [x] Deleting artifact via ArtifactsPanel → chip in message also disappears (bidirectional sync)
- [x] ArtifactsPanel row click (outside `···`) toggles checkbox; "N개 선택됨" count stays in sync when items are deleted
- [x] CopyToSharedDialog overlay covers the full viewport including sidebar


ljhh-0611 (approved):
Thx..

grf53 (approved):

PR #115, opened 2026-05-23
feat: short slug + 12-char prefix URLs for projects and sessions

ljhh-0611:
## What this PR does

Project and session URLs become human-readable, and they stay stable when a project is renamed.

- A project lives at `/projects/{slug}` instead of `/projects/{uuid}`.
- A session lives at `/projects/{slug}/sessions/{prefix}` where `prefix` is a hyphen-free 12-char hex head of the session UUID (`aaaaaaaa2222`).
- Renaming a project keeps every old link alive — the old slug redirects to the new one.

The backend stays the source of truth on the full UUID, but it no longer cares whether a request names a project by UUID or by slug: **one resolver accepts both (plus retired slugs), and every project/session route flows through it.**

### Creating a project → a slug is minted

`POST /projects` with just a name. The server slugifies the name, guarantees uniqueness, and stores the slug as a permanent column on `projects`.

- `"KlientCo Q2 Analysis"` → `klientco-q2-analysis`
- `"한글 프로젝트"` → `hangeul-peurojegteu` (romanised)
- same name again → `klientco-q2-analysis-2` (numeric suffix on collision)
- a name that slugifies to nothing (`"...!!!"`) → `project-<6-char nanoid>` fallback

A caller may also pass an explicit `slug`; it's validated server-side (1–64 chars, lowercase/digit/hyphen, alnum ends, no `--`) so a malformed slug can never reach the database. The frontend then routes to `/projects/{slug}` and the project card / sidebar key off the slug.

### Opening any project URL → one resolver handles it

Every project-scoped route — `GET/PATCH/DELETE /projects/{ref}`, `/projects/{ref}/members…`, `/projects/{ref}/dirents…` — takes a single `{ref}` segment. The handler resolves it the same way session handlers already resolve their id:

```
resolve_project_id(ref):
    parse as UUID?        → use it
    else active slug?     → map to project id
    else retired slug?    → map to project id   (renamed-project links)
    else                  → 404
```

So `GET /projects/{uuid}` and `GET /projects/{slug}` return the exact same project, and the frontend can put a slug in the URL while the backend keeps operating on the UUID (on-disk paths, FK columns, response `id` fields all stay UUID).

This replaces the earlier draft's split surface (a dedicated `GET /projects/by-slug/{slug}` endpoint plus `Path<Uuid>` everywhere else) and lets the frontend drop its client-side `resolveProjectId` shim + process-lifetime cache entirely — the API client now passes the slug straight through to the URL.

### Renaming a project → old links keep working

`PATCH /projects/{ref}` with a new slug, in one transaction:

1. the current slug is written to `project_slug_history` (`ON CONFLICT DO NOTHING`, so a slug retired by another project is never reassigned),
2. the project row flips to the new slug.

Now both URLs resolve to the same project. When the frontend loads a project and notices the resolved `project.slug` differs from the slug in the URL (i.e. an old link was followed), it `redirect({ replace: true })`s to the canonical slug URL, preserving whatever sub-path the user was on. A retired slug also can't be claimed by a *different* project — that returns `409`.

### Session URLs → a short prefix that round-trips

The frontend shortens a session UUID to its first 12 hex chars (`shortSessionId`). The backend reinserts the canonical `8-4-4-4-12` hyphens and does a `LIKE 'prefix%' LIMIT 2` lookup against the PK index:

- exact/unique match → resolve,
- two or more matches → `409 "ambiguous session prefix; provide more characters"`,
- nothing → `404`.

Full UUIDs and the legacy hyphenated 12-char form keep resolving too, so links already in flight don't break. `POST /sessions` (body `project_id`) and `GET /sessions?project_id=…` (query) accept a slug for the project as well, via the same resolver.

## Why a single endpoint (design note)

The session work in this PR already taught the backend to accept "UUID or short form" via `resolve_session_id`. Keeping projects on a separate `by-slug` endpoint would have left the two halves of the same PR asymmetric — and the asymmetry was actively hiding bugs (see below). Collapsing project routes onto the same pattern made the contract narrower, deleted a frontend shim, and brought the suite back to green.

## DB note

New migration `0007_project_slug.sql` (slug column + unique index + `project_slug_history` table). Numbered `0007` because `0006_automations.sql` (PR #87) landed on main first. Per the CLAUDE.md dev policy, wipe `backend/data/*.db` and restart to pick it up.

New crates: `slug`, `nanoid`, `sha2`.

## Verification (run locally on this branch)

Automated:
- `cargo test -p agent-k-backend` → **28 passed, 1 ignored (LLM-key test), 0 failed**
- `pnpm tsc --noEmit` → 0 errors · `pnpm lint` → clean · `pnpm build` → bundle OK
- fresh DB boot applies migration `0007` cleanly and the server comes up

Manual E2E against a throwaway DB (curl), all confirmed:

| Flow | Result |
|---|---|
| English name → auto slug | `klientco-q2-analysis` |
| Korean name → romanised | `hangeul-peurojegteu` |
| Duplicate name → suffix | `klientco-q2-analysis-2` |
| Unslugifiable name → fallback | `project-B25ftb` |
| Explicit valid slug | accepted |
| Explicit invalid slug (uppercase) | `400` with reason |
| `GET /projects/{uuid}` vs `/projects/{slug}` | identical project |
| `PATCH` slug, then GET old slug | resolves, `slug` field shows the new one (redirect signal) |
| `POST` reusing a retired slug | `409 slug already in use` |
| Unknown slug | `404` |
| `POST /sessions` with slug in body | created |
| `GET /sessions?project_id={slug}` | lists |
| `GET /sessions/{hex-prefix}` / full UUID / legacy hyphenated | all resolve to the same session |
| Ambiguous 1-char prefix | `409 ambiguous session prefix; provide more characters` |
| Unknown prefix | `404` |

## Test plan (reviewer)

- [x] `cargo test -p agent-k-backend`
- [x] `pnpm tsc --noEmit` / `pnpm lint`
- [x] fresh DB → migration `0007` applies cleanly
- [x] auto-slug (en/ko/collision/empty-fallback) + explicit-slug validation
- [x] rename → old slug redirects to current; retired slug reuse → 409
- [x] `GET /projects/{uuid}` and `/projects/{slug}` return the same project
- [x] session prefix / full UUID / legacy hyphenated all resolve; ambiguous → 409
- [x] `pnpm build` (production bundle)
- [x] in-browser click-through (Chrome): login -> /projects -> card -> files/members/session, slug + 12-char prefix URLs, retired-slug redirect with sub-path preserved, 0 console errors


grf53 (approved):
LGTM

backend/src/repository/user.rs
```diff
@@ -199,8 +199,9 @@ impl SqliteRepository {
         new_user: NewUser,
     ) -> RepositoryResult<(DbUser, crate::repository::DbProject)> {
         let user = self.create_user(new_user).await?;
+        let personal_slug = format!("personal-{}", nanoid::nanoid!(6));
```
grf53: There can be a conflict. 
You could move and use `is_slug_taken()` in `handlers/project.rs`, or its codes like:

```suggestion
        let mut personal_slug = format!("personal-{}", nanoid::nanoid!(6));
        while self.get_project_by_slug(&personal_slug).await?.is_some()
            || self
                .get_project_id_by_retired_slug(&personal_slug)
                .await?
                .is_some()
        {
            personal_slug = format!("personal-{}", nanoid::nanoid!(6));
        }
```

PR #122, opened 2026-05-26
Session fork UI

grf53:
## Ticket
resolves #106 

## Description
- **Session duplicate (fork)** — exposes `POST /sessions/{id}/fork` in three UI spots (**sidebar menu**, **session header**, **automation run detail**).
  - Inline buttons use two-click confirm; sidebar menu uses a confirm dialog.
  - Success navigates to the new session in all three.
- **`IconButton` primitive** — ghost icon-only button with `expandedText` (hover-expand) and `confirmText` (two-click confirm). `RailAction` is refactored to use it.
- Removes the unused `SessionIntent` feature (the transformer hard-coded every session to `general`, so the badges/icons carried no signal).

## Test plan
- [x] Sidebar menu → Duplicate → confirm dialog → lands on the new session
- [x] Session header Duplicate button: hover expands, two clicks fire
- [x] Automation run detail Duplicate (verify new session is `origin = user`)
- [x] Trigger row hover/focus reveal animation still works (RailAction)
- [x] No `IntentBadge`/`IntentIcon` rendered anywhere; session lists still load
### Added tests
- [x] No `GET /sessions//artifacts` 400 in network tab right after fork
- [x] Duplicate from sidebar menu / session header / automation run drawer all land on `/projects/{slug}/sessions/{prefix}` with a success toast
- [x] Automation run: button disabled for `queued`/`running`, enabled for terminal statuses
- [x] Session detail: button disabled while streaming and on a 0-message session, enabled otherwise
- [x] Pending: every Duplicate IconButton goes disabled + shows "복제 중..." until navigation
- [x] Sidebar menu: Duplicate item disabled on empty sessions; confirm dialog body mentions "in use may fail"; dialog stays open until success/error

### Sidebar menu
스크린샷 2026-05-28 오후 12 58 20
image
image

### Session page header
basic / hover / click once(twice for confirm)
(you can see the features of `IconButton` component here.)
스크린샷 2026-05-28 오후 12 58 42
스크린샷 2026-05-28 오후 12 58 48
스크린샷 2026-05-28 오후 12 58 53
image

### Automation run detail panel
basic / hover / click once(twice for confirm)
스크린샷 2026-05-28 오후 12 59 02
스크린샷 2026-05-28 오후 1 04 05
스크린샷 2026-05-28 오후 12 59 28
image

### DIFF
--- a/app/src/components/SessionCardMenu.tsx
+++ b/app/src/components/SessionCardMenu.tsx
@@ -9,12 +9,13 @@ import { Icon } from './Icon';
 
 interface SessionCardMenuProps {
   onDuplicate?: () => void;
+  duplicateDisabled?: boolean;
   onDelete?: () => void;
 }
 
 interface MenuRect { top: number; left: number; }
 
-export function SessionCardMenu({ onDuplicate, onDelete }: SessionCardMenuProps) {
+export function SessionCardMenu({ onDuplicate, duplicateDisabled, onDelete }: SessionCardMenuProps) {
   if (!onDuplicate && !onDelete) return null;
   const [open, setOpen] = useState(false);
   const [rect, setRect] = useState<MenuRect | null>(null);
@@ -106,7 +107,13 @@ export function SessionCardMenu({ onDuplicate, onDelete }: SessionCardMenuProps)
             <button
               type="button"
               role="menuitem"
-              onClick={(e) => { e.stopPropagation(); setOpen(false); onDuplicate(); }}
+              disabled={duplicateDisabled}
+              onClick={(e) => {
+                e.stopPropagation();
+                if (duplicateDisabled) return;
+                setOpen(false);
+                onDuplicate();
+              }}
               style={{
                 display: 'flex',
                 alignItems: 'center',
@@ -116,12 +123,16 @@ export function SessionCardMenu({ onDuplicate, onDelete }: SessionCardMenuProps)
                 padding: '7px 10px',
                 border: 0,
                 background: 'transparent',
-                color: 'var(--cw-ink-1)',
+                color: duplicateDisabled ? 'var(--cw-ink-4)' : 'var(--cw-ink-1)',
                 fontSize: 12.5,
                 borderRadius: 'var(--cw-radius-sm)',
-                cursor: 'pointer',
+                cursor: duplicateDisabled ? 'not-allowed' : 'pointer',
+                opacity: duplicateDisabled ? 0.6 : 1,
               }}
-              onMouseEnter={(e) => { (e.currentTarget as HTMLButtonElement).style.background = 'var(--cw-paper-3)'; }}
+              onMouseEnter={(e) => {
+                if (duplicateDisabled) return;
+                (e.currentTarget as HTMLButtonElement).style.background = 'var(--cw-paper-3)';
+              }}
               onMouseLeave={(e) => { (e.currentTarget as HTMLButtonElement).style.background = 'transparent'; }}
             >
               <Icon name="sticky-notes" size={13} />
--- a/app/src/components/SessionCardMenu.tsx
+++ b/app/src/components/SessionCardMenu.tsx
@@ -16,7 +16,7 @@ interface SessionCardMenuProps {
 interface MenuRect { top: number; left: number; }
 
 export function SessionCardMenu({ onDuplicate, duplicateDisabled, onDelete }: SessionCardMenuProps) {
-  if (!onDuplicate && !onDelete) return null;
+  const hasActions = Boolean(onDuplicate || onDelete);
   const [open, setOpen] = useState(false);
   const [rect, setRect] = useState<MenuRect | null>(null);
   const buttonRef = useRef<HTMLButtonElement>(null);
@@ -60,6 +60,8 @@ export function SessionCardMenu({ onDuplicate, duplicateDisabled, onDelete }: Se
       document.removeEventListener('keydown', onKey);
     };
   }, [open]);
+
+  if (!hasActions) return null;
 
   return (
     <>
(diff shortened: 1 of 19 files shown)


app/src/routes/_app.projects.$projectId.index.tsx
khj809: Not included in this PR, but this points to the old link.

```suggestion
        <button onClick={() => navigate({ to: '/projects/$projectId/automation', params: { projectId } })}>
```

app/src/routes/_app.projects.$projectSlug.automation.index.tsx
```diff
@@ -345,26 +347,20 @@ function AutomationsPage() {
...
         </div>
         <div className="cw-run-drawer-actions">
+          <IconButton
+            icon="sticky-notes"
+            label="이 run의 세션 복제"
+            title="이 run의 세션을 복제"
+            expandedText="세션으로 복제"
+            confirmText="한 번 더 눌러 복제"
+            onClick={() => duplicateMutation.mutate(run.sessionId)}
+          />
```
khj809, line 377: This button is always displayed without any guards, so users may try to fork the runs still in progress. Like the below cancel button, I think this should be guarded by sth like `forkable`.

grf53: Added guards that block duplications when:
- queued/running automation runs
- empty(no message) sessions
- streaming sessions
  - forking a streaming session triggered on sidebar would be blocked by backend (no frontend guard).
  This is communicated to the user in the confirm dialog.
- disable duplicate buttons during in-flight duplication

app/src/components/SessionCardMenu.tsx
ljhh-0611: This early return runs before the component calls hooks, so the same `SessionCardMenu` instance can render with zero hooks when both actions are absent, then later render with `useState`/`useRef`/effects once `onDuplicate` or `onDelete` becomes available.

I was able to reproduce a React runtime error by toggling this state in a browser harness. Can we keep the action check, but move the conditional return after all hooks? That keeps the hook order stable while preserving the same rendered output.


```suggestion
  const hasActions = Boolean(onDuplicate || onDelete);
  const [open, setOpen] = useState(false);
  const [rect, setRect] = useState<MenuRect | null>(null);
  const buttonRef = useRef<HTMLButtonElement>(null);
  const popoverRef = useRef<HTMLDivElement>(null);

  // Compute popover position relative to the ⋯ trigger. Right-align so the
  // menu fits whether trigger sits at the edge of a 200px sidebar or in a
  // wider card grid cell. Recompute on resize/scroll while open.
  useLayoutEffect(() => {
    if (!open || !buttonRef.current) return;
    function place() {
      const r = buttonRef.current!.getBoundingClientRect();
      const POPOVER_WIDTH = 150;
      const left = Math.max(8, Math.min(window.innerWidth - POPOVER_WIDTH - 8, r.right - POPOVER_WIDTH));
      const top = r.bottom + 4;
      setRect({ top, left });
    }
    place();
    window.addEventListener('resize', place);
    window.addEventListener('scroll', place, true);
    return () => {
      window.removeEventListener('resize', place);
      window.removeEventListener('scroll', place, true);
    };
  }, [open]);

  // Close on outside click or Escape.
  useEffect(() => {
    if (!open) return;
    function onDocClick(e: MouseEvent) {
      const target = e.target as Node;
      if (buttonRef.current?.contains(target)) return;
      if (popoverRef.current?.contains(target)) return;
      setOpen(false);
    }
    function onKey(e: KeyboardEvent) { if (e.key === 'Escape') setOpen(false); }
    document.addEventListener('mousedown', onDocClick);
    document.addEventListener('keydown', onKey);
    return () => {
      document.removeEventListener('mousedown', onDocClick);
      document.removeEventListener('keydown', onKey);
    };
  }, [open]);
  
  if (!hasActions) return null;

  return (
```

ljhh-0611: 19 & 64 line are the changed code suggestion

app/src/routes/_app.projects.$projectSlug.automation.index.tsx
```diff
@@ -341,30 +344,37 @@ function AutomationsPage() {
   const renderRunDetail = (run: Run) => {
     const status = effectiveStatus(run);
     const cancellable = status === 'queued' || status === 'running';
+    const forkable = !cancellable;
```
ljhh-0611, line 347: `effectiveStatus()` with `cancelledIds` turns `forkable = !cancellable` into `true` even though the backend may still have the session locked.

grf53: Regarding this matter, it seems we ultimately need to introduce a `cancelling` status to the automation run. While the necessity for this was not a major issue previously, it became more apparent with the addition of the duplicate function. Due to the delay between the cancel trigger and the session lock release, there is potential for utilizing the `cancelling` status beyond just the duplicate function, making its addition well worthwhile. However, since this requires considering the automation system as a whole, I believe it is best to separate this into a distinct issue(https://github.com/brekkylab/agent-k/issues/131).

The issue here is that 'duplicate can fail even though the duplicate button is enabled,' which is not critical. Therefore, I intend to close this PR by providing a more detailed toast message. https://github.com/brekkylab/agent-k/pull/122/changes/eaa7e2f68c8a9d6c8684058d6bc7b1c57dd457b2

app/src/components/SessionCardMenu.tsx
grf53: applied!

ljhh-0611 (approved):
It's good after merge conflict

khj809 (approved):

PR #124, opened 2026-05-28
feat(app): redirect to /login on JWT expiry with session toast

khj809:
## Summary

- **Centralized 401 handling** — `api/client.ts` now exports `setUnauthorizedHandler` / `notifyUnauthorized`. All three HTTP exit points (`request()`, `streamSse()`, `fetchFileBlob()`) funnel 401 responses through this hook. A module-level `unauthorizedTriggered` flag prevents duplicate handler fires from concurrent in-flight requests; the flag resets on `setToken()` so the next session's expiry is still caught after re-login.

- **`forceLogout` helper** (`src/lib/forceLogout.ts`) — single source of truth for the logout flow: disconnects WebSocket, clears zustand auth state, stores `auth.expiredReason` and `auth.redirectAfterLogin` in `sessionStorage`, then navigates using `router.navigate` (injected from `main.tsx`). If `router.navigate` rejects, falls back to `window.location.href` to guarantee the redirect. The router-based navigation prevents a race where a full-page reload would consume the `sessionStorage` items before the login page could read them.

- **Proactive token validation** — `_app.tsx` `beforeLoad` is now async and calls `getMe()` on cold cache (first load or after stale eviction). Subsequent in-app navigations use the cached result without blocking. A 401 response redirects immediately to `/login`; other errors (network, 5xx) fall through so the page handles them in context.

- **WebSocket reconnect guard** — `api/ws.ts` stops the reconnect loop on close code `4401` (app-defined auth failure). Additionally, a rapid-close heuristic fires after ≥ 3 consecutive connections that each close within 2 s of being established — measured as connection lifetime, not the interval between closes, so the heuristic correctly accumulates across the 3-second reconnect delay. Close code `1008` (generic Policy Violation) is intentionally excluded — proxies and firewalls emit it for non-auth reasons and it must not trigger logout. After either trigger, `getMe()` is called over HTTP to confirm the token is actually invalid before forcing logout; if `getMe()` succeeds the disconnect is treated as a transient WS flap, not an auth failure. This also correctly classifies the failure as `expired` vs `invalid` based on the actual server response.

- **Session-expired toast on login page** — `login.tsx` reads `auth.expiredReason` from `sessionStorage` on mount and shows a 5-second auto-dismissing toast. The dismiss callback is memoized with `useCallback` so the auto-dismiss timer is not reset by re-renders caused by typing in the form. Messages are localized (`'expired'` → "세션이 만료되어 다시 로그인이 필요합니다." / `'invalid'` → "인증 정보가 유효하지 않습니다."). After successful login, navigates to `auth.redirectAfterLogin` (validated as a same-origin path) instead of the default `/projects`.

- **Unified logout** — sidebar's inline `reset() + window.location.href` is replaced with `forceLogout({ reason: 'manual' })`. Manual logout carries no `auth.expiredReason`, so no toast appears.

- **Query retry** — `QueryClient` default retry now skips retries on 401 to avoid wasted round-trips with a known-invalid token.

- **`JWT_EXPIRED_MESSAGE` constant** — the `'token has expired'` string matched against backend error responses is extracted to a named constant with a comment pointing to `backend/src/auth/jwt.rs`, making the cross-layer coupling explicit.

## Findings during testing

> **Backend JWT leeway is 60 s** — `jsonwebtoken` v9's `Validation` defaults `leeway` to 60 seconds. A `JWT_EXPIRY_SECONDS=10` token therefore stays accepted for up to 70 s. Manually crafted tokens with `exp = now − 90` were used to reliably trigger 401 in tests.

## Test Records

https://github.com/user-attachments/assets/4f70c8db-b9d7-466d-a20e-3051d54c896c

## Test plan

- [x] Expired token → navigate to `/projects` → redirected to `/login` with "세션이 만료…" toast
- [x] Invalid token → "인증 정보가 유효하지 않습니다." toast
- [x] Login form with wrong credentials → shows credential error only, no session toast (regression)
- [x] Post-login redirect to original URL (e.g. `/projects/foo`)
- [x] Manual logout via sidebar → `/login` with no toast
- [x] Re-login → second session expiry correctly triggers the handler again (`unauthorizedTriggered` reset)
- [x] WS disconnect with code 4401 → `getMe()` HTTP call determines expired vs invalid before logout
- [x] `pnpm lint` passes
- [x] 14 unit tests pass (`ws.test.ts` 7개, `client.test.ts` 5개, `layout.test.ts` 2개)

### DIFF
--- a/app/src/api/client.test.ts
+++ b/app/src/api/client.test.ts
@@ -1,0 +1,36 @@
+import { describe, it, expect, beforeEach } from 'vitest';
+import { setUnauthorizedHandler, notifyUnauthorized } from './client';
+
+describe('notifyUnauthorized reason classification', () => {
+  let receivedReason: string | undefined;
+
+  beforeEach(() => {
+    receivedReason = undefined;
+    setUnauthorizedHandler((reason) => { receivedReason = reason; });
+  });
+
+  it('classifies JWT_EXPIRED_MESSAGE as "expired"', () => {
+    notifyUnauthorized(401, { error: 'token has expired' });
+    expect(receivedReason).toBe('expired');
+  });
+
+  it('classifies any other message as "invalid"', () => {
+    notifyUnauthorized(401, { error: 'unauthorized' });
+    expect(receivedReason).toBe('invalid');
+  });
+
+  it('classifies missing error field as "invalid"', () => {
+    notifyUnauthorized(401, {});
+    expect(receivedReason).toBe('invalid');
+  });
+
+  it('does not fire for non-401 status', () => {
+    notifyUnauthorized(403, { error: 'token has expired' });
+    expect(receivedReason).toBeUndefined();
+  });
+
+  it('does not fire when skipAuth is true', () => {
+    notifyUnauthorized(401, { error: 'token has expired' }, true);
+    expect(receivedReason).toBeUndefined();
+  });
+});
--- a/app/src/api/client.ts
+++ b/app/src/api/client.ts
@@ -4,6 +4,26 @@
 
 const BASE_URL_KEY = 'cowork.v2.baseUrl';
 const TOKEN_KEY = 'cowork.v2.token';
+
+export type UnauthorizedReason = 'expired' | 'invalid';
+let unauthorizedHandler: ((reason: UnauthorizedReason) => void) | null = null;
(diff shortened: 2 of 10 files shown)


grf53 (approved):
LGTM

ljhh-0611 (approved):

PR #127, opened 2026-05-28
chore(dep): update ailoy using microsandbox 0.5.2, fix frontend bugs

khj809:
### Backend: bump ailoy + microsandbox (msb 0.5.2)

- Updated ailoy to rev `58e0c38` (`38d32c4` → `58e0c38`)
- microsandbox (msb) crates bumped from `0.5.0` (git tag) → `0.5.2` (crates.io)
  - Source changed from the `superradcompany/microsandbox` git repo to the official crates.io registry release

### Frontend: fix premature `useQuery` calls returning 404 / 400

Both frontend bugs are race conditions: `useQuery` calls fired before their upstream data dependencies had resolved, passing empty strings into URL-constructed paths.

- **`Sidebar.tsx`** — The sessions query was `enabled` as soon as `activeProjectSlug` was present in the URL, but `activeProject` (resolved from `projectsQuery.data`) could still be `undefined` at that point. This caused `listSessions('')` to fire with `GET /sessions?project_ref=` (empty string) → 404. Fixed by gating the query on both `activeProjectSlug` and `activeProject`, and removing the `?? ''` fallback from the `queryFn`.

- **`ArtifactsPanel.tsx`** — The artifacts dirents query had no `enabled` guard. The parent initializes `sessionId` as `session.data?.id ?? ''`, so while the session is still loading `sessionId` is `''`. This produced a malformed URL `GET /dirents?path=projects/.../sessions//artifacts` (double slash) → 400. Fixed by adding `enabled: Boolean(sessionId)`.

## Test plan

- [x] Verify the backend builds cleanly against the updated ailoy rev and msb 0.5.2
- [x] Navigate to a project page — confirm no 404 for `/sessions?project_ref=` in the network tab
- [x] Open a session — confirm no 400 for `/dirents?path=...sessions//artifacts` in the network tab
- [x] Confirm the sidebar session list loads correctly once the projects query resolves
- [x] Confirm the artifacts panel loads correctly once the session UUID is available

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -151,7 +151,7 @@ dependencies = [
 [[package]]
 name = "ailoy"
 version = "0.3.0"
-source = "git+https://github.com/brekkylab/ailoy.git?rev=38d32c4a3e2db3333e0287819770ed77bc7f5e66#38d32c4a3e2db3333e0287819770ed77bc7f5e66"
+source = "git+https://github.com/brekkylab/ailoy.git?rev=58e0c38af84b7a38257065d4f623a8060fa20f4c#58e0c38af84b7a38257065d4f623a8060fa20f4c"
 dependencies = [
  "anyhow",
  "async-stream",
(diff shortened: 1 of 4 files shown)


nuri-yoo (approved):
LGTM

PR #116, opened 2026-05-25
Automation UI

grf53:
## Ticket
resolves #99 

## Description

Wire the `/automation` UI to the real backend, replacing the prior mocked pages.

- **Sessions API** — add `origin: 'user' | 'automation'` so automation-spawned sessions can be hidden from sidebar and project surfaces; switch client off the nested `/projects/:id/sessions` shape onto flat `/sessions?project_id=`.
- **Automation API** — domain + backend types and a thin client covering CRUD, manual run, cancel, and one-shot webhook token reveal.
- **Pages** — listing (rail + filterable runs with retry-chain grouping, paginated; drawer polling messages + events), create (cron via SchedulePicker / webhook / none), and settings (inline trigger CRUD, immediate save; automation body via Save changes).
- **Shared primitives**
  - `ConfirmDialog` portals to body (so it stacks above other portaled dialogs). New `confirmOnEnter` prop: when set, Enter triggers `onConfirm`, a [`corner-down-left`](https://lucide.dev/icons/corner-down-left) glyph renders next to the confirm label, and an `Esc` hint appears next to cancel. Used for the manual-run confirm.
  - **Two-click-to-confirm pattern** for in-place destructive actions (no dialog needed):
    - Trigger row trash icon (`$automationId.tsx`): first click swaps the icon for a filled-red "한 번 더 눌러 삭제" pill via `.cw-trigger-action.is-armed`; second click fires the delete.
    - `WebhookTokenDialog`'s `onDiscard` button: caller drives the `discardArmed` state so the secondary button flips to a filled-destructive style on first click and only commits on the second.
  - `WebhookTokenDialog` also gains a curl-copy button next to the token reveal.
- **New Lucide icons** —  [`copy`](https://lucide.dev/icons/copy),  [`rotate-ccw`](https://lucide.dev/icons/rotate-ccw),  [`corner-down-left`](https://lucide.dev/icons/corner-down-left). 
- **Demo** — idempotent automation seeder hooked into `run_cowork_demo.sh`.

### Automation page
<details>
  <summary>Show</summary>
스크린샷 2026-05-25 오후 10 32 52
</details>

#### Run selected (wide)
<details>
  <summary>Show</summary>
  스크린샷 2026-05-25 오후 10 33 44
</details>

#### Run selected (narrow)
<details>
  <summary>Show</summary>
스크린샷 2026-05-25 오후 10 34 05
</details>

### Runs list (center panel)
<details>
  <summary>Show</summary>
[...]

PR #128, opened 2026-05-28
feat(i18n): English source + Korean via react-i18next

ljhh-0611:
**This feature keep goes on all of other app changes, so decide to adapt ASAP and follow up**

## Summary
- Externalize all user-facing copy with react-i18next (English source-of-truth, Korean translations) across 8 lazy-loaded namespaces.
- Add backend `users.preferred_language` column (migration 0009) with a `PATCH /me { preferred_language }` flow; frontend caches the choice in localStorage and treats the backend as the source-of-truth on hydrate.
- Reduce hardcoded Korean lines in `app/src` (excluding new automation pages that landed on main after this PR's phases) from 235 to **37** — every line in that count is intentional residue (josa fixtures, `JosaPair` literal type, code comments, language-toggle "한" glyph).

> **Coverage note**: This PR was opened against an earlier `main`. While in review, `main` absorbed PR #116 (Automation UI: `automation.*.tsx`, `SchedulePicker`, `WebhookTokenDialog`) which adds **~121 hardcoded Korean lines** of its own. Those files arrived via merge and are **out of scope here** — they need a follow-up Phase E pass. Total `app/src` Korean residue post-merge: 158 (37 retained intentional + 121 from #116).

## What's in the box
- **react-i18next infra**: namespace-per-feature (`common`, `auth`, `project`, `session`, `files`, `dialogs`, `errors`, `members`), browser-language-detector chain (`localStorage → navigator`), Vite chunk-split per locale × namespace, dayjs locale tied to `i18n.language`.
- **Route-loader namespace prefetch**: every route declares the namespaces its component tree consumes via a `loadNs(...)` helper. TanStack Router awaits these in parallel before mounting the route, so lazy ns no longer leak into the outer Suspense boundary and produce blank flashes on first navigation. The `_app` shell loader covers everything the always-mounted Sidebar can pull in (`common`, `project`, `session`, `dialogs`), and `login.tsx` primes the `['me']` query so `_app` hits cache instead of refetching.
- **`josa()` helper**: Korean particle agreement with proper `와/과` asymmetry handling. `localizedNoun(noun, pair, lang)` gates the rule to `ko` so English filenames never receive a stray Korean particle.
- **Language toggle**: segmented `EN | 한` control in the sidebar user area (single row with the logout button via a 4-col grid). Persists to localStorage and (when signed in) `PATCH /me`. `signupAndLogin` syncs the browser-detected language to the backend so Korean signups don't get flipped to English on hydrate.
- **Error mapping**: `app/src/api/error-messages.ts` maps `(ApiError, scope)` to i18n keys with a raw English-message fallback for unmapped statuses (backend stays English-only by design).
- **Test coverage**: 43 vitest tests dedicated to i18n — 14 `josa` edge cases (받침 / 와·과 asymmetry / non-Hangul fallback), 8 namespace parity assertions, 19 exhaustive enum-keyed lookups (`SessionIntent` × field × locale, `ShareMode` × field × locale). After merging main, total test count is 55 (12 new from main: `client.test.ts`, `ws.test.ts`).

## Backend migration (0009)
- `ALTER TABLE users ADD COLUMN preferred_language TEXT NOT NULL DEFAULT 'en';`
- `UpdateMeRequest.preferred_language` is validated against `'en' | 'ko'`; new tests `validate_language_*` cover accept / reject / case-sensitivity.
- Forward-only. Existing rows backfill with `'en'`. Dev DBs do not need to be reset.

## Commit structure
Thirteen commits — eight alternating `feat`/`chore` for the phase review loop, three post-review fixes, then a merge resolution + chore cleanup pair after `main` absorbed four other PRs:

- `3934d68` Phase A — infra
- `e55be0a` chore — Phase A review feedback
- `95c6ee9` Phase B — common components / dialogs / login
- `bb682de` chore — Phase B review feedback
- `c86fe8a` Phase C — route hotspots + ArtifactsPanel + Sidebar
- `5d6cdb4` chore — Phase C review feedback
- `3e7e5a1` Phase D — residual placeholder pages
- `3da6e86` chore — Phase D review feedback
- `6405c2d` chore — sidebar 4-col grid keeps LanguageToggle + logout on one row
- `1ce59f2` fix — prefetch i18n namespaces in route loaders (removes blank-flash after login)
- `fec0e83` fix — cover sidebar-mounted components in shell ns prefetch + externalize the lone hardcoded "Activity" label
- `5c32c2d` docs — correct init comment about Suspense fallback role
- `601d26c` Merge `main` — resolve 6 file conflicts (globals.css, transformers.ts, ConfirmDialog, Sidebar, settings.tsx rewrite, login.tsx)
- `0787ea4` chore — post-merge cleanup from architect + critic review (i18nify two stray aria-labels, drop two dead keys)

The chore commits intentionally land alongside the feat commits so the history records each round of review feedback; squashing on merge is also fine if a cleaner main history is preferred.

## Known follow-ups
- **Phase E — Automation UI i18n**: PR #116 added `automation.index.tsx` / `automation.$automationId.tsx` / `automation.new.tsx` / `SchedulePicker.tsx` / `WebhookTokenDialog.tsx` with ~121 hardcoded Korean lines. They came in via merge but were not part of any phase in this PR. Needs its own pass.
- **settings.tsx error mapping**: the rewritten `messageOf` inside `settings.tsx` still has a raw `${status} — ${message}` fallback for unmapped HTTP statuses. The cleaner pattern is `apiErrorToMessage` (used by `login.tsx`). Worth unifying in a small follow-up.
- **schedule.tsx orphan route**: PR #116 replaced the sidebar "Schedule" nav with "Automation" but left `_app.projects.$projectSlug.schedule.tsx` reachable by URL. The `common:sidebar.nav.schedule` key is now unreferenced from the sidebar; both should be cleaned up when the route file is removed.
- `compactTime()` was removed as dead code (zero call sites, hardcoded demo-era date matches). If relative-time formatting becomes useful again, build it on top of dayjs's locale-aware `fromNow()`.
- Optional: auto `navigator.language === 'ko' → PATCH /me` on first signup; a real `/settings` surface that hosts the language toggle.
- The root Suspense boundary in `main.tsx` is still `fallback={null}`. Route loaders make this safe today, but adding a dedicated boundary around `<Outlet />` inside `AppShell` would shrink the blast radius if a future component adds a ns its parent loader missed.

## Test plan
- [x] `pnpm lint` clean (post-merge)
- [x] `pnpm test:unit` 55/55 (post-merge; 12 added by main)
- [x] `pnpm build` no INEFFECTIVE_DYNAMIC_IMPORT warning (pre-merge)
- [x] `cargo test -p agent-k-backend --lib` (48 incl. new `validate_language_*`)
- [x] Manual: language toggle round-trip across refresh; signup as a ko user without getting flipped to en
- [x] Manual: ProjectsPage, files, sessions/$prefix, dialogs all render in both `ko` and `en` with no key-literal leaks
- [x] Manual: login → /projects transition no longer flashes blank


app/src/api/auth.ts
```diff
@@ -36,16 +37,44 @@ export async function signup(input: SignupInput): Promise<User> {
...
+// backend before returning so the auth store doesn't flip the UI on hydrate.
 export async function signupAndLogin(input: SignupInput): Promise<User> {
   await signup(input);
-  return login({ username: input.username, password: input.password });
+  const user = await login({ username: input.username, password: input.password });
+
+  const detected = i18n.language;
+  const isSupported = (SUPPORTED_LANGUAGES as readonly string[]).includes(detected);
+  if (isSupported && detected !== user.preferredLanguage) {
+    return updateMe({ preferredLanguage: detected as SupportedLanguage });
```
khj809: `updateMe` is called without try/catch, which may result in users get stucked in login page.

app/src/i18n/josa.ts
```diff
@@ -0,0 +1,35 @@
...
+  '은/는': { withBatchim: '은', without: '는' },
+  '으로/로': { withBatchim: '으로', without: '로' },
+  '와/과': { withBatchim: '과', without: '와' },
+};
+
+function hasBatchim(ch: string): boolean {
+  if (!ch) return false;
+  const code = ch.charCodeAt(0);
+  if (code < HANGUL_START || code > HANGUL_END) return false;
+  return (code - HANGUL_START) % 28 !== 0;
```
khj809: When the josa is `으로/로`, `ㄹ` should be considered as non-batchim, unless some illegal examples such as `파일으로`, `서울으로` are accepted.

app/src/api/auth.ts
ljhh-0611: Thanks! Wrapped the `updateMe` in a `try/catch` with a `console.warn`, so a failed language sync no longer throws back into `login.tsx::onSubmit`. The user proceeds with the backend default (`en`) and can toggle from the sidebar afterwards.

Also added an `onError` to the matching mutation in `lib/useLanguage.ts` for the toggle path, since the same silent-fail pattern lived there.

app/src/i18n/josa.ts
ljhh-0611: Added the `ㄹ`-jongseong exception inside `hasBatchim(pair)`: `jong === 8 && pair === '으로/로'` now returns `false`, so `서울 + 으로/로 → 서울로`.

Other particle pairs still treat `ㄹ` as a regular batchim — added a regression guard so a future "treat ㄹ as no-batchim everywhere" patch can't silently slip in: `서울을` / `서울이` / `서울은` / `서울과` all stay correct.

khj809 (approved):
Fixed ColumnNotFound("preferred_language") bug on getting project members, along with broken tests on auth, automation, project and sandbox.
LGTM!

grf53 (approved):
Thanks

PR #132, opened 2026-05-30
feat(ui): generic <Select> (custom panel + native dropdown, portaled & collision-aware); adopt app-wide

ljhh-0611:
## Summary
Extract ShareSelect's bespoke dropdown into a **reusable, app-wide `<Select>` component**, and refactor `ShareSelect` to consume it. The component keeps the *functional* core (adaptive trigger width, custom listbox, keyboard/a11y, collision-aware positioning) generic and leaves per-consumer styling open, so other "pick one value" surfaces can adopt it.

> Originally branched off `feat/i18n` (#128, now merged); rebased onto `main`. Uses `useTranslation('common')` + the `share.*` keys from there.

## Why
The native `<select>` locked the trigger to the widest option's width (a long empty tail for short labels), and its dropdown/option/disabled rendering is OS-drawn and unstyleable. ShareSelect solved this with a custom dropdown — this PR generalizes that solution so it isn't ShareSelect-only.

## `<Select>` API
```tsx
type SelectOption<T> = { value: T; label: string; icon?: IconName; className?: string; disabled?: boolean };
type SelectGroup<T>  = { label?: string; options: SelectOption<T>[] };

<Select
  value={value}
  onChange={setValue}
  options={flatOrGrouped}
  renderTrigger?   // escape hatch
  renderOption?    // escape hatch
  triggerClassName?  // per-consumer trigger look (no styling baked in)
  adaptiveWidth?     // default false
  nativeDropdown?    // default false → our panel
  boundary?          // CSS selector: confine the panel inside this ancestor
  ariaLabel? disabled? placeholder?
/>
```

## Behavior
- **Two modes, default = our panel.** `nativeDropdown` keeps our trigger but overlays a transparent real `<select>` so the **OS draws the menu** (native mobile pickers); omitted → our custom listbox panel.
- **Keyboard** handled on the trigger's `onKeyDown` (a document listener would re-read the opening keypress and close the panel): open on Enter/Space/↓; ↑/↓ wrap, skipping disabled; Home/End; **typeahead** (type-to-jump, disabled-skipping); Enter/Space commit; Esc/Tab close; outside-click dismiss.
- **State vs interaction split:** selected value shows a check only (no background); hover / keyboard-nav draws the accent bar. Nothing is pre-highlighted on open (`kbActive`).
- **Adaptive trigger width** via a hidden measurement clone + `ResizeObserver` (re-measures on label/locale change without the component knowing i18n).
- **Portaled, collision-aware panel.** The panel is portaled to `document.body` and positioned `fixed`, so an `overflow:hidden`/flex ancestor can no longer clip it or shove the surrounding layout. Placement is computed against a collision box = **`boundary` container ∩ viewport**: it flips above only when it won't fit below, shifts left when it would overrun the right edge, and caps the height to ~9.5 rows (a half-row "peek", à la MUI). `boundary` (a CSS selector matched via `trigger.closest`) keeps the panel inside a logical pane — ShareSelect confines its dropdown to the chat surface so it never drifts over the members column. Omitted → viewport only.
- **Grouped or flat options**, disabled options, non-string (e.g. numeric) values.
- **A11y:** `role=listbox/option`, `aria-haspopup/expanded/controls/activedescendant/selected/disabled`. Text is non-selectable (`-webkit-user-select` incl. Safari) with a default cursor throughout.

## ShareSelect
Now maps share modes → `<Select>` options + `triggerClassName="cw-share-trigger"` + `boundary=".cw-chat-surface"`; its pill shape and per-mode colors live in `.cw-share-trigger`. **Public API unchanged** — `<ShareSelect mode onChange />`.

## Adopters (in this PR)
All remaining native `<select>` surfaces now use `<Select>`:
- **ShareSelect** — adaptive pill + per-mode colors (custom panel), confined to the chat surface.
- **Automation runs filter** — status / trigger-kind. (Wrapper is a `<span>`, not a `<label>`: a `<label>` re-dispatches trusted clicks to the trigger `<button>`, so picking an option reopened the panel; the accessible name comes from `ariaLabel`.)
- **Composer model picker** — grouped by provider.
- **Schedule picker** — unit / monthly mode / nth / weekday (numeric + dynamic-value options); number/time/text inputs unchanged.

A `className` prop was added to `<Select>` to size the wrapper as a flex/grid item (e.g. the schedule selects fill their row). Per-consumer trigger looks live in `cw-share-trigger` / `cw-runs-filter-trigger` / `cw-model-picker-trigger` / `cw-sched-select`.

## Test plan
- [x] `pnpm lint` (tsc) clean
- [x] `pnpm test:unit` 67/67 — incl. 12 new `<Select>` behavior tests (`Select.test.tsx`: open/close, mouse + keyboard selection, typeahead, disabled skip, grouping, outside-click, numeric values, aria). Pixel placement (portal coords / collision / boundary) is left to e2e — jsdom has no layout.
- [x] `pnpm build` clean
- [x] Manual (driven via Chrome DevTools): adaptive width across labels/locales; keyboard open/nav/typeahead/Home-End/commit/Esc/Tab; outside-click; placement flip-up near viewport bottom; horizontal shift-in near the right edge; `boundary` keeps the share panel inside the chat surface (no longer spills into the members column); opening no longer shifts the session content; grouped + disabled + numeric; `nativeDropdown` opens the OS menu with our trigger; non-selectable text + default cursor; ShareSelect parity (colors, pill, all behaviors); runs filter closes on a real-mouse selection.

### DIFF
--- a/app/src/components/SchedulePicker.tsx
+++ b/app/src/components/SchedulePicker.tsx
@@ -6,6 +6,7 @@
 
 import { useEffect, useRef, useState } from 'react';
 import { Icon } from '@/components/Icon';
+import { Select } from '@/components/Select';
 
 export type SchedulePickerValue = { expr: string; tz: string };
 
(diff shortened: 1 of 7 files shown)


grf53 (approved):
LGTM

Now that a universal component has been added, do you think all select/dropdowns in the app should use this?

PR #135, opened 2026-06-01
fix: batch of sidebar, dialog, composer, and auth fixes

ljhh-0611:
## Summary

Eleven scoped fixes across the sidebar, dialog stack, session composer, sessions, invite, auth, and chat scroll — grouped by area, each numbered so reviewers can read them top-to-bottom. The sidebar scrollbar (#2) and chat scrollbar (#11) each took a couple of intermediate passes in the branch; those superseded commits stay in the git history but are omitted here — each section below describes only the final shape that lands against `main`.

### 1. `fix(sidebar): unread badge fits within icon footprint`

**Bug.** The session-row unread badge was a 32px-wide pill (`min-width: 32px; padding: 0 10px`), while the resting-state IconPocket is 16×16. Toggling unread shifted the entire row layout.

**Fix.** The badge keeps a 16px height and starts as a 16×16 circle for one digit, growing into a tight pill for 2–3 chars. Counts over 99 collapse to `99+` in JSX and the exact value is exposed via the `title` tooltip, so accuracy is preserved without breaking layout.

**Result.** Row baseline no longer shifts when unread state changes. Visually verified with mock unread states 1/3/99/99+ via DOM injection in the running app.

### 2. `fix(sidebar): always-visible faint scrollbar, brightens on hover, narrower gutter`

**Bug.** The sidebar reserved the OS-default ~15px scrollbar gutter on macOS Chrome, leaving an empty strip on the right, and the default thumb either disappeared at rest or read as visual noise.

**Fix.** A ChatGPT-style static treatment, packaged as a reusable `.cw-scroll-quiet` class (also applied to the sessions overlay and the main shell):
- A faint thumb (20% of `fg-3`) is always painted, so the scrollable signal never disappears at rest; hovering the scroll area brightens it to ~55%, near the OS default tone.
- `scrollbar-width: thin` is critical: it's the only CSS knob that narrows the reserved gutter on macOS Chrome when the system pref is "Show scroll bars: Always". Without it, the gutter falls back to OS default ~15px and `::-webkit-scrollbar` width is silently ignored on Chrome 121+. With `thin`, the gutter drops to ~11px — the floor reachable purely in CSS.
- Sidebar right padding cut from 10px → 4px; the gutter itself provides the visual breathing room.

**Result.** Total gap from session text to sidebar edge: 22-26px → 16px (~35% narrower). Thumb visible at all times, brightens on hover, no flickering.

### 3. `fix(dialogs): route Escape through a modal stack`

**Bug.** Every dialog attached its own window-level `keydown` listener, and pages like Files attached their own to clear selection or close menus. When a dialog opened on top of a page with a selection, pressing Escape fired both handlers — the selection cleared at the same time the dialog closed, leaving users confused about which thing they had just dismissed.

**Fix.** New `useDialogEscape(onClose, { disabled })` hook with a module-level stack and a single capture-phase window listener. The listener walks the stack top-down, picks the topmost non-disabled entry, calls `onClose`, and calls `stopImmediatePropagation` so background window-level handlers never see the event. Page handlers stay as-is; they only run when no dialog claims the press. Migrated nine dialogs and the SessionCardMenu dropdown to the new hook. ConfirmDialog and FolderPickerDialog keep their own listener for Enter shortcuts, intentionally outside the modal-stack contract.

**Result.** Verified in browser: with a file selected and the move dialog open, Escape closes only the dialog and the "1 item selected" toolbar stays. A second Escape clears the selection. Delete-confirm flow behaves the same.

### 4. `fix(composer): make session composer multi-line and guard IME Enter`

**Bug.** The session-page composer was a single-line `<input>`, so users couldn't write multi-line prompts. Worse, when typing Korean (or Japanese/Chinese), one character is assembled from several keystrokes and lives in a temporary *composing* buffer — shown underlined — until it is committed. Pressing Enter to finalize a still-composing syllable submitted the half-assembled message instead of just completing the character.

**Fix.** Two changes:
- `<input>` → `<textarea>` with the same auto-grow pattern as ProjectHomeComposer (one row at rest, capped at 200px, overflow flips to `auto` only at the cap). Enter sends; **Shift+Enter** inserts a newline.
- The keydown handler skips send while `e.nativeEvent.isComposing` is true, so an Enter that commits a composing syllable no longer fires a send with incomplete text.

**Result.** Korean users no longer lose partial input, and multi-line prompts work in session view as they already did on the project home.

> **Note on IME Enter (verified).** The original bug came from the single-line `<input>` + native form submission: pressing Enter to commit a Korean composition submitted the unfinished text. Switching to `<textarea>` removes that path (a textarea doesn't submit a form on Enter), and the `isComposing` guard keeps the new keydown handler from re-firing send on the mid-composition keystroke. Verified on macOS Chrome, Edge, and Safari via an event probe: a single Enter sends the fully-composed text exactly once — no double-send and no "press Enter twice" behavior.

### 5. `fix(composer): clear live messages even when post-send refetch fails`

**Bug.** The `send()` finally-block awaited a messages refetch before clearing `liveMessages`. If the backend went down, the refetch itself threw, the throw aborted the rest of the finally block, and the empty streaming AI bubble stayed parked on screen until full refresh.

**Fix.** Wrap the refetch in its own nested `try/finally` so `setLiveMessages([])` always runs — the success path still waits for the refetch (no cross-fade flash from optimistic bubbles disappearing before persisted ones arrive), and the failure path clears immediately.

**Result.** Sending a message with the backend down now leaves a clean failure state: a toast, no orphan bubble, ready for retry.

### 6. `fix(composer): morph session composer between compact pill and stacked layout`

**Bug.** Even after #4 (multi-line + IME guard), the compact-pill layout walled off the right portion of the textarea with the clip + send buttons. Long text either wrapped into a tiny remaining width or felt cramped — the buttons effectively limited how comfortable a long prompt could be.

**Fix.** Match the ChatGPT pattern: stay compact (pill, inline buttons) at rest, and the moment text would wrap to a second row in the inline-width context, morph into a stacked layout — textarea spans full width, clip + send drop to a right-aligned row beneath. The flip is stable: once expanded, the layout stays put until the field empties (send or manual clear), so partial edits don't oscillate. JSX wraps clip + send in `.cw-composer-actions` so a single `flex-direction` toggle swaps between row (compact) and column (expanded). CSS animates `border-radius` (24px ↔ 18px) and padding over 200ms so the morph reads as a single gesture even though `flex-direction` itself is instant. The auto-grow effect reads computed `line-height + padding` to decide when to flip — using styles instead of magic constants means the trigger survives typography changes. In expanded mode the textarea switches from `flex: 1` to `flex: 0 0 auto` so the column flex container doesn't collapse it back to its min-height.

**Result.** Verified visually: empty → compact pill, short text → still compact, long text that would wrap → smooth morph into stacked layout, empty again → smooth morph back. Korean-IME and multi-line behavior from #4 still work inside the expanded layout.

### 7. `fix(sessions): refresh unread badge after re-entering within 30s`

**Bug.** Session entry triggers the messages query, which fires the backend mark-read side effect. But the QueryClient's default `staleTime` of 30s meant re-entering the same session within that window used cached data — no refetch, no backend mark-read, no `dataUpdatedAt` change, no sidebar invalidate. The unread badge stayed visible even though the user clearly opened the session.

**Fix.** Set `staleTime: 0` on the per-session messages query. Every entry refetches, the backend marks read, and the existing `dataUpdatedAt` effect invalidates the sidebar query as designed.

**Result.** Sidebar unread badge clears on every session entry, not just the first entry per 30s window.

### 8. `fix(invite): rename InviteDialog projectId prop to projectRef`

**Bug.** The InviteDialog prop was named `projectId` but every call site passed a slug. The cache-invalidation key (`['members', projectId]`) happened to match because the members page also keys by slug, so the bug never surfaced. A future caller passing the actual UUID would silently miss the invalidate and the members list wouldn't refresh after an invite.

**Fix.** Rename the prop to `projectRef` to align with main's convention (backend's `resolve_project_id` accepts either UUID or slug), and update the single call site.

**Result.** The convention foot-gun is removed; future refactors that touch project identifier types won't silently break invalidation.

### 9. `fix(auth): guard forceLogout against concurrent 401 callers`

**Bug.** When the JWT expired mid-session, multiple in-flight requests (e.g. SSE + REST) could hit 401 within the same tick. `notifyUnauthorized` had an `unauthorizedTriggered` flag, but `forceLogout` itself was re-entrant — a second caller could overwrite REASON_KEY in sessionStorage (e.g. `expired` → `invalid`), surfacing the wrong banner.

**Fix.** Module-level `logoutInProgress` flag short-circuits repeat calls. New `resetLogoutGuard()` runs on successful re-login so the next expiry cycle can fire normally.

**Result.** The first 401 wins the logout reason; subsequent 401s are absorbed silently.

### 10. `fix(login): give form inputs name attributes for password managers`

**Bug.** The browser console warned `A form field element should have an id or name attribute (count: 2)` because the login form inputs had neither. The form worked (React-controlled state, submit via `onSubmit`), but password managers and autofill rely on `name` for reliable field identification.

**Fix.** Add explicit `name` attributes — `backendUrl`, `username`, `displayName`, `password` — alongside the existing `autoComplete` hints.

**Result.** No more console warning. 1Password / Chrome / etc. can save and fill credentials reliably.

### 11. `fix(chat): move scroll to full-width wrapper so scrollbar sits at chat panel edge`

**Bug.** `.cw-messages` was both the scroller (`overflow-y: auto`) and the centered max-width column (`max-width: 860px; margin: 0 auto`). On any viewport wider than the column, the scrollbar rendered at the right edge of the centered column — floating mid-air between the bubbles and the side panel — instead of at the chat panel's right edge like other AI chat apps.

**Fix.** Split scroll vs. centering into two layers:
- `.cw-messages-scroll` is a new full-width wrapper that owns `overflow-y: auto`. The scrollbar lives at this wrapper's right edge → always flush with the chat panel boundary regardless of inner content width.
- `.cw-messages` keeps `max-width: 860px; margin: 0 auto` for the centered bubble column but no longer scrolls itself.

Default OS scrollbar size is intentional for the chat scroller — this is the primary engagement surface and the chunky thumb gives a clearer affordance; the thin/quiet treatment (#2) stays on the sidebar.

**Result.** Scrollbar sits exactly where the user expects — flush with the chat panel's right edge, immediately before the right side panel — regardless of viewport width.

## Verification

- TypeScript clean (`pnpm exec tsc --noEmit`) at every step.
- Ran backend (port 8091) + Vite (port 4120) against an isolated `data-test` dir; each fix verified visually in the browser.
- Escape modal-stack confirmed via Chrome DevTools: with a file selected and the `이동` dialog open, Escape closed the dialog and preserved the "1개 선택됨" toolbar; a second Escape cleared the selection.

## Out of scope (filed as separate issues)

Discovered during this round and intentionally deferred — each needs separate design discussion or larger scope:

- #136 — File upload silent overwrite when filenames collide
- #137 — Cache key identity: WS payload uses `project_id` (UUID) but every TanStack Query key is `projectSlug`
- #138 — `AppState::agents` has no TTL eviction; idle sandboxes accumulate
- #139 — Session fork creates empty `artifacts/inputs` host dirs
- #140 — `/documents` is unauthenticated with no body limit (DoS surface)

## Test plan

- [x] Verify sidebar unread badge with multiple counts (1, 3, 99, 99+)
- [x] Verify sidebar scrollbar always visible (faint), brightens on hover, gutter narrower
- [x] Verify chat scrollbar at the chat panel's right edge, default size, regardless of viewport width
- [x] Verify Escape closes the move dialog without clearing file selection; a second Escape clears it
- [x] Verify Escape closes the delete confirm without clearing file selection
- [x] Verify composer morphs: empty → compact, long text → stacked, cleared → compact
- [x] Verify TypeScript compile is clean
- [x] Korean IME — one Enter sends the fully-composed text exactly once (verified on macOS Chrome / Edge / Safari)
- [x] Manual: multi-line prompt in session composer (Shift+Enter)
- [x] Re-enter a session within 30s — messages query refetches on every entry (verified 2 `GET /messages` for 2 entries < 30s apart; `staleTime: 0` defeats the default 30s cache), which drives the backend mark-read + sidebar invalidation that clears the unread badge
- [x] Kill backend, send message — optimistic bubble appears then clears, "전송 실패" toast shown, no orphan AI bubble (verified with simulated backend-down: peak 2 bubbles → final 0)
- [x] Login & signup inputs all carry `name` + `autoComplete`, and the "form field should have an id or name attribute" console warning is gone (enables 1Password / Chrome credential save)


grf53 (approved):
Let's merge after resolving conflicts

PR #144, opened 2026-06-02
fix(session): copy host artifacts/inputs on fork (#139)

khj809:
## Summary

Fixes #139. `POST /sessions/{id}/fork` was creating **empty** `artifacts/` and `inputs/` host directories for the forked session. Since the coworker agent bind-mounts those host paths over `/workspace/artifacts` and `/workspace/artifacts`, any files produced by the source session were invisible inside the fork — and the UI artifacts panel showed an empty folder.

- **Root cause**: `fork_session` called `create_dir_all` on fresh empty paths instead of copying the source session's files.
- **Fix**: recursively copy the source session's `artifacts/` and `inputs/` host directories into the fork immediately before sandbox creation, matching the point-in-time snapshot semantics of the `upper.ext4` clone. `shared/` is project-level and intentionally excluded.

## Changes

| File | Change |
|------|--------|
| `backend/src/handlers/dirent.rs` | Expose existing `copy_dir_recursive` as `pub(crate)` — no logic change |
| `backend/src/handlers/session.rs` | Replace empty `create_dir_all` loop with per-directory recursive copy; falls back to `create_dir_all` when the source dir is absent (e.g. session never ran) |
| `backend/tests/session_test.rs` | Add two new tests (see below) |

## New Tests

**`fork_copies_host_files`**
- Seeds the source session's `artifacts/` and `inputs/` with real files.
- Asserts the forked session's host directories contain the same files with matching content.
- Asserts snapshot independence: a file added to the source *after* the fork does not appear in the fork.

**`fork_empty_session_succeeds`**
- Forks a session whose host directories were never created.
- Asserts the fork still returns `201` and that empty `artifacts/` and `inputs/` directories are created (required for bind-mount).

## Test plan

- [x] `cargo test fork_copies_host_files` passes
- [x] `cargo test fork_empty_session_succeeds` passes
- [x] Fork a real session with artifacts in the UI — artifacts panel shows files in the forked session
- [x] Verify files added to the source after forking do not appear in the fork

### DIFF
--- a/backend/src/handlers/dirent.rs
+++ b/backend/src/handlers/dirent.rs
@@ -808,7 +808,7 @@ async fn find_available_name(parent: &Path, base_name: &str) -> PathBuf {
 /// untrusted input), and any symlink encountered is skipped, so a malicious
 /// symlink planted inside the source tree cannot escape `dst`'s subtree.
 // nosemgrep: rust.actix.path-traversal.tainted-path
-async fn copy_dir_recursive(src: &Path, dst: &Path) -> std::io::Result<()> {
+pub(crate) async fn copy_dir_recursive(src: &Path, dst: &Path) -> std::io::Result<()> {
     tokio::fs::create_dir_all(dst).await?;
     let mut rd = tokio::fs::read_dir(src).await?;
     while let Some(entry) = rd.next_entry().await? {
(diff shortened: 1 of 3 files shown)


grf53 (approved):
Looks clear.

ljhh-0611 (approved):
🫡

PR #134, opened 2026-06-01
feat(ws): Migrate agent response streaming from sse to websocket

khj809:
## Summary

Replaces the SSE-based agent response delivery (`POST /sessions/{id}/messages/stream`) with a WebSocket fan-out architecture so that **all users connected to the same session receive real-time agent output simultaneously**.

### Problem

The previous `send_message_stream` handler held the agent lock and streamed tokens exclusively back to the HTTP connection of the requesting user. Other members with access to the same session had no way to observe the run in real time.

### Architecture

```
POST /sessions/{id}/messages  →  202 Accepted (fire-and-forget)
                                   │
                              tokio::spawn
                                   │
                         agent.run() loop
                                   │
                      state.push_output()  ──►  broadcast (ws_tx)
                                                     │
                                           WS clients filtered by
                                           subscribed session set
```

- **Single `/ws` endpoint multiplexed** — clients send `{ "action": "subscribe"|"unsubscribe", "session_id": "..." }` control frames; the server filters broadcasts per-session.
- **Fire-and-forget HTTP trigger** — `POST /sessions/{id}/messages` returns `202 RunAck` immediately; agent execution runs in a background `tokio::spawn` task. The requester disconnecting does not abort the run.
- **`ActiveAgentRun` buffer** — an in-process `DashMap<Uuid, Arc<RwLock<ActiveAgentRun>>>` accumulates `(seq, MessageOutput)` pairs for the duration of a run, enabling catch-up snapshots for mid-session joiners.
- **Catch-up on subscribe** — when a client subscribes to an in-progress session, the server immediately replays `AgentRunStarted` + all buffered `AgentMessage` frames. Subscription is inserted *before* the snapshot read to avoid missing live broadcasts during replay.
- **`AgentRunIdle` sentinel** — if subscribe finds no active run, the server sends `AgentRunIdle` so clients stuck in `streaming=true` (after reconnect or server restart mid-run) can reset UI state.
- **Idempotent client merge** — the frontend accumulates outputs in a `Map<number, MessageOutput>` keyed by `seq`; `Map.set` is idempotent, so catch-up frames and live frames can interleave safely.
- **Panic recovery** — a monitor `tokio::spawn` awaits the agent task's `JoinHandle` and forces `end_run` + `AgentError` broadcast if the task panics.

### Changes

**Backend**
[...]

PR #147, opened 2026-06-04
Better scrolls on session pages

grf53:
## Related PR
#134 : This PR continues from the [comment](https://github.com/brekkylab/agent-k/pull/134#issuecomment-4622397584) in #134. We have changed the auto-scroll criteria applied in that PR.(And added several other scroll-related features.)

## Changes
- **Widened auto-follow threshold** (150px → 700px): new messages smooth auto-scroll only when within 700px of the bottom
- **Instant jump on own send**: sending a message scrolls straight to the bottom (no animation), even if scrolled far up
- **Start at the bottom on session entry/refresh**: first history load jumps instantly to the latest message
- **New-message pill**: when a message arrives while scrolled far up, a pill appears centered above the composer — click jumps instantly to the bottom; auto-dismissed when the user scrolls back near the bottom
image

- **Ghost scroll-down button**: a translucent circular button appears at the right end above the composer whenever scrolled past 700px — click jumps instantly to the bottom
image

- Misc: new i18n keys (`new_message`, `scroll_to_bottom` en/ko), related CSS (`.cw-new-msg-pill`, `.cw-scroll-down-btn`)

### DIFF
--- a/app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
+++ b/app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
@@ -92,6 +92,9 @@ function SessionPage() {
   const messagesEndRef = useRef<HTMLDivElement>(null);
   const scrollContainerRef = useRef<HTMLDivElement>(null);
   const fileInputRef = useRef<HTMLInputElement>(null);
+  // Set when the local user sends a message — the next scroll effect jumps to
+  // the bottom instantly, regardless of how far up the user has scrolled.
+  const forceScrollRef = useRef(false);
 
   // WS 구동: seq 순서로 정렬된 outputs 맵 (catch-up + live 멱등 병합용)
   const wsOutputsRef = useRef<Map<number, MessageOutput>>(new Map());
@@ -142,6 +145,7 @@ function SessionPage() {
     optimisticUserIdRef.current = null;
     currentRunIdRef.current = null;
     doneRunIdsRef.current.clear();
+    forceScrollRef.current = false;
   }
 
   // After messages load, mark-read side effect has run on the backend — sync badge in session list.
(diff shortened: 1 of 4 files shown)


app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -159,12 +174,53 @@ function SessionPage() {
...
+
     const distanceFromBottom = el.scrollHeight - el.scrollTop - el.clientHeight;
-    if (distanceFromBottom <= 150) {
+    if (distanceFromBottom <= 700) {
       messagesEndRef.current?.scrollIntoView({ behavior: 'smooth' });
+    } else if (allMessages.length > prevCount) {
+      // Scrolled too far up to auto-follow — surface a "new message" pill instead.
+      setShowNewMsgPill(true);
     }
   }, [allMessages.length, streaming]);
```
ljhh-0611: **critical — auto-follow is dead during streaming.** `agent_message` mutates only the AI bubble's `body`; `allMessages.length` doesn't change mid-stream and `streaming` stays `true`, so neither dep changes → the effect never re-runs and the view stops following token growth even when pinned to the bottom.

```suggestion
  }, [allMessages.length, streaming, allMessages.reduce((n, m) => n + m.body.length, 0)]);
```

Re-fires on body growth. Verified locally: typechecks, near-bottom follows, scrolled-up (>700px) still isn't yanked down. `ResizeObserver` on the container is the heavier-but-robust alternative.

grf53: Thank you for suggestion!

I added another term related to the tool calls. Tool-call rows appearing and tool results arriving grow the rendered content without changing any message body or count, so they must bump `contentVersion` too.
Please check the commit 3ffe265 to see that.

ljhh-0611 (approved):

PR #148, opened 2026-06-04
Stop runs on session pages

grf53:
## Description
Once `POST /sessions/{id}/messages` fires a run, there was no way to interrupt — every client can only wait for the run to finish.

Adds **stop-run** button (■) for in-progress agent runs. Only the tab that sent the message can stop; output up to the cancellation point is **persisted**, and dangling tool calls are sealed so the history stays valid for the next turn.

image

## Design

```
■ click → POST /sessions/{id}/runs/{run_id}/stop → 202
            └─ CancellationToken.cancel()  (sender-only: 403 otherwise, 404 stale/no run)
                 │
   run loop (biased select!):
    ├─ model call in flight → wait ≤10s (grace) → broadcast + persist it → stop
    └─ tool executing       → cut immediately → seal dangling tool_calls
                 │
   persist partial → end_run → AgentRunDone { stopped: true } → all tabs clean up
```

- **Sender-tab scoping** — the stop button is enabled **only in the tab that sent the message**. `RunAck` now carries `run_id`; only the HTTP response reaches the sending tab (spectators learn run_ids via WS only), so `streaming && ownedRunId` gates the stop button with no shared state. Server enforcement is user-level (run starter only).
- **10s grace** — ailoy's `model.run()` is a single non-streaming POST; dropping it mid-await only closes our side while the provider completes (and bills) the generation, so discarding would leave a record/UI divergence. Instead we wait up to 10s for the in-flight message and persist it. This is a workaround for the lack of token streaming in ailoy — **once ailoy streams tokens, the grace period goes away entirely**: stop cuts immediately and persists the partial text.
- **Sealing** — unanswered `tool_calls` break the next model call (in-memory and DB replay alike), so they're closed with synthetic `Tool` messages in both.
image
(Sealing message is changed into english now.)

## Changes

**Backend**
- `state.rs` — `ActiveAgentRun.cancel: CancellationToken`; `start_run()` → `(run_id, token)`; `run_cancel_info()`
- `handlers/session.rs` — run loop → biased `select!` + grace; tool_call sealing; `stop_run`; `RunAck { status, run_id }`
- `events.rs` / `handlers/ws.rs` — `AgentRunDone.stopped`
- `router.rs` — `POST /sessions/{session_id}/runs/{run_id}/stop`

**Frontend**
- `messages.ts` (`RunAck`, `stopRun()`) · `ws.ts` (`stopped`) · session page (`ownedRunId`/`stopping`, send↔stop button swap, cleanup + toast) · `Icon.tsx` (`stop` glyph)

**Tests**
- `stop_run_authz_and_lifecycle`: no run → 404 / non-sender → 403 / stale run_id → 404 without cancelling / sender → 202 + token cancelled

## Known limitations

- Grace timeout (>10s model call) falls back to discard (warn-logged).
- Tool side effects survive a stop (ailoy runs tools on detached tasks) — pre-existing behavior.
- Automation runs keep their own lease-based cancellation (their sessions are never continued).

## Test Plan

- [x] ■ appears only in the sender tab; spectators see none
- [x] Stop mid-generation → in-flight message lands (≤10s) in both tabs, then stops with toast
- [x] Stop during tool execution → immediate stop; sealed tool result visible after refetch
- [x] Re-send works immediately after stop (no stuck 423)
- [x] Non-sender API stop → 403; finished run_id → 404
- [x] `cargo test` / `tsc` / `vitest` green


app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -358,46 +414,72 @@ function SessionPage() {
...
+    } catch (err) {
+      // 404 means the run already ended — reconcile instead of surfacing an error.
+      if (err instanceof ApiError && err.status === 404) {
+        resetEndedRun();
+        return;
+      }
+      const msg = err instanceof ApiError ? err.message : err instanceof Error ? err.message : 'stop failed';
+      showToast(t('toast.stop_failed', { message: msg }));
+      setStopping(false);
+    }
```
khj809: I think `setStopping(false)` should be in `finally` block, so it can be always stopped regardless of the result of `stopRun()`.

grf53: Currently `setStopping(false)` should only be in `catch`, because we do graceful stop here so need to maintain `stopping` as `true` until the run really stopped. It is not designed to be tightly coupled.

But thanks to your concern, I found that there is no guarantee to turn `stopping` back into `false` now. (especially it can be stuck when the 'end' event is missed)
So I reused the `armWatchdog()`(that runs `resetEndedRun()` if no ws events arrive during timeout seconds) for that to make sure the `stopping` eventually get back into `false`.
(20e18cd)

ljhh-0611 (approved):
👍

backend/src/handlers/session.rs
khj809: If `run_error` is set during the stop grace period, it always truncate the history and return the `AgentError` instead of the clean stop.
Suggested fix: just skip this when `stopped` is true.

```suggestion
        if let Some(err) = run_error {
            if stopped {
                // The error occurred during the grace period after the user requested a stop.
                // Treat it as a clean stop rather than a hard failure: keep whatever partial
                // outputs were already accumulated instead of rolling back the entire turn.
                tracing::warn!(%session_id, "agent error during stop grace period (treating as stop): {err}");
            } else {
                tracing::error!(%session_id, "agent run failed: {err}");
                // Truncate in-memory history to match DB state so the agent stays consistent.
                agent.state.history.truncate(prev_len);
                drop(agent);
                state2.end_run(&session_id);
                let _ = state2.ws_tx.send(WsEvent::AgentError {
                    session_id: session_id.to_string(),
                    run_id: run_id.to_string(),
                    message: err,
                });
                return;
            }
        }
```

backend/src/handlers/session.rs
```diff
@@ -977,6 +1017,60 @@ pub async fn send_message(
...
+                }
+            }
+            // Skip the note if the model finished its turn anyway (last output is an
+            // assistant message with FinishReason::Stop) — not truncated.
+            let completed_naturally = matches!(
+                depth0_outputs.last(),
+                Some(o) if o.message.role == Role::Assistant
+                    && matches!(o.finish_reason, FinishReason::Stop {})
+            );
+            if !completed_naturally {
```
khj809: If the stop request and natural completion occur at the same time, `stopped` is not reset to `false` and users may see "Run stopped" toast even on normal completion.
```suggestion
            );
            if completed_naturally {
                // Cancel fired at the exact moment the model finished; treat as a normal
                // completion so the client doesn't see a spurious "Run stopped" toast.
                stopped = false;
            } else {
```

khj809 (approved):
Looks good overall, but left some rare edge cases below.

PR #150, opened 2026-06-05
feat: resync run state on session re entry

selloriwoo:
## What
- Re-sync a session's agent run state with the server every time you (re-)enter it.
- On entry the client now re-requests the WS catch-up replay, so it restores either the in-progress stream (`agent_run_started` + buffered messages) or the latest history (`agent_run_idle`).

## Why
- Run state was only replayed on the WS `subscribe` action, but `subscribeSession` is ref-counted and the sidebar permanently subscribes to every session — so re-entering a session never re-triggered a replay.
- Result: switching away from a session mid-run and back left the message/reply hidden until the next send.

## How
- Add `resyncSession()` — sends `subscribe` bypassing the ref count (the server's handler is idempotent and replays every time).
- Call it on session entry, alongside the ref-counted `subscribeSession`.
- Refetch history unconditionally on `agent_run_idle`.
- previous `subscribeSession` manages broadcast membership across all sidebar sessions; `resyncSession` triggers a replay for just the currently-viewed session.

## Test
- [x] Manually, dev stack: send in A → switch to B → return to A. Reply is now restored both mid-stream and after completion.

Resolved #145


grf53 (approved):
LGTM

btw, would you remove the comments in Korean, or replace them into English, before merge?

PR #153, opened 2026-06-07
Improve sidebar session list

grf53:
## Changes

- **`GET /sessions` now accepts `?origin=user|automation`** — the repository list
  queries filter by origin, so each surface fetches only the sessions it renders.
- **Sidebar / sessions overlay / project cards** switch to `['sessions', slug, 'user']`
  with the server-side `origin=user` filter, dropping client-side `origin` filtering.
- **Unread badge updates in place**: on session entry the selected session's
  `unreadCount` is zeroed directly via `setQueriesData` instead of invalidating
  (and refetching) every cached session list.
- **"Mark as read" in the sidebar session menu** — `POST /sessions/{id}/read`
  marks a session read without fetching history (reusing the existing
  `mark_session_read` repository call), and the badge is zeroed in place via the
  same `setQueriesData` path.

## Notes

- Existing `['sessions', slug]` prefix invalidations (create/delete/duplicate/run-done)
  still match the new origin-suffixed keys.

## Test plan

- [x] `cargo test` — including new `list_sessions_filters_by_origin`
  (user/automation/no filter/invalid → 400)
- [x] `tsc --noEmit`

### DIFF
--- a/app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
+++ b/app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
@@ -204,9 +204,14 @@ function SessionPage() {
   // instead of refetching whole lists on each session entry.
   useEffect(() => {
     if (history.isSuccess && sessionId) {
+      void queryClient.cancelQueries({ queryKey: ['sessions', projectSlug] });
       queryClient.setQueriesData<Session[]>({ queryKey: ['sessions', projectSlug] }, (old) => {
         if (!old?.some((s) => s.id === sessionId && s.unreadCount > 0)) return old;
         return old.map((s) => (s.id === sessionId ? { ...s, unreadCount: 0 } : s));
+      });
+      void queryClient.invalidateQueries({
+        queryKey: ['sessions', projectSlug],
+        refetchType: 'none',
       });
     }
   }, [history.isSuccess, history.dataUpdatedAt, projectSlug, sessionId, queryClient]);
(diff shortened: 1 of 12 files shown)


app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -199,12 +199,17 @@ function SessionPage() {
...
+  // session's unread badge directly in every cached session list (any origin filter)
+  // instead of refetching whole lists on each session entry.
   useEffect(() => {
-    if (history.isSuccess) {
-      void queryClient.invalidateQueries({ queryKey: ['sessions', projectSlug] });
+    if (history.isSuccess && sessionId) {
+      queryClient.setQueriesData<Session[]>({ queryKey: ['sessions', projectSlug] }, (old) => {
+        if (!old?.some((s) => s.id === sessionId && s.unreadCount > 0)) return old;
+        return old.map((s) => (s.id === sessionId ? { ...s, unreadCount: 0 } : s));
+      });
```
khj809, line 211: Considering the case when `mark_session_read()` has been failed on backend, suggest to mark stale so the next background refetch reconciles if it silently failed.
Also suggest to cancel in-flight queries in advance, so the result cannot be overwritten by the previous unreadCount.


```suggestion
      void queryClient.cancelQueries({ queryKey: ['sessions', projectSlug] });
      queryClient.setQueriesData<Session[]>({ queryKey: ['sessions', projectSlug] }, (old) => {
        if (!old?.some((s) => s.id === sessionId && s.unreadCount > 0)) return old;
        return old.map((s) => (s.id === sessionId ? { ...s, unreadCount: 0 } : s));
      });
      void queryClient.invalidateQueries({
        queryKey: ['sessions', projectSlug],
        refetchType: 'none',
      });
```

khj809 (approved):
LGTM

PR #154, opened 2026-06-07
feat(session): paginate message history by turn with seq cursors

grf53:
## Changes

- **`GET /sessions/{id}/messages` supports cursor-based pagination in turn units**
  (a user message + the agent/tool messages that follow it):
  - `limit` + `before_seq` — newest N turns below the cursor (older pages)
  - `after_seq` — tail catch-up with every message after the cursor
  - windows never split a turn and are immutable under appends
- **Responses carry `seq`** for stable client-side message identity across windows.
- **mark-read only fires on tail requests** (no `before_seq`) — paging older
  history doesn't clear unread.
- **Session page loads progressively**: latest turn window on entry (scrolled to
  bottom), and older pages load when the user scrolls within **120px of the top**
  of the message list (scroll position preserved across the prepend; pages also
  auto-fill while the list is shorter than the viewport).
- **Composer keeps focus after sending** — the textarea is no longer disabled while
  a run streams, so focus stays in the input (and you can draft the next message
  during the response); the send button and send guard still block the actual send
  until the run completes.

## Fetch flow
message-pagination-flow

## Notes

- Pages merge oldest→newest and collapse tool messages once over the merged
  array, so `tool_call`/result pairs survive window boundaries; `seq` dedupe
  absorbs overlap from stale `after_seq` refetches.
- While streaming, the new turn renders via WS `liveMessages` only — the tail
  fetch happens after the run persists.

## Test plan

- [x] `cargo test` — cursor windows, whole-turn boundaries (incl. preamble &
  agent-only histories), `before_seq`+`after_seq` → 400, mark-read gating
- [x] `tsc --noEmit`
- [x] Manual test: Look into the network tab in browser dev tool during 'entering the seesion page', 'scrolling up to top', 'sending a new message(start a new turn)'

#### Enter session page
Get messages with latest 10 turns. (seq: 27-60)
<details>
  <summary>Show</summary>
스크린샷 2026-06-08 오전 11 08 38
</details>

#### Scroll up to top
Get messages of 10 turns _before_ the oldest message(seq: 27) that the page had
<details>
  <summary>Show</summary>
스크린샷 2026-06-08 오전 11 09 15
</details>

#### New turn
Get all messages _after_ the newest message(seq: 60) that the page had
<details>
  <summary>Show</summary>
스크린샷 2026-06-08 오전 11 10 25
</details>

### DIFF
--- a/app/src/api/backend-types.ts
+++ b/app/src/api/backend-types.ts
@@ -104,6 +104,8 @@ export type BackendMessageSender =
   | { kind: 'agent'; name: string };
 
 export interface SessionMessageItem {
+  /** Session-global insertion order — stable identity across paginated windows. */
+  seq: number;
   message: AiloyMessage;
   sender: BackendMessageSender;
   created_at: string;
--- a/app/src/api/messages.ts
+++ b/app/src/api/messages.ts
@@ -1,5 +1,5 @@
 import { request } from './client';
-import type { AiloyMessage, AiloyPart, AiloyToolCall, MessageOutput, SessionMessageList } from './backend-types';
+import type { AiloyMessage, AiloyPart, AiloyToolCall, MessageOutput, SessionMessageItem, SessionMessageList } from './backend-types';
 import { aiMessageText, collapseToolMessages } from './transformers';
 import type { Message } from '@/domain/types';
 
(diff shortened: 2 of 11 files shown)


khj809 (approved):
Confirmed there's no critical issues here. LGTM

PR #161, opened 2026-06-09
feat(preview): in-app file preview (PDF, image, HTML, Markdown, code, text, CSV)

ljhh-0611:
## Summary

Adds in-app **file preview** to the Cowork SPA: a JWT-authenticated, blob-backed
lightbox that opens from Files rows (double-click / row menu), the Artifacts
panel, AI-message artifact chips, and user attachment chips — no native `src`
embedding, so previews respect the bearer-token file API.

## Supported types

| Kind | Renderer |
|------|----------|
| Image (png/jpg/gif/webp/…) | zoom + drag-pan stage |
| PDF | `react-pdf` (pinned pdf.js worker), multi-page, zoom |
| HTML | sandboxed iframe (`allow-scripts`, live render) |
| Markdown | `react-markdown` on a readable sheet |
| Code | syntax-highlighted sheet |
| CSV/TSV | parsed table (`papaparse`, RFC-4180) |
| Plain text | mono sheet |
| Anything else | fallback card with download |

A separate `resolvePreviewKind()` taxonomy (independent of the file-list
`categorise()`) maps extensions to renderers.

## UX

- **Google-Drive-style lightbox**: the whole screen dims, the content floats,
  clicking the dim dismisses. Image/PDF render full-bleed on a dark stage;
  html/markdown/code/text/table render on a light "sheet".
- **Zoom** (image + PDF): a floating pill with −/＋ and a reset control.
- **Image zoom rework** (latest commit): a single fit-based model —
  `fitWidth × scale` measured on load — replaces the old dual model that made
  "100%" collapse to a thumbnail for portrait images and clamped every level
  ≥125% to one size. Plain mouse wheel zooms images; drag-to-pan with a move
  cursor when the image overflows (scrollbars hidden, mouse-only so touch keeps
  native scrolling); the image zoom readout is a magnifier reset-to-fit button.
  PDF keeps Ctrl/pinch-zoom, its scrollbars, and the % readout.
- **Responsive sheet width**: the light sheet sizes with `clamp()` instead of a
  fixed cap, so it breathes with the viewport (per-type ceiling — prose keeps a
  readable measure, code/HTML/table get more room); the full-file code preview
  now renders flush on the sheet instead of inside a nested card.

## Safety

- **Oversized-file guard**: a client-side `Content-Length` pre-check rejects
  files over the cap (**50 MiB** media / **5 MiB** text) before downloading the
  body, showing a "too large" fallback. No backend change. Caps bound
  client-side *render* cost (we fetch the original and render it), not storage —
  image/pdf/html decode off the main thread so they get headroom; text-family
  previews decode + highlight synchronously so they stay tighter.
- HTML preview runs in a sandboxed iframe; the modal's focus trap deliberately
  excludes the iframe so ESC/Tab keep working.

> [!IMPORTANT]
> **Follow-up — large PDFs need Range streaming, not a bigger cap.**
> Previews fetch the *original* file into the browser, so even the raised
> 50 MiB cap still blocks the 30–100 MB PDFs that are common in practice.
> Bumping the number further is not the fix — the right approach is `pdf.js`
> HTTP **Range requests** (fetch only the viewed pages; `206 Partial`), which
> makes a PDF's size irrelevant. That needs the backend to serve
> `Accept-Ranges`/honor `Range` (today the dirent endpoint returns a
> fully-buffered `Vec`, so ranges aren't honored) and a PDF-only fetch path
> (`url` + `range: true` instead of blob→objectURL). **Tracked as a separate
> task — out of scope for this PR.**

## Screenshots

1. Text Types
 - plain text
image

 - code type text
image

 - markdown text
image

2. live html
image

3. csv
image

5. pdf (with zoom)
image

6. image (with zoom)
 - jpeg 
image

 - gif
 image

7. other (docx, pptx, ...)
image

8. huge file(text 5MiB or image/pdf 50MiB)
image

9. In Session
 - user attachment
 image

 - agent reply artifact
image

 - artifact in sidebar
image

## Testing

- `tsc` clean; **112** vitest unit/component tests pass (kind resolver, table
  parsing, modal fallbacks, zoom state). Playwright e2e covers
  image/html-sandbox/pdf/fallback and is excluded from the vitest run.
- A jsdom `ResizeObserver` stub lets the size-measuring `ImageView` mount under
  component tests.
- **Live-verified in Chromium** via `getBoundingClientRect` measurement, which
  is what actually caught the layout bugs unit tests/reviews missed:
  double-click off-by-one (header reflow), PDF top-of-page clipping, and the
  image zoom model (linear scaling — no 100% thumbnail, no ≥125% clamp; wheel
  zoom; drag-pan scroll deltas; magnifier reset).

## Notes

`routeTree.gen.ts` untouched. No DB/schema changes.


grf53 (approved):
I tested with some formats like PDF, HTML, Markdown, Python code, CSV.
And there was no big difference with the screenshots, as I saw.

I believe some visual improvements could be done later.
Thanks for quick implementation!

PR #143, opened 2026-06-02
Model selection

grf53:
## Ticket
resolves #120 

## Background
Different agent surfaces suit different model tiers, and which models can actually run depends on which providers are configured on the server. So the model is **recommended by default** (resolved per agent type at run time) but **user-overridable**, and the UI reflects real provider availability.

### Recommendation chains & model resolution
- A catalog of selectable models grouped by capability tier (light / standard / max);
  tiers are display-only groupings, never selected directly.
- Each agent type (`coworker`, `speedwagon`, `deep-research`, `buddy`) has an ordered,
  provider-diverse recommendation chain. Resolution picks an explicit pin if its
  provider is available, else the first available chain entry, else the chain's
  terminal default.
- "Recommended" is an explicit, selectable option (`model = NULL`), resolved at
  agent-build time from the agent type's chain and the server's configured providers.
- Once a recommended session/run is first built, the resolved model is **materialized**
  so it stays stable if the chain or provider availability later changes.
- **Per-project custom chains**: owners can reorder/override the recommendation chain
  per agent type (project settings), provider-agnostic.
- Applied across:
  - **Home composer** — agent picker + model picker, with a send guard when no model can run.
  - **Automations** — agent_type/model chosen on create/edit; triggered runs copy them
    onto the session they create; the run detail shows the run's agent and model.
  - **Session view** — header shows the session's agent and the model in use.

### New `buddy` agent
- Adds a lightweight, non-sandboxed conversational `buddy` agent surface and wires it
  into dispatch/recommendations.

> [!NOTE]
> Buddy was created arbitrarily instead of a placeholder, and no thought was given to the system prompt or design ('no sandbox' etc). It may be replaced with something completely different in the future.

### Client-side model preferences
- A user's model pick is remembered per project + agent surface in `localStorage`
  (no DB/user table), so switching agents or reloading restores the last choice.

> [!NOTE]
> This can be replaced if it is decided to add a feature for setting preferred models per user.

### Internationalization
- Applies the existing react-i18next setup to the automation/model-selection UI via a
  new `automation` namespace (en/ko), with the recommendation-chain editor strings
  living in the `project` (settings) namespace.

### Misc UI / cleanup
- standalone vs. attached agent-picker variant, run-detail meta formatting (`when · status(duration)`), pager arrow consistency, removed a redundant header button.
- Renamed the cron trigger label `recurring` → `schedule`.
- Removed dead `schedule`/`skills` placeholder routes and unused domain types.

## Schema & API
[...]

PR #164, opened 2026-06-09
fix(backend): make resolve-model test env-independent

nuri-yoo:
## Problem

`model::llm::tests::resolve_always_returns_a_catalogued_model` fails whenever the environment has an `OPENAI_API_KEY` set. It passes when that key is absent, so it passed in  #143.

## Root cause

The test's `bogus` case pins `openai/does-not-exist` and asserts the resolver ignores it and falls through to the chain:

```rust
let bogus = resolve_model(Some("buddy"), Some("openai/does-not-exist"));
assert!(catalog_entry(&bogus).is_some());
```

`provider_available` is a provider-level glob check: a provider is "available" when its API key is set, which ailoy reflects as an `openai/*` glob. So when `OPENAI_API_KEY` is present, `provider_available("openai/does-not-exist")` is `true`, the resolver honors the pin as documented ("an explicit pin if its provider is available"), and `openai/does-not-exist` — not in the catalog — is returned, failing the assertion.

The resolver behavior is correct and intentional. The test's `bogus` case is the problem: it isn't a provider-unavailable pin when an OpenAI key exists, so it never exercises the "unknown pin is ignored" path it claims to. This is not OpenAI-specific — any pin of the form `<configured-provider>/<not-in-catalog>` reproduces it (e.g. `anthropic/does-not-exist` when `ANTHROPIC_API_KEY` is set). OpenAI is just the case the test happened to hardcode.

## Fix

Pin a provider that is never registered (`nonexistent/model`), so the pin is ignored regardless of which API keys are present. Resolution then walks the buddy chain in order and returns the first entry whose provider is available, falling back to the chain's terminal entry (`moonshotai/kimi-k2.6`) when none are. Every chain entry is catalogued, so the result is always a real catalog entry — env-independent, as the test's comment promises. No production code changes; `resolve_model_in` is untouched.

## Verification

`cargo test -p agent-k-backend --lib resolve_always_returns_a_catalogued_model` passes both with and without provider keys in the environment.

### DIFF
--- a/backend/src/model/llm.rs
+++ b/backend/src/model/llm.rs
@@ -365,8 +365,11 @@ mod tests {
         }
         let pinned = resolve_model(Some("coworker"), Some("anthropic/claude-opus-4-8"));
         assert!(catalog_entry(&pinned).is_some());
-        // An unknown/empty pin is ignored, falling through to the chain.
-        let bogus = resolve_model(Some("buddy"), Some("openai/does-not-exist"));
+        // A pin whose provider is not configured is ignored, falling through to
+        // the chain. Use an unregistered provider so the case holds regardless of
+        // which API keys are present — a `<configured-provider>/<unknown>` pin
+        // resolves via the provider glob and would be honored as-is.
+        let bogus = resolve_model(Some("buddy"), Some("nonexistent/model"));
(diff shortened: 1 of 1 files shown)


grf53 (approved):
Thanks

PR #163, opened 2026-06-09
feat: switch to threaded PDF pipeline (docling 2.99)

selloriwoo:
## Summary
Switch the docling PDF converter to docling's **threaded pipeline** (`ThreadedStandardPdfPipeline` + `ThreadedPdfPipelineOptions`), which processes pages through concurrent stages (parse / layout / table) connected by bounded queues. Multi-page PDFs convert faster than the sequential `StandardPdfPipeline`.

The threaded backend requires docling-parse v6 (PR docling-project/docling#3377, landed in docling 2.96), so this also bumps docling.

## Changes
- `python/run_docling.py`: `build_converter` now wires `ThreadedStandardPdfPipeline` via `PdfFormatOption(pipeline_cls=...)` and builds `ThreadedPdfPipelineOptions`. Option fields and defaults are unchanged.
- `python/pyproject.toml`: `docling>=2.91.0` → `>=2.99.0`.
- `python/uv.lock`: docling 2.99.0, docling-parse 6.2.0 (+ dependency restructure into docling-slim; macOS-only ocrmac/pyobjc dropped in favour of rapidocr, matching docling 2.99 defaults).

## Test
- [x] `cargo test -p agent-k --features internal
  translate_pdf_from_financebench -- --ignored` passes (283s): builds
  the PyInstaller bundle with docling-parse v6 and converts a real
  FinanceBench PDF end-to-end through the Rust → bundle path.
- [x] Threaded pipeline confirmed active (`pipeline_cls =
  ThreadedStandardPdfPipeline`).

PR #165, opened 2026-06-09
feat(speedwagon): knowledge-folder corpus, model picker, and sub-agent delegation

nuri-yoo:
## Summary

Reworks Speedwagon from a separate "Documents" surface into a corpus that lives inside the project's Files. A fixed `knowledge` folder *is* the corpus: drop files in, they get indexed, and Speedwagon answers from them with verifiable citations. Coworker and Deep Research can now hand document questions to Speedwagon as a first-class ailoy sub-agent (not a tool wrapper), Speedwagon gets its own recommended model chain (with a soft "(Not Recommended)" hint on the models that fare poorly in its corpus loop, not a hard block), and the corpus tooling is hardened for a long-running multi-project server.

The change touches three layers and is easiest to read by layer:

- **agent-k corpus core** — a `compact()` step on the tantivy store, a `PdfEngine` selector threaded through `ingest`, and a Speedwagon sub-agent spec plus two delegation notes.
- **backend** — a new `handlers::knowledge` module (rescan + status + per-file + citation verification), four dirent hooks that trigger a rescan, a project-level PDF-engine setting with corpus rebuild, and resource hardening in `AppState` (per-project resync serialization, an LRU store cache with bounded derived caches, an indexing Drop-guard, an id cache, a resync-error map).
- **frontend (`app/`)** — the `knowledge` folder rendered as a protected corpus folder with a plain-language banner, per-file Ready/Preparing/Indexing-failed badges, an error banner, a model picker that flags Speedwagon's discouraged models "(Not Recommended)" (without blocking them), and a PDF-engine toggle in project settings. The separate Documents tab, route, API, and page are deleted.

### Corpus = the `knowledge` folder

Each project's `shared/knowledge/` folder is the Speedwagon corpus. Files dropped there are auto-indexed into a per-project tantivy store at `data_root/projects/{project_id}/.speedwagon`; the separate Documents tab, route, API, and page are gone, so file management lives in one place.

The whole pipeline keys off content, not filenames. Store ids are `Uuid::new_v5(&Uuid::NAMESPACE_OID, &bytes)` — a UUIDv5 of the file bytes — so there is no filename-to-id table to keep in sync. A rescan recomputes the desired id set from what is currently on disk and purges anything in the store that is no longer present. Re-uploading the same bytes is idempotent (same id, one document), which the `duplicate_content_indexed_once` test pins.

Sync is a **full rescan on each dirent change** (upload / delete / move / copy), not an incremental diff. It runs once per request on a background `tokio::spawn`, so the dirent response returns immediately. The rescan lives in `handlers::knowledge::resync_inner` and runs four steps in order:

1. **Scan** — recursively walk `{root}/knowledge/`, skipping symlinks (`ft.is_symlink()`), collecting `(bytes, FileType)` for each indexable file. `indexable_filetype(name)` maps the extension: `pdf` to `FileType::PDF`, `md|markdown|txt` to `FileType::MD`, `html|htm` to `FileType::HTML`. Anything else is ignored.
2. **Ingest** — `store.ingest_many(items, pdf_engine).await`. Per-file ingest failures (a corrupt PDF, a parse error) are recorded and skipped, not propagated, so one bad file does not abort the pass. Title/purpose metadata is generated by an LLM but is non-blocking: the helper walks its model chain and falls back to a deterministic value (`"Untitled"` / empty purpose) if every registered model fails, so a provider outage degrades metadata rather than failing the document. The set of ids that ingested successfully becomes the *desired* set.
3. **Purge** — any document in the store whose id is not in the desired set is removed. This is what makes a delete, a move out of the folder, or a whole-subfolder removal converge; `deleting_a_subfolder_purges_its_documents` covers the recursive case.
4. **Compact** — `store.compact()`.

A file that cannot be read mid-rescan is logged and skipped rather than aborting the whole pass, so one unreadable file never blocks the purge of files that genuinely are gone. The rescan records its outcome: on success it clears the project's resync error, on failure it records the error string (see *Indexing status* below).

`maybe_trigger_resync(state, project_id, touched)` is the single entry point — it spawns the background rescan only when `touched` is true, and the four dirent handlers each compute `touched` once and call it once per request.

#### Why `compact()`

agent-k's indexer functions each create a fresh `index.writer(128MB)`, commit, and drop it. That pattern means tantivy's background merge thread essentially never runs, so segments and tombstones from deletes accumulate across rescans. `Store::compact()` (delegating to `indexer::compact`) merges down: it reads `searchable_segment_ids()`, and **only if there are 2 or more** calls `writer.merge(&ids).wait()` followed by `writer.garbage_collect_files().wait()`. The "1 or fewer" guard avoids handing an empty slice to `merge`, which would panic. Without compaction, a project that re-ingests on every upload would grow its on-disk index without bound.

#### Protected folder

The `knowledge` folder auto-creates per project (lazily on the first Shared-scope dirent list, plus best-effort at project creation) and is protected: it cannot be renamed, deleted, moved, or copied, because the indexer's path contract depends on it. `is_knowledge_root(scope, tail)` gates this on the backend — `matches!(scope, DirentScope::Shared { .. }) && tail.trim_matches('/') == KNOWLEDGE_PREFIX` (`KNOWLEDGE_PREFIX = "knowledge"`). Delete returns `400`; move/rename and copy reject the item. The frontend mirrors this by keeping the folder out of selection entirely (single/range/meta click, Ctrl+A, arrow-key move and Shift+arrow range, and rubber-band all skip it) and out of drag, on top of hiding its row menu — so no selection-driven action (bulk delete/move/copy, F2 rename, drag) can target it. Navigating into the folder still works. Four tests pin the guards: `knowledge_folder_cannot_be_deleted`, `knowledge_folder_cannot_be_moved_or_renamed`, `knowledge_folder_cannot_be_copied`, and `upload_to_knowledge_indexes_and_delete_removes` for the happy path.

### PDF parsing engine is a project setting

PDFs are converted to text by one of two engines, chosen per project in Settings. The choice is the `PdfEngine` enum in `knowledge_base::translator`:

- **`Kreuzberg`** (`#[default]`): a native Rust crate with bundled PDFium. No system dependencies, no PyInstaller bundle, fast.
- **`Docling`**: higher-fidelity layout, slower. The merge with `main` picked up the docling 2.99 pipeline (#163); `translate_pdf` calls its async `convert_pdf_file` and awaits it.

`ingest_many` takes the engine as a parameter (`ingest_many(items, pdf_engine)`), and the rescan reads the project's setting and passes it through — the engine is not stored on the `Store`, it is supplied per ingest call. The setting lives in a new `pdf_engine TEXT` column (migration `0011_project_pdf_engine.sql`); the column is nullable and a `NULL`/absent value resolves to `kreuzberg`, so existing projects keep working without a backfill.

Because a corpus id is the hash of the **original** bytes, not of the extracted text, switching the engine would otherwise keep serving text parsed by the old one (same bytes, same id, ingest sees the document already exists, skip). So `update_project` treats an engine change as a corpus rebuild: it detects `engine_changed` by comparing the incoming engine against the stored one, then evicts the cached store (`state.evict_store`), removes the derived corpus directory (`remove_dir_all` of `{data_root}/projects/{project_id}/.speedwagon`, ignoring `NotFound`), and triggers a forced resync that re-parses every PDF with the new engine. The original uploads in `knowledge/` are untouched; only the derived index is rebuilt.

Evicting the store is not enough on its own: a live Speedwagon session caches its agent, and that agent's corpus tools captured the old `SharedStore` at build time, so it would keep querying the now-deleted index. The engine change therefore also drops every cached session agent for the project (`remove_agent` over `list_all_sessions_in_project`); each rebuilds against the freshly reopened store on its next message, while the sessions and their sandboxes stay intact. `changing_pdf_engine_drops_cached_agents` pins this.

### Indexing status, per file and per folder

Two endpoints back the Files UI, both under the project:

- `GET /projects/{project_ref}/knowledge/status` returns `{ indexing: bool, document_count: Option<u32>, error: Option<String> }`. `indexing` reflects the in-flight resync count for the project. The handler reads the corpus count with a **non-blocking** `try_read()` on the store lock, so a large PDF parse holding the write lock leaves `document_count` as `None` rather than blocking the status poll — the UI stays responsive while indexing.
- `GET /projects/{project_ref}/knowledge/files` returns `{ files: [{ path: String, indexed: bool, failed: bool }], indexing: bool }`. Each file's id is derived the same way `ingest` does (UUIDv5 of the bytes) and checked against the store, so per-file status needs no filename-to-id table. `failed` is set for a file the latest resync recorded as failed to index (a corrupt PDF, a parse error), distinguishing a real failure from "still indexing". Ids are resolved during the directory walk without the store lock, and membership is checked under a single non-blocking `try_read()`, so a resync holding the write lock never stalls the poll (matching the status endpoint).

A banner in Files shows when a background resync is in flight, polled against the status endpoint. A failed resync is recorded per project (`set_resync_error` / `resync_error` over a `DashMap<Uuid, String>`) and returned in the same status as `error`; a later success clears it. The banner shows an explicit error state rather than leaving a silently empty or stale corpus — for example after a PDF-engine change whose rebuild failed.

Each indexable file in the knowledge folder shows a **Ready / Preparing / Indexing failed** badge backed by the files endpoint; an unsupported file sits in the folder without a badge rather than showing a "Preparing" state that never resolves. A file that failed to index gets the failed badge (with a hover hint to check or re-upload it, in en/ko) and no longer keeps the poll alive — previously such a file stayed "Preparing" forever. To avoid re-reading and re-hashing every file on every poll, each file's id is cached against its `(mtime, size)`: `cached_file_id(project_id, path, mtime, size)` returns the cached id only when both match, and `cache_file_id(...)` stores `(SystemTime, u64, Uuid)` keyed by `(Uuid, PathBuf)`. An unchanged file is never re-hashed.

### Speedwagon model selection

Speedwagon has its own recommendation chain in `AgentType::chain()`, independent of Coworker. Both share the same `openai/gpt-5.4-mini` then `anthropic/claude-sonnet-4-6` shape, but Speedwagon's Gemini slot is `google/gemini-3.1-flash-lite` (fast, with good corpus-loop accuracy) where Coworker's is `google/gemini-3-flash`.

Model choice is not blocked, only nudged: any catalogued model can be selected, and the chain just drives the ★ recommendation and the default. `AgentType::discourages_model(model_id)` flags models that are merely *not recommended* for an agent — for Speedwagon, `google/gemini-3.5-flash` (slow, times out in the corpus loop). The picker tags it "(Not Recommended)" / "(권장되지 않음)" but keeps it selectable; only a model whose provider has no API key is disabled. `validate_chains` and `ProjectChains::chain_for` no longer reject or filter by agent, and a session's pinned model is honored as-is, so a project may put any catalogued model in any chain.

### Sub-agent delegation

Coworker and Deep Research register Speedwagon as a real ailoy sub-agent via `AgentSpec::subagent`, exposed to the model as a `subagent_speedwagon` tool that takes the full question. `speedwagon_subagent_spec(name, model)` builds the spec with `.with_web_search(false).with_shell(false)` — the sub-agent is corpus-only; its corpus tools (`search_document`, `find_in_document`, `read_document`, `calculate`) resolve against the parent's provider, which is why the parents call `register_corpus_tools(&mut provider.tools, store)` before attaching the sub-agent. The sub-agent carries its own `AgentCard` (`subagent_card()`), so the model has a description to route against.

Delegation no longer rests on the card description alone. When a corpus is attached, each parent also gets an explicit instruction appended to its system prompt (`.instruction(format!("{inst}{NOTE}"))`) telling it to route project-document questions through the sub-agent and to prefer the corpus over web search or its own prior knowledge. The store is read at delegation time, so documents added mid-session are visible and an empty corpus yields "not found". When the corpus store fails to open, the session runs without the sub-agent and logs it (`corpus_store_or_log`) rather than dropping corpus access silently.

The sub-agent runs on the **corpus-recommended model, not the parent's**. Both builders take an optional `corpus_model` (`CoworkerSandboxOptions.corpus_model`, a `get_deep_research_agent` parameter); `None` inherits the parent's model (unchanged default), and the backend sets it to `speedwagon_model_for_parent(&model)` — the same-provider slot from Speedwagon's own chain (`model/llm.rs`). Without this, a session on a model that fares poorly in the corpus loop (e.g. the discouraged `google/gemini-3.5-flash`) dragged the sub-agent down with it; now the parent keeps the user's model while the delegated corpus QA moves to, say, `google/gemini-3.1-flash-lite`. The parent and sub-agent share one provider, so credentials and routing stay consistent.

The two parents handle the sub-agent's citations differently, through two distinct notes:

- **`SPEEDWAGON_DELEGATION_NOTE`** (Coworker) — Coworker has no citation system of its own, so the note tells it to carry the `[^N]` markers and the `## Sources` block through **verbatim**: keep each marker where it sits, do not renumber, paraphrase, or drop a source.
- **`SPEEDWAGON_DELEGATION_NOTE_DEEP_RESEARCH`** (Deep Research) — Deep Research runs its own `artifacts/citations.json` plus `[^N]` numbering, so the note tells it to **fold** a corpus source into that single numbering space as a new entry `{ "title": ..., "source": "corpus" }` with no `url`, and its verify phase exempts a `source: "corpus"` citation from the fetched-URL check while still requiring a URL for web ones.

### Cited answers, verified against the corpus

Speedwagon's system prompt now ends with an explicit answer format: lead with the direct answer, cite every fact with a numbered footnote marker (`[^1]`, `[^2]`, and so on), and close with a `## Sources` section where a corpus fact is `[^1]: <document title>` and a web fact is `[^2]: <page title> — <url>`. The chat already renders markdown with remark-gfm, so the markers show as superscript links plus a references block; the marker and the Sources list are styled to read like a references section.

The citation rules are written to keep markers honest. A footnote is a citation of *support*, so when the answer reports an absence (the corpus or web does not contain the answer, a document does not mention a topic) it is stated with no `[^N]` and no `## Sources` section — a document is not the source of what it fails to say. And markers and definitions must match exactly: every body `[^N]` needs one `[^N]:` line and vice versa, with no dangling marker or uncited definition. Adversarial testing (below) showed these prompt rules sharply reduce, but do not fully guarantee, the absence-citation case; the deterministic semantic check is tracked as a follow-up.

Each cited source is checked against the corpus with no extra LLM call. `verify_citations(text, docs)` (where `docs` is `&[(title, line_count)]`, cached per project and refreshed after each resync so the check never reloads document content) returns one `CitationCheck { index, label, kind, verified, referenced }` per cited source, plus cross-checks markers against the Sources list:

- a **corpus** citation is `verified` when its title matches a real document (case-insensitive, whitespace-collapsed). The prompt no longer asks for a line range, but for backward compatibility `parse_line_range` still validates a `(lines a-b)` / `(line N)` span against `start >= 1 && end <= line_count` when one is present;
- a **web** citation (label contains `http://` / `https://`) is reported as `kind: "web"` and shown as-is;
- `referenced` is `false` for an **orphan definition** (listed in `## Sources` but never cited by a body `[^N]` marker), and a body marker with no matching definition is reported as a `missing` citation — so a dangling `[^N]` or an unused source is flagged rather than passing silently.

The result is rendered as a chip under the answer — plain when verified, otherwise highlighted ("not found in this project's documents", or "no source" for a missing definition). To keep this off the message-fetch hot path, the per-message result is cached in memory keyed by `(project, session, seq)` and recomputed only when the corpus changes (the cache is dropped per project on resync and on engine-rebuild eviction), so a cached result always reflects the current corpus without re-running on every poll.

Example of a direct Speedwagon answer:

```
The office mascot is a capybara named Pibble.[^1]

## Sources
[^1]: Team Handbook
```

### System prompt and shell

Speedwagon's prompt was finalized (the previous one was an interim placeholder) in the Coworker house style, and shell is exposed as a secondary tool. The agent is built via `AgentBuilder`.

The corpus-tool guidance in the prompt is written to match the tools' real signatures. `find_in_document(id, pattern)` and `read_document(id, offset, len)` work in **byte offsets** — a find match reports byte `start`/`end`, and `read_document` reads a byte range — so the prompt describes them that way and the citation format carries no line range (there is no tool that returns line numbers for the model to cite). The `search_document` tool description also points at `read_document`, not the never-registered `open_document` it named before.

### Concurrency and resource hardening

This corpus is per-project and the server is long-running and multi-project, so the shared state was hardened:

- **Indexing Drop-guard.** The in-flight resync count is held in an `IndexingGuard` returned by `begin_indexing(project_id)`; its `Drop` decrements with `saturating_sub(1)`. A panic or early return inside the rescan can no longer leak the count and leave the UI stuck on "indexing".
- **Per-project resync serialization.** A resync builds its desired-id set from an on-disk scan; a per-project lock spans the whole pass (scan included) so two rescans never interleave and the last-queued one reflects the final on-disk state. Without it, a rescan that won the store lock with an older scan could purge a document a newer scan had just indexed, leaving it on disk but unindexed until the next change.
- **Per-project store-init serialization.** `store_for` opened the store outside any lock on a cache miss, so two concurrent callers (a background resync and a status/files poll) could both run `Store::new` on the same directory and hit tantivy's `open_or_create` create/remove TOCTOU — one failing with "No such file or directory". A per-project init lock now spans the open-and-cache step with a re-check under the lock, so only one task opens a given store and the others reuse its handle.
- **LRU store cache.** `document_stores: DashMap<Uuid, StoreEntry>` (with `StoreEntry { store, last_access }`) is capped at `MAX_OPEN_STORES = 64`. `store_for` evicts the least-recently-accessed entry when at capacity, skipping any project that is currently indexing so its store is never reopened as a second handle while a resync holds the tantivy writer lock. `evict_store` (used by the engine-change rebuild and by LRU eviction) drops the store together with the project's derived caches, so those stay bounded with the store cache rather than growing per project.
- **Id cache** keyed by `(project_id, path)` on `(mtime, size)`, so per-file status polls do not re-hash unchanged files; it is dropped with the store on eviction.
- **Cached corpus summary.** The `(title, line_count)` list backing citation verification is cached and refreshed after each resync, so reading a Speedwagon session's history checks citations without loading every document's content.
- **Per-file ingest failures are surfaced**, not silently dropped: the latest resync records which scope-relative paths failed, and the per-file status endpoint returns a `failed` flag so the UI shows an "indexing failed" badge instead of a perpetual "Preparing" spinner.
- **Resync-error map**, so a failed background rescan surfaces in the status endpoint instead of vanishing into a log line.

### agent-k corpus core

The `knowledge_base` suite runs 79 `#[test]`s, including the new compaction behavior: after adds and deletes a `compact()` leaves at most one searchable segment, the guard skips the merge instead of panicking on an empty slice, and search results survive a compaction. Ingest idempotency (same bytes, one document) and purge are covered here too.

### backend — model recommendation (`model/llm.rs`)

- `speedwagon_chain_uses_gemini_3_1_flash_lite` / `chains_are_provider_diverse` — Speedwagon's chain carries `gemini-3.1-flash-lite` and not `gemini-3-flash`, Coworker the reverse, and every chain stays provider-diverse.
- `speedwagon_discourages_gemini_3_5_flash` — `discourages_model` flags `gemini-3.5-flash` for Speedwagon only, and nothing for other agents.
- `validate_chains_accepts_any_catalogued_model` — a discouraged Gemini on Speedwagon is accepted; only an unknown agent_type or uncatalogued model is rejected.
- `chain_for_returns_override_verbatim` — a stored override is returned as-is (nothing filtered out), and the built-in chain is used when there is no override.
- `speedwagon_model_for_parent_maps_within_provider` — a parent model maps to the same-provider Speedwagon slot (`gemini-3.5-flash` → `gemini-3.1-flash-lite`, opus → sonnet), falling back to the parent model when the provider has no slot.

### backend — citation verification (`handlers/knowledge.rs`)

- `verify_citations_matches_corpus_and_flags_unknown` — a citation whose title matches a corpus document verifies; an unknown title does not.
- `verify_citations_rejects_out_of_range_lines` — a line range past the document's line count fails verification even when the title matches.
- `verify_citations_flags_dangling_marker` / `_flags_orphan_definition` — a body `[^N]` with no `## Sources` definition is reported `missing`; a definition no marker cites is `referenced: false`.
- `verify_citations_counts_marker_inside_a_definition_line` — a marker that appears within another definition's line (e.g. `[^1]: … see also [^2]`) still counts as a body reference, so `[^2]` isn't mis-flagged as an orphan.

### backend — knowledge integration (`tests/knowledge_test.rs`, 12 tests)

These drive the real router (upload, background resync, store), mostly over Markdown so they need no PDF engine:

- `upload_to_knowledge_indexes_and_delete_removes` — a file dropped in `knowledge/` becomes searchable, and deleting it removes it from the store.
- `duplicate_content_indexed_once` — the same bytes uploaded twice yield a single document.
- `upload_outside_knowledge_is_not_indexed` — a file outside the folder is not indexed.
- `deleting_a_subfolder_purges_its_documents` — removing a subfolder purges every document under it.
- `knowledge_folder_cannot_be_deleted` / `_cannot_be_moved_or_renamed` / `_cannot_be_copied` — the protection guards.
- `knowledge_files_reports_per_file_indexed_status` — the per-file endpoint reports `indexed` correctly.
- `knowledge_files_reports_failed_status_for_bad_file` — a non-PDF uploaded as `.pdf` is reported `failed: true`, not left pending.
- `concurrent_uploads_all_indexed` — eight files uploaded at once each open the store concurrently and all end up indexed (exercises the per-project store-init serialization).
- `corpus_change_clears_cached_citation_checks` / `deleting_a_session_clears_its_cached_citation_checks` — the in-memory citation-check cache is dropped when the corpus changes (resync) and when a session is torn down, so it never serves a stale result or leaks per deleted session.

`changing_pdf_engine_drops_cached_agents` (in `project_test.rs`) covers the engine-change path: switching the PDF engine drops this project's cached session agents from state so they rebuild against the freshly reopened store.

Two pre-existing, environment-dependent backend failures are unrelated to this PR and reproduce on the base branch (neither touches files this PR changes): a `resolve_*` model test that fails only when provider keys are present (tracked in PR #164), and the `fork_*` session tests, which fail locally because the installed `microsandbox-db` has applied migrations whose files are missing (`create /tmp named volume: ... migration file ... is missing`). The full suite is otherwise green — `knowledge_test` 12/12, `project_test` 14/14, the backend lib suite 62/62, and the agent-k `knowledge_base` suite 79/79.

### Browser and full-session E2E

- **Browser:** the knowledge folder auto-creates, and an upload triggers a background resync after which the file becomes searchable. (The original picker disable / 400-rejection checks no longer apply — model blocking was later removed; the picker now only tags discouraged models "(Not Recommended)" and keeps them selectable.)
- **Full session E2E (real OpenAI, real microsandbox VM):** a Coworker session delegated a corpus-only question to `subagent_speedwagon` and answered from a corpus-only fact, preserving the footnote and `## Sources` block through the parent. A direct Speedwagon session produced the footnote-cited answer shown above. (E2E scripts and fixtures are intentionally not committed.)
- **Adversarial run over the live server (3 models × 10 questions, en + ko):** drove the real HTTP API (signup, upload, background index, Speedwagon session, ask, poll) against `gpt-5.4-mini`, `claude-sonnet-4-6`, and `gemini-3.1-flash-lite`, over an English handbook and a Korean regulations doc. Questions mixed corpus facts, an arithmetic sum, a trap ("does the plan cover vision?", answer: no), Korean-only facts, a cross-document comparison (citing both docs), and unanswerable ones (a fictional NASDAQ ticker, a missing policy). With run-lifecycle tracking so answers align with questions, all 30 answered correctly with zero citation-integrity violations: corpus answers always carried a verified citation, the cross-doc answer carried two, and the new checks fired in practice (a `missing` marker and an `orphan` `referenced: false` definition were both flagged on real model output). The absence-citation tightening above came directly from this run, where one model cited a document for what it didn't contain; after the prompt change, the decline cases drop their spurious citations in 8/9 runs with no regression on normal answers.

### test_case harness coverage (3 models × 14 cases, en + ko)

The `test_case` dev harness now covers Speedwagon directly and as a delegated sub-agent. `Case` gained a `corpus_files` field; the harness builds a `SharedStore` from them (`Store::new` + `ingest_many`) and binds it directly for `speedwagon <n>`, or as a `subagent_speedwagon` sub-agent for `coworker`/`deep-research` cases that ship a corpus. New ko/en corpus fixtures (`maple_library_guide.md`, `haneul_coop_rules.md`) plus reused business docs back 10 direct Speedwagon cases (fact / number / cross-document / business-doc / unanswerable-decline) and 4 delegation cases. Run them with `cargo run -p agent-k --bin test_case -- speedwagon 0 --model claude`, etc.

A full matrix (3 models × all 14 new cases) was run on the latest code. Results:

| Group | Cases | openai/gpt-5.5 | claude-opus-4-7 | gemini-3.5-flash |
|---|---|---|---|---|
| Direct — fact/number/cross-doc/biz-doc | 8 (×3) | all answer + cite | all answer + cite | all answer + cite¹ |
| Direct — unanswerable (decline) | 2 (×3) | decline, no citation | decline, no citation | decline, no citation |
| Delegation — Coworker (flyer.html / 안내문.pdf) | 2 (×3) | delegates + builds file | delegates + builds file | delegates + builds file |
| Delegation — Deep Research (report.md + citations.json) | 2 (×3) | delegates + folds citation | delegates + folds citation | delegates + folds citation |

All 30 direct runs indexed cleanly (0 failures), every fact answer carried a verified citation (cross-doc carried two), and all 6 decline runs answered with no citation. All 12 delegation runs invoked `subagent_speedwagon` and produced their artifact (HTML flyer, PDF notice, or research `report.md` + folded `citations.json`).

¹ Direct Speedwagon runs on the corpus-recommended model for the chosen provider — `speedwagon_model_for(provider)`, the same mapping delegation uses — rather than the harness's heavier shared default. So `--model gemini` runs Speedwagon on `gemini-3.1-flash-lite`, not the discouraged `gemini-3.5-flash`. An earlier run that fed the heavy default to Speedwagon directly hit a 200s timeout on the gemini cross-document case; with the recommended model that case finishes in ~16s. (Coworker/DeepResearch keep their own heavier defaults; only Speedwagon's own corpus loop moves to the light model.)

Delegation artifacts were inspected, not just logged: the Coworker `flyer.html` carries the corpus's hours / borrowing limit / late fee, and the Deep Research `report.md` weaves the corpus figures (412,000 items, 1.02M circulation) into a web-researched briefing with a `"source": "corpus"` entry folded into `citations.json`.

## Dependency (ailoy pin)

This branch pins ailoy at `74f4e23`. It was originally stacked on two unmerged ailoy PRs — #427 (`shell_tool()` builder) and #426 (Gemini missing-role fix) — pinned via `f6e533d`. Both have since merged into ailoy `develop`, so the merge with `main` moved the pin forward to `74f4e23` (develop with #426/#427 plus #428–#430), and the sub-agent mechanism (`AgentSpec::subagent`, the `subagent_<name>` tool, `AgentCard`) it relies on is on `develop` as well. No unmerged ailoy work is left.

## Notes and follow-ups

- Citation verification confirms the cited document exists and that markers and Sources line up; it is not yet a content check (whether the cited text actually backs the claim). Two follow-ups are scoped but intentionally left out of this PR, since each adds a new responsibility the backend does not have today:
  - **Semantic check** — confirm the cited passage actually supports the marker's claim, with a cheap model. Needs an LLM-call path in the backend (it only reads the model catalog today) and a way to point at the cited span (the prompt no longer carries one), so it is a design of its own.
  - **Web reachability** — confirm a web citation's URL resolves. The backend has no HTTP client and citation URLs are model-generated, so this needs SSRF defense (internal-address blocking, timeouts) before it is safe to fetch them. The citation cache above is the hook both would hang off so neither runs per poll.
- A converted file's original bytes exist twice on disk: once as the user's upload in `shared/knowledge/`, once as the store's own copy in `.speedwagon/origin/{uuid}.ext` (markdown skips this; only PDF/HTML are copied). The `origin/` copy is deliberate — it keeps the store self-contained, so an engine change can re-convert from it and a purge owns the file it deletes, without reaching back into the dirent tree. Replacing the copy with a hard link would reclaim the bytes, but the store no longer takes a path (resync reads files into memory and passes bytes), so it would need an API change to thread the source path through, and the linked inode would then share the user file's fate — losing the self-containment that matters most for long-running, multi-environment servers (a moved or remounted `knowledge/` original could strand the origin). The right first step is to measure: compare `origin/` size against `shared/knowledge/` per project to see whether the duplication is a real disk cost before trading that durability for it.
- The two arrow-key selection paths skipped the protected-folder guard the other paths enforce, so `knowledge` could land in the selection via plain Arrow (single-select) or Shift+Arrow (range), and the bulk bar then offered Delete/Move/Copy on it (the server rejects each, but the UI shouldn't). `84d104c` filters the range and guards the single-select with `isProtectedPath`, matching click / Ctrl+A / marquee. Confirmed end-to-end: arrowing onto `knowledge` keeps it out of the selection, and a Shift+Arrow range across it selects the plain folders only.
- Two scroll fixes for the footnoted-answer layout reported in #183. A footnote's `## Sources` section makes the answer tall, and the message-list auto-scroll used `scrollIntoView` on an end sentinel, which walks every scrollable ancestor up to the document root and scrolled the page itself, leaving the list parked above the viewport with only the composer visible (a window resize forced a relayout that masked it). `5c647a8` scrolls the container directly and drops the sentinel, fixing run-completion and session-entry. The footnote click handler had the same `scrollIntoView` problem, so clicking a citation in a scrollable conversation pushed the page too; `5b91272` centers the target by scrolling the message-list container by hand. Verified on a real Speedwagon session (`openai/gpt-5.4-mini`, a corpus doc cited with real `[^n]` footnotes): `document.documentElement.scrollTop` stays at 0 on entry, run completion, and marker/back-ref clicks, versus ~180–670px before.
- The branch was merged up to `main` twice to resolve `files.tsx` conflicts. First with the in-app file-preview work (#161), which touched the same `RowMenu`; both the preview action and the protected-folder menu hiding are kept. Then with the drag-files-into-a-chat work (#166), which extracted the rubber-band selection into a `useMarqueeSelection` hook plus `MarqueeOverlay`. That hook is adopted, and the protected-folder rule this branch added (the `knowledge` folder is excluded from selection, rubber-band included) is preserved by adding an optional `canSelect` gate to the hook — so `SharedFilesPanel`, the other caller, is unchanged. A marquee over the file list selects the plain folders and skips `knowledge`.

## Screenshots

<details>
<summary>Knowledge folder = corpus</summary>

**The knowledge folder, with the AI-reads-these banner (EN)**
01-knowledge-banner-en

**Same banner (KO)**
02-knowledge-banner-ko

**Per-file status — Ready badges once indexed**
04-per-file-ready-badges

**Per-file status — Preparing while a resync is in flight (folder banner shows "indexing")**
03-indexing-banner

**Preparing badge on a single file**
05-per-file-preparing-badge

**Sidebar — no separate Documents tab; files live under Files**
06-sidebar-no-documents-tab

</details>

<details>
<summary>PDF parsing engine (project setting)</summary>

**Settings — Kreuzberg / Docling toggle (EN)**
07-pdf-engine-toggle-en

**Same toggle (KO)**
08-pdf-engine-toggle-ko

</details>

<details>
<summary>Speedwagon model selection</summary>

**Coworker picker — all Gemini models available (control)**
10-picker-coworker-en

**Speedwagon picker — `gemini-3.5-flash` marked "(권장되지 않음)" but still selectable (KO)**
11-picker-speedwagon-ko

**Same hint in EN ("(Not Recommended)")**
12-picker-speedwagon-en

**Per-agent recommendation chains in Settings (Coworker: Gemini 3 Flash, Speedwagon: Gemini 3.1 Flash-Lite)**
13-model-chains-ko

<details>
<summary>Cited answers and verification</summary>

**Speedwagon answer — inline footnote marker plus a Sources section**
13-speedwagon-footnote-answer

**Verified citation chip (matches a corpus document)**
14-citation-chip-verified

**Unverified citation chip (not found in this project's documents)**
15-citation-chip-unverified

</details>

<details>
<summary>Sub-agent delegation</summary>

**Coworker delegates to `speedwagon` and reproduces its answer**
16-coworker-delegation

**The footnote and Sources block are preserved through the parent**
17-coworker-source-preserved

</details>

<details>
<summary>Protection and edge states</summary>

**The knowledge folder offers no rename/move/copy/delete actions**
18-knowledge-folder-no-menu

**A failed resync surfaces as an error banner**
19-resync-error-banner

**Empty corpus — banner only**
20-empty-corpus

</details>


backend/src/handlers/knowledge.rs
```diff
@@ -0,0 +1,587 @@
...
+    match resync_inner(&state, project_id).await {
+        Ok(()) => state.set_resync_error(project_id, None),
+        Err(e) => {
+            tracing::warn!(%project_id, "knowledge resync failed: {e}");
+            state.set_resync_error(project_id, Some(e));
+        }
+    }
+}
+
+async fn resync_inner(state: &Arc<AppState>, project_id: Uuid) -> Result<(), String> {
```
grf53, line 84: In the project, cache entries for file paths among knowledge_file_ids that are no longer on disk are cleaned up only in evict_store; they are not cleaned up unless they are evicted.

`resync_inner` has been suggested as the appropriate location to complement this by executing the action of keeping only 'indexable file IDs'.

backend/src/state.rs
```diff
@@ -56,14 +103,221 @@ impl AppState {
...
+        if self.document_stores.len() >= MAX_OPEN_STORES
+            && let Some(oldest) = self
+                .document_stores
+                .iter()
+                .filter(|e| !self.is_indexing(*e.key()))
+                .min_by_key(|e| e.last_access)
+                .map(|e| *e.key())
+        {
+            self.evict_store(oldest);
+        }
```
grf53: Although it seems unlikely that a situation will occur where it goes into 'else', this if statement is logically somewhat incomplete. Since the possibility of failing to find an item that is not currently indexed while satisfying the MAX condition is not covered by the code(and no logs for that situation), it would be good if this part could be improved (even if simply).

backend/src/handlers/knowledge.rs
nuri-yoo: applied in [ba1ad49](https://github.com/brekkylab/agent-k/pull/165/commits/ba1ad49ed3690c0c57ddda21ae6b5adfd20aecca)

backend/src/state.rs
nuri-yoo: applied in [f922b34](https://github.com/brekkylab/agent-k/pull/165/commits/f922b34f1af4172ffab964123ea5633cd5880566)

backend/src/handlers/project.rs
```diff
@@ -258,13 +284,41 @@ pub async fn update_project(
...
+            .data_root
+            .join("projects")
+            .join(project_id.to_string())
+            .join(".speedwagon");
+        if let Err(e) = tokio::fs::remove_dir_all(&speedwagon_dir).await
+            && e.kind() != std::io::ErrorKind::NotFound
+        {
+            tracing::warn!(%project_id, "failed to clear corpus on engine change: {e}");
+        }
+        super::knowledge::maybe_trigger_resync(&state, project_id, true);
```
khj809, line 329: The race condition scenario:
If the engine is changed during the resync is in progress, the `.speedwagon` directory is deleted immediately and the second resync is reserved via `maybe_trigger_resync`. The first resync fails since it's trying to write on already absent `.speedwagon` (ENOENT) and the error message is displayed. After then, the second resync generates `.speedwagon` again and re-index using the changed engine, which makes the error message disappears.

Suggested fix:
Just try to get `resync_lock` before removing `.speedwagon`, so it can be safely removed after the in-flight resync is finished.

```suggestion
        let speedwagon_dir = state
            .data_root
            .join("projects")
            .join(project_id.to_string())
            .join(".speedwagon");
        // Acquire `resync_lock` to wait until the in-flight resync is finished
        let resync_lock = state.resync_lock_for(project_id);
        let _guard = resync_lock.lock().await;
        if let Err(e) = tokio::fs::remove_dir_all(&speedwagon_dir).await
            && e.kind() != std::io::ErrorKind::NotFound
        {
            tracing::warn!(%project_id, "failed to clear corpus on engine change: {e}");
        }
        // Release `_guard` and start a new resync
        drop(_guard);
        super::knowledge::maybe_trigger_resync(&state, project_id, true);
    }
```

nuri-yoo: Thanks for catching this. Confirmed and applied at [56e17b9](https://github.com/brekkylab/agent-k/pull/165/commits/56e17b9ddedd9de26eed72f2ab8c00bb9d575073)

grf53 (approved):
Thank you for your hard work!

backend/src/handlers/knowledge.rs
```diff
@@ -0,0 +1,594 @@
...
+    let store = state
+        .store_for(project_id)
+        .await
+        .map_err(|(status, _)| format!("store_for failed: {status}"))?;
+    let mut store = store.write().await;
+
+    let result = store
+        .ingest_many(items, pdf_engine)
+        .await
+        .map_err(|e| e.to_string())?;
```
khj809, line 148: When `ingest_many` fails (returns Err) it returns immediately here, which skips `state.set_failed_files()` and may result in showing the previous failed files in UI.

Suggested fix: clear the previous failed files before running `ingest_many`, so we can track the failed files exactly.

```suggestion
    // Clear the failed files of previous resync before running ingest
    state.set_failed_files(project_id, HashSet::new());

    let result = store
        .ingest_many(items, pdf_engine)
        .await
        .map_err(|e| e.to_string())?;
```

khj809 (approved):
Thanks for the huge work!
Left a minor comment below.

selloriwoo (approved):

backend/src/handlers/project.rs
```diff
@@ -258,13 +284,46 @@ pub async fn update_project(
...
+        let _guard = resync_lock.lock().await;
+        if let Err(e) = tokio::fs::remove_dir_all(&speedwagon_dir).await
+            && e.kind() != std::io::ErrorKind::NotFound
+        {
+            tracing::warn!(%project_id, "failed to clear corpus on engine change: {e}");
+        }
+        // Release `_guard` and start a new resync
+        drop(_guard);
+        super::knowledge::maybe_trigger_resync(&state, project_id, true);
+    }
```
ljhh-0611, line 330: **Engine-change rebuild evicts the store before the resync lock → reopen-on-deleted-dir race.**

  Between `evict_store` and the lock, the cache is empty but `.speedwagon` still exists. A concurrent `store_for` (status poll / citation recompute / agent build) cache-misses, reopens the **old** index, caches it — then `remove_dir_all` deletes it underneath, and the resync gets a stale handle on a deleted dir. `resync_lock` doesn't help since `store_for` doesn't take it. Fix: delete first, evict last, all under the lock, so a post-eviction reopen creates a fresh empty index.

  **Low-probability and self-healing via the error banner**, but the order would better to be fixed.

```suggestion
    if engine_changed {
        let speedwagon_dir = state
            .data_root
            .join("projects")
            .join(project_id.to_string())
            .join(".speedwagon");
        // Hold the resync lock across the rebuild. Delete the derived index
        // FIRST, then evict + drop agents, so a store reopened after eviction
        // is created fresh, not caching the about-to-be-deleted old index.
        let resync_lock = state.resync_lock_for(project_id);
        let _guard = resync_lock.lock().await;
        if let Err(e) = tokio::fs::remove_dir_all(&speedwagon_dir).await
            && e.kind() != std::io::ErrorKind::NotFound
        {
            tracing::warn!(%project_id, "failed to clear corpus on engine change: {e}");
        }
        state.evict_store(project_id);
        if let Ok(sessions) = state.repository.list_all_sessions_in_project(project_id).await {
            for session in sessions {
                state.remove_agent(&session.id);
            }
        }
        drop(_guard);
        super::knowledge::maybe_trigger_resync(&state, project_id, true);
    }
  ```

File: app/src/routes/_app.projects.$projectSlug.files.tsx
Comment (ljhh-0611) [inline app/src/routes/_app.projects.$projectSlug.files.tsx:544]: **Arrow-key selection skips the `isProtectedPath` guard → the `knowledge` folder becomes selectable.**

Click (`handleSelect`), Ctrl+A, marquee, and drag all exclude the protected folder, but the two arrow-key paths don't: `Shift+Arrow` range selection and plain `Arrow` single-select can put `knowledge` into `selectedPaths`.

The bulk bar then shows it as a Delete/Move/Copy target. Server-side guards reject every op so it's cosmetic (no data loss), but it breaks the "knowledge is never selectable" invariant the other paths uphold and surfaces a confusing partial-failure toast.

```suggestion
        if (e.shiftKey && anchorRef.current != null) {
          const aIdx = rowIndex.get(anchorRef.current);
          if (aIdx != null) {
            const [lo, hi] = aIdx <= next ? [aIdx, next] : [next, aIdx];
            setSelectedPaths(new Set(allRows.slice(lo, hi + 1).map((r) => r.path).filter((p) => !isProtectedPath(p))));
          }
        } else if (!isProtectedPath(nextEntry.path)) {
          setSelectedPaths(new Set([nextEntry.path]));
          anchorRef.current = nextEntry.path;
        }
```

ljhh-0611: I couldn't sure that this Suggestion Action will be right-well functioned in this comment. @nuri-yoo, you has to be double-check if the change is done well after you submit the suggestion

backend/src/handlers/project.rs
nuri-yoo: Addressed in [f8e2c9b](https://github.com/brekkylab/agent-k/pull/165/commits/f8e2c9bffb0e4a6980c4c1b5160355fcb87c8fe4)

app/src/routes/_app.projects.$projectSlug.files.tsx
nuri-yoo: Confirmed and addressed in [84d104c](https://github.com/brekkylab/agent-k/pull/165/commits/84d104c7547166893862f59314f1ed8ad56e02c2)

ljhh-0611: Thanks!

ljhh-0611 (approved):
🫡

PR #166, opened 2026-06-09
feat: import project files into a chat session via drag-and-drop

selloriwoo:
## Overview
Adds an end-to-end flow for pulling project files into a chat session as attachments.
A shared-file browser lives in the session's right sidebar, and files/folders can be dragged from there (or from the Files page) to attach them to the next message.

## Features
1. **Sidebar shared-file browser** — browse the project's shared files in the right sidebar with folder navigation (breadcrumb, back, home + project name).
   A bottom segmented toggle switches between Info ↔ Files; the active tab is accent-highlighted.
2. **Drag-and-drop to attach**
   - Sidebar browser → chat: adds to the next message's attachments (no upload — references the existing shared path).
   - Files page → a session row in the left nav: navigates to that session and attaches.
   - Dropping a folder expands it into the files it contains, recursively (skips `.keep`/dotfiles).
3. **Marquee multi-select + bulk drag** — rubber-band select files and folders from empty space, then drag the whole selection at once.
Selection keeps up while scrolling mid-drag; the box is clamped to the list bounds.
5. **Attachment cap** — 30 on the frontend (overflow dropped with a toast) plus a server-side hard ceiling of 30 in `validate_attachments` (with a unit test).

## Internal / refactor
- Extracted the marquee selection logic into a shared `useMarqueeSelection` hook (used by both the Files page and the sidebar browser).
- Dedupe/cap attachments inside the state updater so it's idempotent (prevents StrictMode double-attach).

## UX polish
- Sidebar rows: color-chip icons + drag handle; long names ellipsize.
- Uniform accent drop-zone overlay on the chat while dragging (replaces the ragged text selection).
- Fixed file-row selection fill (cleaned up conflicting legacy CSS).

## Verification
- [x] Frontend `tsc -b` passes
- [x] Backend `cargo check` passes
- [x] Backend Attachment-cap unit test passes
- [x] Core drag-and-drop flows exercised manually in local dev

## Notes
- `.keep` is a placeholder the backend writes when an empty folder is created, so it's excluded from listings/attachments (consistent with the Files page).


app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -828,30 +924,69 @@ function SessionPage() {
...
 
       <aside className="cw-session-side">
-        <h3>{t('side.members')}</h3>
-        {userList.map((user) => (
-          <div className="cw-side-row" key={user.id}>
-            <Avatar user={user} small />
-            {user.name}
+        <div className={`cw-side-views${sideView === 'files' ? ' show-files' : ''}`}>
+          {/* ── default info view (left) ──────────────────────────── */}
+          <div className="cw-side-view">
```
grf53:
```suggestion
          <div className="cw-side-view" inert={sideView !== 'info'}>
```
This would disable interactions with the view when it does not come up front.
interactions are hard to occur cuz the view would be hidden, but just better to be safer.

app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -828,30 +924,69 @@ function SessionPage() {
...
-          {t('side.project_label', { name: project.data?.name ?? '...' })}
-        </p>
-        <ArtifactsPanel
-          projectId={projectId}
-          sessionId={sessionId}
-          onCopyToShared={(scope, paths) => setCopyToShared({ scope, paths })}
-        />
+
+          {/* ── files browser view (right) ────────────────────────── */}
+          <div className="cw-side-view">
```
grf53:
```suggestion
          <div className="cw-side-view" inert={sideView !== 'files'}>
```

Here too.

selloriwoo: Good catch, applied to both views (info ↔ files). 👍

app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -423,6 +430,80 @@ function SessionPage() {
...
+    if (pendingAttachments.length + fresh.length > MAX_ATTACHMENTS) {
+      showToast(t('shared_files.attach_limit', { max: MAX_ATTACHMENTS }));
+      return; // over the cap — don't attach any of them
+    }
+    // Dedupe against the authoritative `prev` inside the updater, so a repeated
+    // or concurrent call (e.g. StrictMode's double-invoked effect) can't append
+    // the same files twice.
+    setPendingAttachments((prev) => {
+      const existing = new Set(prev.map((a) => a.globalPath).filter(Boolean));
+      const toAdd = fresh.filter((it) => !existing.has(it.globalPath));
```
ljhh-0611: The cap check moved out of the `setPendingAttachments` updater, so it now reads the stale closure `pendingAttachments` instead of the authoritative `prev`. Two imports that race before a re-render — most realistically the async `attachShared` effect path, which calls `importSharedFiles` after `await fetchQuery` — both read the same stale length, both pass the pre-check, and the updater only dedupes. Distinct-file batches can then push the total past `MAX_ATTACHMENTS`. Over the cap, send hits the backend 400, and since `send()` clears the composer/attachments before `await`, the draft is lost.

  The pre-check above stays as the UX fast-path (early return + toast). Restore the hard clamp inside the updater so the invariant holds under the race — it's a no-op on the normal path (already `<= MAX`):

```suggestion
      const toAdd = fresh
        .filter((it) => !existing.has(it.globalPath))
        .slice(0, Math.max(0, MAX_ATTACHMENTS - prev.length));
```

ljhh-0611 (commented):
**Out of this PR's strict scope**, but the new backend attachment cap turns this pre-existing path into data loss:

  - `handleFileSelect` appends every picked file with no client-side cap (unlike the shared-import path, which caps at MAX_ATTACHMENTS).
  - `send()` clears `composerText` / `pendingAttachments` before `await sendMessage(...)` and the `catch` doesn't restore them — so any send failure (network, 403, 423, and now the 400 from >30 attachments) **drops the user's typed text and attachment tray.**

  Pick one (or both):
  - **Light (in this PR)**: cap `handleFileSelect` like the shared-import path, so the UI can't reach the backend 400.
  - **Root (can be a follow-up)**: capture `pendingAttachments` before clearing and restore it (with `setComposerText(text)`) in the `catch`, so drafts survive any failed send.

  **Not blocking** — flagging because the new cap is what makes the >30 case reachable.

grf53 (approved):
For me, it looks good to go.

ljhh-0611 (approved):
LGTM too

PR #167, opened 2026-06-10
Session title improvement

grf53:
## Problem
New sessions only showed their auto-generated title after the first turn finished. The backend generates the title concurrently and broadcasts `session_title_updated`, but the WS handler patched the `['session', <uuid>]` query cache while the session page queries `['session', <prefix>]` (since #115's short-prefix URLs). The live update never hit the page — the title only appeared when `agent_run_done` invalidated the prefix-keyed query.

## Fix
In the `session_title_updated` handler, derive the prefix from the event's UUID with `shortSessionId(...)` and `setQueryData` the prefix-keyed query directly. Frontend-only, no refetch, no backend change; the prefix rule stays in one place (`shortSessionId`).

### model chain for session title agent
The AI ​​model of the existing session title agent was fixed to `gpt-5.4-nano`; considering the lack of a provider for this, we applied a chain similar to the Model Selection function. The applied chain is hard-coded and consists of the (almost) cheapest non-deprecated components (excluding kimi).

## Test
- Start a new session → title appears as soon as it's generated, before the first response completes.
- `tsc` passes; backend unchanged.

### DIFF
--- a/app/src/routes/_app.tsx
+++ b/app/src/routes/_app.tsx
@@ -9,6 +9,7 @@ import { useLayoutStore } from '@/stores/layout';
 import { Sidebar } from '@/components/layout/Sidebar';
 import { appWs } from '@/api/ws';
 import type { Session } from '@/domain/types';
+import { shortSessionId } from '@/lib/sessionId';
 
 export const Route = createFileRoute('/_app')({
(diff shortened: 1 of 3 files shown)


khj809 (approved):
Works well. Thanks!

PR #178, opened 2026-06-11
Copy replies and tool calls

grf53:
## Ticket
resolves #177 

## Description
Adding a 'Copy to Clipboard' feature for each Agent response message.
And also, a separate 'Copy to Clipboard' function was added to the tool call results.

c.f. Since the 'Regenerate' feature requires backend modifications and design decisions, it is separated into a separate issue (#176) along with the 'User Message Edit' feature.

## Changes
- AI reply copy: Add expandable Copy button that copies the message body
- Tool-call details: a Copy button (shown when details expanded) that serializes
  the call's name/inputs/results to markdown, building on the structured
  parse from the previous commit
- Markdown uses dynamic backtick fences so values containing backticks
  don't break inline code / code blocks

## Related PR
- #179: While resolving this issue, I also changed the tool call rendering and separated it into a separate PR. In terms of commit and merge order, 179 comes first.

### Writing a python script file
<details>
  <summary>Show</summary>

#### Message

image

#### Copied message text
```
간단한 파이썬 스크립트를 하나 만들어드리겠습니다.
```

#### Tool call
<details>
  <summary>Show</summary>

image

</details>

#### Copied tool call text
<details>
  <summary>Show</summary>

````
## 🔧 write

### Inputs

- **path**: `/workspace/artifacts/hello_world.py`
- **content**:

```
#!/usr/bin/env python3
"""
간단한 파이썬 스크립트 예제
"""

def greet(name):
    """사용자의 이름을 받아 인사합니다."""
    return f"안녕하세요, {name}님! 반갑습니다."

def calculate_sum(numbers):
    """숫자 리스트의 합을 계산합니다."""
    return sum(numbers)

def main():
    """메인 함수"""
    print("=" * 40)
    print("파이썬 스크립트 예제")
    print("=" * 40)
    
    # 인사 함수 실행
    message = greet("김철수")
    print(f"\n{message}")
    
    # 합계 계산
    numbers = [10, 20, 30, 40, 50]
    total = calculate_sum(numbers)
    print(f"\n숫자 리스트: {numbers}")
    print(f"합계: {total}")
    
    # 간단한 반복문
    print("\n구구단 (3단):")
    for i in range(1, 10):
        print(f"3 × {i} = {3 * i}")
    
    print("\n" + "=" * 40)

if __name__ == "__main__":
    main()

```

### Results

- **ok**: true
- **bytes_written**: 891
````
</details>

#### Rendered copied tool call text
<details>
  <summary>Show</summary>

### Inputs

[...]

PR #179, opened 2026-06-11
Render tool-call inputs/results as structured rows

grf53:
## Description
Improve how tool-call inputs/results are displayed, replacing the raw JSON dump with a structured view.

## Changes
  - raw JSON `<pre>` dump → structured key/value HTML rows. Values are rendered
    directly via React (never through a markdown parser), so arbitrary tool data
    can't be misinterpreted (no escaping/XSS issues).
  - Per-type rendering: nested objects/arrays are pretty-printed, single-line
    strings render inline, multi-line strings drop to a `<pre>` block.
  - Width: collapsed wraps its content as a compact chip; expanded fills the
    bubble's full width so open tool calls line up consistently.
  - Lazy rendering: inputs/results (and their JSON parsing) render only while
    expanded, so collapsed tool calls cost nothing. Parsing is memoized per value.
  - Extracted into `components/chat/ToolCallDetails.tsx`; pure formatting logic
    lives in `lib/toolCallFormat.ts` with unit tests.

- Parsing/classification extracted to `lib/toolCallFormat.ts`, memoized once per value.

## Tests
- frontend `tsc` + added unit tests(`app/src/lib/toolCallFormat.test.ts`) pass.

## Screenshots
See the tool call renderings in 'Examples' part of #178

### DIFF
--- a/app/src/components/chat/ToolCallDetails.tsx
+++ b/app/src/components/chat/ToolCallDetails.tsx
@@ -1,0 +1,68 @@
+import { useMemo, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { parseToolCallValue, classifyFieldValue } from '@/lib/toolCallFormat';
+import type { ToolCallInvocation } from '@/domain/types';
+
+/// Renders a parsed tool-call value as React rows — values are shown verbatim (no
+/// markdown parsing), mirroring the structure `valueToMarkdown` uses for copy.
+function ToolCallValueView({ value }: { value: unknown }) {
+  // Memoized so the JSON.parse (in parseToolCallValue) and per-field JSON.stringify
+  // (in classifyFieldValue) run once per value, not on every re-render. String values
+  // compare by value, so an unchanged result string is a cache hit.
+  const view = useMemo(() => {
+    const parsed = parseToolCallValue(value);
+    if (parsed.kind !== 'fields') return parsed;
+    return {
+      kind: 'fields' as const,
+      fields: parsed.fields.map((f) => ({ key: f.key, display: classifyFieldValue(f.value) })),
+    };
+  }, [value]);
+
+  if (view.kind === 'empty') return <p className="cw-toolcall-empty">(empty)</p>;
+  if (view.kind === 'raw') return <pre className="cw-toolcall-raw">{view.text}</pre>;
+  return (
+    <dl className="cw-toolcall-fields">
+      {view.fields.map(({ key, display }) => (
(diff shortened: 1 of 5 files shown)


app/src/lib/toolCallFormat.ts
```diff
@@ -0,0 +1,59 @@
...
+///  - 'code':   a single-line string shown as inline code (verbatim, no quotes)
+///  - 'block':  a multi-line string, or a nested object/array, shown as a block.
+///              Objects/arrays are pretty-printed so nesting stays readable.
+export type FieldDisplay =
+  | { kind: 'inline'; text: string }
+  | { kind: 'code'; text: string }
+  | { kind: 'block'; text: string };
+
+export function classifyFieldValue(value: unknown): FieldDisplay {
+  if (typeof value === 'string') {
```
nuri-yoo, line 52: Tool outputs can legitimately contain empty arrays/objects, null fields, or empty strings. For example, no-match search/find results may include [] or null fields, and existing stream tests already use {} as tool-call arguments.

With the new row rendering, empty values can become visually ambiguous: a value like "" may look like just `key:` with no visible value, making it hard for users to tell whether the value is intentionally empty, missing, or not rendered. The old raw JSON view preserved that distinction.

Suggested fix: render ambiguous empty values explicitly, e.g. `""`, `null`, `{}`, and `[]`, and add formatter tests for those representations.

grf53: Makes a lot of sense, so I applied your suggestion at 10ccd06. Thanks!

nuri-yoo (approved):
Tested the latest commit. LGTM

app/src/components/chat/ToolCallDetails.tsx
```diff
@@ -0,0 +1,68 @@
...
+              <span className="cw-toolcall-section-label">Inputs</span>
+              <ToolCallValueView value={tc.arguments} />
+            </div>
+          )}
+          {tc.result !== undefined && (
+            <div className="cw-toolcall-section">
+              <span className="cw-toolcall-section-label">Results</span>
+              <ToolCallValueView value={tc.result} />
+            </div>
+          )}
```
ljhh-0611, line 63: Are `Inputs` and `Results` labels Intended not to make keys for international localized Label?

ljhh-0611 (approved):
If checked above comment, then LGTM

app/src/components/chat/ToolCallDetails.tsx
grf53: For now, yes. Please refere to https://github.com/brekkylab/agent-k/pull/178#issuecomment-4687381643.
And leave your opinions if you have some!

PR #180, opened 2026-06-12
fix(session): fork sandbox only when source session has one

khj809:
## Summary

Fixes #152.

Previously, `fork_session` unconditionally called `Sandbox::new` on the source session with `persist: true`, which causes ailoy to **create** a new VM when one doesn't already exist. This meant:

- Forking a `DeepResearch`, `Buddy`, or `Speedwagon` session (which never uses a sandbox) would silently materialize an empty VM and then fork it.
- Forking a `Coworker` session that was created but never run had the same problem.

## Changes

- **`backend/src/handlers/session.rs`**: Added an early-exit guard in `fork_session` using the new `Sandbox::exists()` API. If the source session has no sandbox, the fork still proceeds (DB row and host files are copied as before) but the VM fork is skipped entirely. The forked session will create its own sandbox on first run if needed.
- **`Cargo.toml`**: Bumped ailoy rev to `d4acfcee` which adds `Sandbox::exists(name) -> bool` — a lightweight, read-only probe that returns `true` only if a persisted VM with that name already exists.

## Test plan

- [x] Fork a `DeepResearch` or `Speedwagon` session → fork returns 201, no sandbox VM is created (check `msb` CLI or log line `session forked without sandbox (source has none)`)
- [x] Fork a `Coworker` session that has never been run → same as above
- [x] Fork a `Coworker` session that has been run at least once (sandbox exists) → fork returns 201, forked session's sandbox contains the source snapshot (regression check)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -150,7 +150,7 @@ dependencies = [
 [[package]]
 name = "ailoy"
 version = "0.3.0"
-source = "git+https://github.com/brekkylab/ailoy.git?rev=d4acfcee6fc626334632f0b7f8c80e6ffa2a2324#d4acfcee6fc626334632f0b7f8c80e6ffa2a2324"
+source = "git+https://github.com/brekkylab/ailoy.git?rev=74f4e238070d2cec752285394b527a86c023a082#74f4e238070d2cec752285394b527a86c023a082"
 dependencies = [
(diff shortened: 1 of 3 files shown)


nuri-yoo (approved):
LGTM

PR #181, opened 2026-06-12
feat: file upload on Home (clip + drag) and drag-drop upload on Session

selloriwoo:
> **Stacked PR — base is `feat/file-import-session`, not `main`.**
> Depends on #<base-PR>; merge that first. After it lands on `main`, this PR
> retargets to `main` automatically (if the base is squash-merged, I'll
> `rebase --onto main` to drop the absorbed commits).

## What & why
Computer-file upload was only available on the Files page. This adds it to the
two places people actually start work, reusing the session's attachment model.

- **Home composer** — clip-button picker **and** drag-and-drop. The whole page
  is the drop zone (mirrors the session chat's full-surface highlight). Home has
  no session yet, so staged files are uploaded to the new session's `inputs/`
  on send, then attached to the auto-sent first message.
- **Session chat** — files dropped on the conversation now upload (previously
  only the 📎 button did). Shared-file drags still attach as before.
- Extracted a reusable `useFileDropzone` hook that reacts only to native file
  drags, so it coexists with other drag payloads on the same element.

## Robustness
- AThe home composer caps attachments at MAX_ATTACHMENTS on both clip and page-drag (whole batch rejected + toast, matching the backend limit); on the session this also extends the existing clip cap to dropped files.
- Upload failures are surfaced, never silently dropped:
  - **partial** failure → the message still sends with the files that uploaded;
    each failed file is listed with its reason.
  - **total** failure (network / 4xx / 5xx) → the just-created session is kept
    (no orphaned session or misleading "session create failed"); the message
    sends without attachments and the files are reported.

## Test
- `tsc -b` clean.
- Manual: Home & Session clip + drag upload; 30-file cap; oversized-file partial
  failure; blocked `POST /dirents` total failure (session kept + toast shown).

Resolved #175


app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
khj809: It seems `useFileDropzone` implements what these inline functions do, but it's not still applied here, which results in code duplication.

app/src/i18n/locales/ko/project.json
```diff
@@ -66,9 +66,11 @@
   },
   "home": {
     "preview_pill_aria": "미리보기 — 아직 서버에 전달되지 않습니다",
-    "attach_coming_soon": "파일 추가 기능은 곧 추가됩니다.",
     "suggested_prompts_aria": "추천 프롬프트 (미리보기)",
-    "send_blocked_hint": "사용 가능한 모델이 없어요. Provider API 키를 설정하거나 다른 모델을 선택하세요."
+    "send_blocked_hint": "사용 가능한 모델이 없어요. Provider API 키를 설정하거나 다른 모델을 선택하세요.",
+    "attach_limit": "한 번에 최대 {{max}}개까지 첨부할 수 있어요.",
```
khj809: Found that this `attach_limit` is duplicated in `session.json` and `common.json`. Need to consolidate them in a single json file.

app/src/lib/useFileDropzone.ts
```diff
@@ -0,0 +1,40 @@
...
+    // Ignore leaves into descendant elements — only clear when leaving the target.
+    if (e.currentTarget.contains(e.relatedTarget as Node)) return;
+    setIsOver(false);
+  }, []);
+
+  const onDrop = useCallback((e: React.DragEvent) => {
+    if (disabled || !e.dataTransfer.types.includes('Files')) return;
+    e.preventDefault();
+    setIsOver(false);
+    onFiles(Array.from(e.dataTransfer.files ?? []));
```
khj809: Browsers return `e.dataTransfer.files` as an empty array if user tries to upload a folder. Since there's a tracking issue #188, are you planning to address this in the next PR?

selloriwoo: Yes. in that #188 issue, I'm planning to implement folder upload for project files, sessions, and home all at once.

app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
selloriwoo: Refactor in [8d18474](https://github.com/brekkylab/agent-k/pull/181/commits/8d18474df0ded26703863e99e8267280a4e34b4a)

Thanks for catching this!

app/src/i18n/locales/ko/project.json
selloriwoo: Consolidated the duplicated parts into `common.json`

nuri-yoo (dismissed):

nuri-yoo (dismissed):

nuri-yoo (approved):
LGTM

app/src/routes/_app.projects.$projectSlug.index.tsx
```diff
@@ -99,22 +105,57 @@ function ProjectHome() {
...
-          ? { initialMessage: firstMessage, initialAgentId: selectedAgentId }
-          : { initialAgentId: selectedAgentId },
+        state: {
+          ...(firstMessage ? { initialMessage: firstMessage } : {}),
+          initialAgentId: selectedAgentId,
+          ...(attachmentPaths.length > 0 ? { initialAttachments: attachmentPaths } : {}),
+        },
         // Morph the composer to its session position/shape (shared view-transition-name).
         viewTransition: true,
       });
```
ljhh-0611, line 161: **Failed uploads still auto-send the first message**

`onSuccess` reports `failed` as a toast but navigates + auto-sends regardless (`initialMessage` is always set), so the agent run starts even when uploads failed. The worst case is a *partial* failure — e.g. one file over `AGENT_K_MAX_UPLOAD_BYTES`, where the backend returns 200 with `{succeeded, failed }`: the run proceeds with the remaining files and the missing one is silent (only a toast that flashes during the view-transition). Staged files are already cleared, so they aren't recoverable from the UI.

  Suggestion (two layers):
  1. **Pre-check size at stage time** in `addFiles`, like the count cap already does — rejects the common cause (oversized files) before send. (No `MAX_UPLOAD_BYTES` on the frontend yet; simplest is a constant in `domain/files.ts` mirroring the backend.)
  2. **Don't auto-send on `failed.length > 0`** — the session already exists (uploads need its `inputs/` scope), so navigate in *without* `initialMessage`, restoring the draft and staging the successful uploads for a manual send.

app/src/components/chat/ProjectHomeComposer.tsx
ljhh-0611, line 99: Could you make aria-label and title to i18n?

grf53 (approved):
I have nothing to add other than the issues already presented.

app/src/routes/_app.projects.$projectSlug.index.tsx
selloriwoo: Went with option 1. I figured rejecting oversized files at stage time is the better fix, so they never reach the auto-send. Thanks for the catch!

app/src/components/chat/ProjectHomeComposer.tsx
selloriwoo: Done. moved the labels to i18n.

ljhh-0611 (approved):
☺️

PR #182, opened 2026-06-15
Add per-agent descriptions to the composer (EN/KO)

nuri-yoo:
## What

This PR closes #160. It adds a short, plain-language description for each of the four agent surfaces (Coworker, Speedwagon, Deep Research, Buddy) so a user picking a mode understands what it does. Today the picker shows only a brand label and an icon, and the only per-agent copy is a casual greeting; there is no explanation of what each agent is for.

Frontend-only, EN and KO, no backend change. The agent's meaning is client-side copy, so it lives in the `automation` i18n namespace rather than the model catalog API.

## Copy

One `agent.<id>.description` key per agent, in `en/automation.json` and `ko/automation.json`. Written for non-technical users: brief, no jargon, parallel between EN and KO.

**Coworker**
- EN: Makes the documents and files you need.
- KO: 작업에 필요한 문서와 파일을 만듭니다.

**Speedwagon**
- EN: Answers your questions using your project's files and the web.
- KO: 문서와 웹에서 답을 찾습니다.

**Deep Research**
- EN: Researches a topic in depth across the web and writes you a report.
- KO: 웹을 심층 조사해 보고서를 작성합니다.

**Buddy**
- EN: A general chat for anything else, from quick questions to ideas.
- KO: 무엇이든 가볍게 묻고 이야기합니다.

Copy is grounded in what each agent actually does in the `test_case` suite: Coworker builds documents/sheets/slides/charts, Speedwagon answers from the project corpus and the web with sources, Deep Research runs multi-hop web research into a report, and Buddy is the general fallback chat.

## Where it shows

- **Project home composer (primary surface).** Hovering or focusing an agent tab in `ComposerAgentPicker` shows a small command-style card floating above that tab, with an arrow pointing down at it, so the description reads as belonging to that specific agent rather than as a detached line of body text. The card is keyboard-reachable (`:focus-visible`).
- **Session header chip tooltip.** Inside a running session the agent chip's tooltip now shows the agent's description instead of the generic "Session agent" label, so the explanation is reachable mid-session too. The session route now also loads the `automation` namespace to resolve it.
- **Automation new/edit forms.** When you pick which agent an automation runs as, the description shows as an inline line under the agent picker, above the model picker, so the choice is explained without a hover.
- **Project settings, model recommendations.** Each agent card in the recommendation-chain editor shows its description under the heading (resolved cross-namespace from `automation`).

## Verification 

- `tsc --noEmit` clean, `vite build` clean, both locale JSON files valid.

## Screenshots (sample)

스크린샷 2026-06-15 오후 4 01 01
스크린샷 2026-06-15 오후 4 01 13

### DIFF
--- a/app/src/components/chat/ComposerAgentPicker.tsx
+++ b/app/src/components/chat/ComposerAgentPicker.tsx
@@ -78,18 +78,22 @@ export function ComposerAgentPicker({
       />
       {AGENT_SURFACES.map((agent) => {
         const isActive = agent.id === value;
+        const description = t(`agent.${agent.id}.description`);
         return (
           <button
             key={agent.id}
             type="button"
             aria-pressed={isActive}
+            aria-label={`${agent.label} — ${description}`}
             data-agent={agent.id}
-            title={t(`agent.${agent.id}.description`)}
             className={`cw-agent-tab${isActive ? ' is-active' : ''}`}
             onClick={() => onChange(agent.id)}
           >
             <Icon name={agent.icon} size={15} />
             <span>{agent.label}</span>
+            <span className="cw-agent-tab-pop" role="tooltip" aria-hidden>
+              {description}
+            </span>
           </button>
         );
       })}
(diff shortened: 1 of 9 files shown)


ljhh-0611 (commented):
Image
In case of Automation, the description below agent picker can be just in one line for English

ljhh-0611 (approved):
After reply above comment, I think that it is just okay for demo stage

grf53 (approved):

PR #185, opened 2026-06-15
Enhance automation

grf53:
## Description
- **Schedule picker rework**
  - add 'By minute' segment(#170)
  - minute/hour interval presets (divisors of 60/24 only)
    : elsewise, it may cause confusion. ('every 5 hours' ≠ (`0 */5 * * *`))
    - minute: 5 / 10 / 15 / 20 / 30 (4 minutes or less have been excluded.)
    - hour: 1 / 2 / 3 / 4 / 6 / 8 / 12
  - monthly *day* includes "last day" (cron `L`; e.g. `0 10 L * *`: Last day of month at 10 o'clock)
  - monthly *day of week* includes "last week" (cron `<dow>L`; e.g. `0 10 * * 1L`: Last Monday of month at 10 o'clock )
  - drag-to-select weekday chips (for 'Weekly')
  - dropdown -> segmented-control UI to pick units(Monthly, Weekly, ...)
  - tidier human summaries (`N o'clock` etc.)
- **Trigger row actions**
  - **duplicate** a trigger (created **disabled** so it can't double-fire)
  - IconButton-based edit/delete with two-step delete confirm.
- **Run detail**
  - surface session **artifacts** (view + copy-to-shared)
  - a single 4s poller keeps the session preview, event log, and artifacts live while a run is in-flight.

## Notes
- cron `L` / `<dow>L` verified against `croner` (unit tests added).
- 'Timezone picker' & 'Prompts with variables' are tried to be included, but not accepted for this PR.

### By minute
<details>
  <summary>Show</summary>

스크린샷 2026-06-16 오후 1 22 11

</details>

### The last day of Month
<details>
  <summary>Show</summary>

스크린샷 2026-06-16 오후 1 23 06

</details>

### Day of the last week of Mongth
<details>
  <summary>Show</summary>

스크린샷 2026-06-16 오후 1 22 54

</details>

### Duplicate trigger (disabled at first)
<details>
  <summary>Show</summary>

스크린샷 2026-06-16 오후 1 24 25

</details>

PR #186, opened 2026-06-15
Runs in calendar view

grf53:
## Description
A calendar view on the automation page (List/Calendar toggle in the runs pane). It unifies the schedule timeline: **past slots show real runs** (with status), **future slots show predicted cron occurrences**, split at "now" and de-duped at the boundary.

- **Backend**: two project-scoped(`?project_ref=...`), date-windowed(`?from=...&to=...`) GET endpoints
  - (new) `GET /automations/occurrences` (expands cron exprs in-memory; nothing persisted)
  - (modified) `GET /automations/runs`
- **App**: Calendar view(refer to screenshots)

## Notes
- Colors are assigned per-automation by index (collision-free up to the palette size`=7`) and kept distinct from run-status hues.

### Segemented toggle at the top of the list
<details>
  <summary>Show</summary>

스크린샷 2026-06-15 오후 10 41 22

</details>

### Overall calendar view
<details>
  <summary>Show</summary>

automation legends with color palette / showing +N more / schedule info for future run / filters('schedule' is default for trigger) / today button / the actual runs for past slots, the expected run schedules for future slots

스크린샷 2026-06-15 오후 10 41 54

</details>

### Speech-bubble-like detail view for narrow screens
<details>
  <summary>Show</summary>

스크린샷 2026-06-15 오후 10 42 25

</details>

### Show all runs in the day with `+N more` button
<details>
  <summary>Show</summary>

스크린샷 2026-06-15 오후 10 42 06

</details>

### DIFF
--- a/app/src/api/automations.ts
+++ b/app/src/api/automations.ts
@@ -137,10 +137,15 @@ export async function listOccurrences(
 export async function createTrigger(
   automationId: AutomationId,
   spec: TriggerSpec,
+  opts?: { enabled?: boolean },
 ): Promise<CreatedTrigger> {
+  const body = {
+    ...specToBody(spec),
+    ...(opts?.enabled !== undefined ? { enabled: opts.enabled } : {}),
+  };
   const raw = await request<CreatedTriggerResponse>(
     `/automations/${automationId}/triggers`,
-    { method: 'POST', body: specToBody(spec) },
+    { method: 'POST', body },
   );
   return toCreatedTrigger(raw);
 }
--- a/app/src/api/automations.ts
+++ b/app/src/api/automations.ts
@@ -6,6 +6,7 @@ import { request } from './client';
 import type {
   BackendAutomation,
   BackendAutomationList,
+  BackendOccurrenceList,
   BackendRun,
   BackendRunEventList,
   BackendRunList,
@@ -17,6 +18,7 @@ import type {
 import {
   toAutomation,
   toCreatedTrigger,
+  toOccurrence,
   toRun,
   toRunEvent,
   toTrigger,
@@ -25,6 +27,7 @@ import type {
   Automation,
   AutomationId,
   CreatedTrigger,
+  Occurrence,
   Run,
   RunEvent,
   RunId,
@@ -115,6 +118,21 @@ export async function listTriggers(automationId: AutomationId): Promise<Trigger[
   return res.items.map(toTrigger);
 }
 
+/**
+ * Expand upcoming scheduled fires for all enabled cron triggers in a project,
+ * within `[from, to)`. Computed server-side from cron expressions (nothing is
+ * persisted). `from`/`to` are ISO instants.
+ */
+export async function listOccurrences(
+  projectRef: string,
+  from: string,
+  to: string,
+): Promise<{ items: Occurrence[]; truncated: boolean }> {
+  const qs = new URLSearchParams({ project_ref: projectRef, from, to }).toString();
+  const res = await request<BackendOccurrenceList>(`/automations/occurrences?${qs}`);
+  return { items: res.items.map(toOccurrence), truncated: res.truncated };
+}
+
 /** Returns the created trigger plus the one-time `webhookToken` (webhook only). */
 export async function createTrigger(
   automationId: AutomationId,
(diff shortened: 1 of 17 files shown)


app/src/routes/_app.projects.$projectSlug.automation.index.tsx
```diff
@@ -138,8 +145,18 @@ function AutomationsPage() {
   const [selectedAutomationId, setSelectedAutomationId] = useState<string | null>(null);
   const [selectedRunId, setSelectedRunId] = useState<string | null>(null);
   const [statusFilter, setStatusFilter] = useState<RunStatus | 'all'>('all');
-  const [triggerFilter, setTriggerFilter] = useState<TriggerKind | 'all'>('all');
+  // Default to schedule so the calendar reads as a schedule timeline; the user
+  // can widen to all / manual / webhook. Shared with the list view.
+  const [triggerFilter, setTriggerFilter] = useState<TriggerKind | 'all'>('cron');
```
khj809: The default filter value was changed for the calendar view, but it also affects the list view. When the user enters the list view, he may not aware that the current trigger list is a filtered result.

grf53: Thank you for pointing that out.

This is what remains from the initial version where only schedule runs were displayed in the calendar view without a separate filter.
However, since the same filter is currently on both sides, setting the schedule as the default seems meaningless. Therefore, I will revert the filter settings to their previous state (all as default). 850a4dd

khj809 (approved):

ljhh-0611 (commented):
**Calendar boundary dedupe is fragile — runs are stamped with the worker's wake time, and now is pinned at mount**

Two coupled issues weaken "split at now + dedupe at boundary":

1. The worker stamps scheduled_for = now (its tick time), but the calendar predicts the exact cron instant and dedupes per-minute. The 15s tick usually lands in the same minute, but a >60s-late tick (restart/suspend) stamps the run in a different minute than the occurrence → both render for one fire.

2. now is pinned at mount (useMemo(() => Date.now(), [])), so a calendar left open across a fire keeps the now-past occurrence shown as a future prediction.

  **Suggested fix:**

// backend/src/worker.rs — fire_cron_trigger_once(): stamp the cron instant, not the wake time.
// next_fire_at is non-null here (the due query filters it); fall back defensively.
trigger.next_fire_at.unwrap_or(now),   // scheduled_for arg
// app/src/components/AutomationCalendar.tsx — advance the split point each minute
const [now, setNow] = useState(() => Date.now());
useEffect(() => {
  const id = setInterval(() => setNow(Date.now()), 60_000);
  return () => clearInterval(id);
}, []);

Safe: claim_due_run already selects scheduled_for <= now, and the cron instant is <= now when due — no execution delay, and the list view's "scheduled at" becomes exact.

Confirmed locally: with the current code, a tick delayed past the minute stamps the run at 09:01 while the occurrence stays 09:00, so both render; stamping the cron instant makes them match and dedupe.

app/src/api/automations.ts
```diff
@@ -180,6 +198,20 @@ export async function listRuns(
   return res.items.map(toRun);
 }
 
+/**
+ * Cron-triggered runs across a project whose scheduled time falls in
```
nuri-yoo: nit: this returns runs of every trigger kind, not just cron — the SQL in `list_runs_in_window` has no kind filter, and a manual run shows up in the calendar's past slots. The Rust handler doc already says "any trigger kind (the client filters)". Mind dropping "Cron-triggered" here to match?

nuri-yoo (approved):
Pulled the branch and ran it end to end (backend lib tests + a live stack driven through the browser). Looks solid.

What I checked:
- Backend lib tests pass (67), including the three new `occurrences_between` cases.
- `GET /automations/occurrences` and `/automations/runs` against seeded triggers: timezone expansion is correct (`0 10 * * *` in Asia/Seoul → `01:00Z`, `0 9 * * 5` → Friday `00:00Z`), results sorted, `truncated` flag works. Validation returns 400 for missing `project_ref`, `to <= from`, window > 366 days, and unknown query fields.
- Calendar in the browser: past slots render real runs with status, future slots render predicted occurrences, split at now with no duplicates at the boundary. Past-only months skip the occurrences request (clamp to today). Trigger filter hides occurrences for manual/webhook as intended. Occurrence detail, run detail, "+N more" day popover, month nav, and the narrow-layout speech bubble all work with no console errors.
- `tsc` is clean for the files this PR touches (the one error in `automation.$automationId.tsx` predates this PR, from #185).

One nit inline on `automations.ts` (JSDoc says "Cron-triggered" but the endpoint returns all kinds). Non-blocking.

app/src/api/automations.ts
grf53: Thanks!

ljhh-0611 (approved):
☺️👍

PR #193, opened 2026-06-18
feat(home): attach project shared files to the first message

selloriwoo:
## Overview
Adds the ability to attach project shared files to the first message from the project-home composer.
Shared files already live on the server, so they're passed by global-path reference (no upload).

## Changes
- Shared-file picker dialog (16:9): a Finder column view — opening a folder appends a column to the right; clicking a file previews it in the trailing pane.
- Selection IS attachment: a checkbox tick stages the file. Multi-select follows
  Finder conventions:
  - Click = preview (+ sets the range anchor)
  - Shift-click = range
  - Cmd (mac) / Ctrl (win/linux) click = toggle an individual file
  - A folder's + attaches every file inside it
- Shared preview layer: extracted useFilePreview + PreviewBody so the full-screen FilePreviewModal and the inline FilePreviewPane share the same fetch/decode/cap logic (the modal shrank ~114 lines into the common layer). 
- Router hand-off: files staged on the home composer ride initialAttachments (router state) into the newly created session and attach to the auto-sent first message. Enforces MAX_ATTACHMENTS; the cap/dedupe is decided against a synchronous ref so two picks in the same tick can't blow past the limit.
- Fix toast z-index (now above modal/dialog layers); add i18n labels (ko/en).

## Testing
- [x] Single / multi-select (Shift range, Cmd/Ctrl individual), then attach
- [x] Folder `+` bulk attach; toast when exceeding `MAX_ATTACHMENTS`
- [x] Send first message after attaching → files appear in the session

<details>
  <summary><h3>image<h3/></summary>
  스크린샷 2026-06-19 오후 12 22 42
</details>

Resolved #187


ljhh-0611 (approved):
🙇‍♂️

grf53 (approved):

PR #198, opened 2026-06-21
App-use tool

grf53:
## Description
Give internal agents (Buddy first) in-process tools to operate the app on the session owner's behalf — instead of calling the REST API or being denied any app control at all. Every call goes through the same `repository` operations and the same authorization the HTTP handlers use.

> [!NOTE]
> Some User RBAC-related concepts (`PermissionResolver`) are included but are merely placeholders and must be implemented separately.

스크린샷 2026-06-21 오후 7 19 29

## Design

**Buddy is the first victim.** I just chose 'Buddy' to get these tools first. There was no special reasons, tools can be given to every agent. It could be up to agent selection design.

**No HTTP.** Agents are embedded in the backend process, so the tools call the
same repository operations the handlers do — no REST round-trip.

**Judged against the session creator.** An agent may only do what the user who
created its session can do, whether the session was started by a person or by an
automation.

**Two-layer permission gate (AND).** Every tool call must pass *both*:

- **Layer A — user permission** (`PermissionResolver`): can this user do it at
  all? Today this wraps project membership/ownership. This is the seam where
  user-level RBAC will plug in — swap the implementation, keep the
  `(actor, capability, resource)` signature.
- **Layer B — agent scope** (`AgentPolicy`): of what the user can do, what is
  *this agent* granted?

```mermaid
flowchart LR
  USER["User · session creator"]
  AGENT["Agent · Buddy"]
  CALL["tool call<br/>action + target"]

  subgraph A["Layer A · user permission"]
    RES["membership · ownership<br/>(where RBAC plugs in)"]
  end
  subgraph B["Layer B · agent scope"]
    POL["granted action set"]
  end

  AGENT -->|has| POL
  CALL --> GATE{"both layers satisfied?"}
  GATE -.->|"granted?"| POL
  GATE -.->|"allowed?"| RES
  USER -->|has| RES
  GATE -->|"A AND B"| OK["allow → repository op"]
  GATE -->|"either fails"| DENY["permission_denied"]
```

The intersection is always safe: granting an agent a capability the user lacks
never executes it, because the resolver still has the final say per resource.

**Layer B comes from ceiling × member grant.** The agent policy is no longer a
hardcoded value — it's computed per (user, project):

```
effective = (member_grant ?? ceiling) ∩ ceiling
```

- **Project ceiling** (`projects.agent_capability_ceiling`) — owner-only; the max
  any member's agent may attempt. `NULL` = no limit.
- **Member grant** (`project_members.agent_capabilities`) — the member edits
  their own. `NULL` = inherit the ceiling.

A member's picks outside the ceiling are **preserved but inert** until an owner
widens the ceiling — the intersection keeps it safe, and the UI flags such items.

**Owner-as-member.** The owner is now a first-class `project_members` row, so the
owner can also scope their own agent within the project (migration backfills
existing owners).

### DIP(Dependency inversion principle): Backend -> Agents
**`ExtraTools` crosses the crate boundary.** Dependencies flow
`backend → agent-k` only, so agent-k can't import backend types. App-control
tools depend on `repository`, so the backend builds them and hands them to the
agent builder as `ExtraTools` (ailoy types only) — pairing tool descriptors with
their function registration so a tool can't be exposed without its handler.

```mermaid
flowchart LR
  subgraph BK["backend crate"]
    direction TB
    REPO["repository · resolver"] --> AT["app_tools<br/>depends on backend types"] --> BUILD["build ExtraTools"]
  end
  subgraph AK["agent-k crate"]
    direction TB
    OWNED["agent-k-owned tools<br/>builtin · corpus"] --> BUDDY["Buddy builder<br/>receives + injects"]
  end
  BUILD ==>|"ExtraTools · ailoy types only"| BUDDY
  BK -.->|"dep: backend → agent-k · no reverse import"| AK
```

### Safety
- Sessions are read-only (no send/run/create/delete tool), so agent→agent loops and self-session deadlock are structurally impossible.
- Automation-session agents can't create/schedule/run automations.
- Destructive ops keep the handlers' owner/admin checks.

---

### Permission core (`backend/src/authz.rs`)
- `Capability` (stable vocabulary, `name()` persisted as grant strings),
  `Resource` (`Some(id)` = object, `None` = collection).
- `PermissionResolver` trait + `RepoPermissionResolver` (membership/ownership).
  - **`PermissionResolver::user_can` is currently hard-coded.** it should be revised with user RBAC implementation
- `ResourceKind` removed — the resolver's `(cap, resource)` match is the single
  source of valid pairings; mismatched pairs fall through to deny.

### Agent policy (`backend/src/app_tools/policy.rs`)
- `AgentPolicy::effective(member, ceiling)`, `read_only()`, `all()`,
  `grants(cap)`.

### Tools (`backend/src/app_tools/*`)
- `AppToolContext::authorize(cap, resource)` = `grants(cap) && user_can(...)`.

```mermaid
flowchart TD
  CALL["tool call (cap, resource)"] --> B{"agent_policy.grants(cap) ?<br/>Layer B · agent scope"}
  B -->|"no"| DENY["permission_denied"]
  B -->|"yes"| A{"resolver.user_can ?<br/>Layer A · user permission (RBAC seam)"}
  A -->|"no"| DENY
  A -->|"yes"| OK["run repository op"]
```

- Automation (read + write/run), Sessions (read-only), Projects/Members, Users.
- **Bug fix:** `list_automations` / `list_sessions` were gated on `ProjectRead`
  instead of `AutomationRead` / `SessionRead`. Since `authorize(cap, …)` feeds
  the Layer-B policy gate too, an agent granted the resource-specific read but
  not `ProjectRead` was wrongly blocked (and the inverse wrongly allowed). They
  now gate on the correct capability, scoped to the project for the membership
  check (new project-scoped resolver arms).

### Schema (`migration 0012`)
- `projects.agent_capability_ceiling`, `project_members.agent_capabilities`,
  owner backfill into `project_members`; `create_project` inserts the owner row
  in a transaction.

### Wiring
- `repository` reads/writes ceiling + member grants; `list_project_members`
  returns grants. `build_session_agent` computes the effective policy for Buddy
  and injects it via `ExtraTools`.
- `auth` module renamed to `authn`; user-permission layer extracted into the
  shared `authz` module.

### Frontend — permission editing UI (project settings)
- `ProjectAgentCeilingEditor` (owner-only) + `MemberAgentGrantEditor` (self;
  owner included as a normal member row).
- Capabilities grouped by resource (user → project → session → automation), one
  icon per action verb, whole-box toggles; "No limit" / "Inherit ceiling" as
  switch cards. Out-of-ceiling picks are struck through and flagged inactive.
- The earlier user-global read-only view (commit `13d9267`) is removed; `/me`
  reverts to `UserResponse`.

<details>
  <summary>Show screenshots</summary>

스크린샷 2026-06-23 오후 2 34 39
스크린샷 2026-06-23 오후 2 34 52

</details>

## Extras
- Deep-link automations/runs via URL hash
  - `.../automation#{automationId}`: Automation selected
  - `.../automation#{runId}`: Run selected

<details>
  <summary>Show screenshots</summary>

스크린샷 2026-06-23 오후 2 35 25
스크린샷 2026-06-23 오후 2 35 17

</details>

- 404 non-user session routes (not in session list, no session page)
- demo seed updated for owner-as-member

## Testing
- backend: build + clippy clean; full suite green incl. `app_tools` 8/8.
- frontend: lint + tsc + unit 176/176.


backend/migrations/0012_agent_capabilities.sql
```diff
@@ -0,0 +1,20 @@
...
+-- make the owner a first-class project_members row (no special owner column or
+-- branch). Backfill existing projects whose owner has no row yet; new projects
+-- insert the owner row in create_project.
+-- INSERT INTO project_members (project_id, user_id, added_at)
+-- SELECT p.id, p.owner_id, p.created_at
+-- FROM projects p
+-- WHERE NOT EXISTS (
+--     SELECT 1 FROM project_members pm
+--     WHERE pm.project_id = p.id AND pm.user_id = p.owner_id
+-- );
```
nuri-yoo, line 20: **This backfill is commented out, so it never runs** — pre-existing project owners are left without a `project_members` row.

Combined with `list_project_members` dropping the old `owner_id` UNION (it now reads only from `project_members`), the owner of any project created before this migration disappears: they're missing from `GET /projects/{ref}/members`, and `PATCH .../members/{owner}/agent-capabilities` finds no row to update → returns `false` → handler 404s, so they can't save their own agent grant. New projects are fine (`create_project` inserts the owner row in a transaction); only pre-migration data is affected. Authorization itself still works, since `user_in_project` / `user_is_project_owner` read `projects.owner_id` directly.

Reproduced on this branch with a project inserted the pre-0012 way (no owner member row): owner missing from the list, `set_member_agent_capabilities(owner)` returns `false`, while a `create_project` project lists its owner fine.

Uncommenting the INSERT fixes it (already idempotent via `WHERE NOT EXISTS`). Worth a regression test that creates a project without an owner member row — the current tests all go through `create_project`, so they mask this.

backend/src/app_tools/mod.rs
```diff
@@ -0,0 +1,72 @@
+//! Agent-invokable tools that drive this app's own features.
+//!
+//! These are in-process [`ailoy`] tools (not HTTP calls) handed to an embedded
+//! agent — currently Buddy, the app-control surface. Each tool runs the same
+//! repository operations and authorization the REST handlers use, judged
+//! against the session creator's identity. See [`policy`] for the two-layer
+//! permission model and [`context::AppToolContext`] for the per-call gate.
+//!
+//! v1 exposes **read-only** tools. Write/run/manage tools are deferred until
```
nuri-yoo: **This "v1 read-only" note no longer matches the wiring.** `build_session_agent` builds the policy with `AgentPolicy::effective(member, ceiling)`, not `read_only()`, and `effective(None, None)` returns `all()` (asserted by `effective_intersects_member_and_ceiling`). So for a fresh project with no ceiling set, Buddy gets the full write/run/delete + member-manage toolset by default, bounded only by Layer A. Reasonable design, but the default posture is open, not read-only — either update this note (and the PR's Safety section) to say so, or default to `read_only()` and require an explicit grant to widen it. Flagging since it decides whether destructive tools are exposed by default.

backend/migrations/0012_agent_capabilities.sql
grf53: Thank you for pointing this out, and I am sorry to cause a confusion.

Since migration files are intended to sequentially bring a database in service to a 'wanted' state, the statement in question should, in principle, be applied.

However, I intentionally commented out that section.

Its function is merely to find the owner ID of a pre-existing project and include it as a member; in development and test environments, this simply ensures that the data is correctly stored.

(In fact, since the demo data seed script runs after the migrations are applied, it was a meaningless situation when rebuilding with run_cowork_demo.sh.)

All migration files are scheduled to be consolidated before being released to the production level, and I was cautious about the current increase in the number of migrations. Consequently, I wanted to exclude that statement from the migrations, but I decided to simply comment it out to preserve the context. (I anticipated that a thread like this would be opened if necessary.)

nuri-yoo (approved):
LGTM

PR #199, opened 2026-06-22
Refactor: PPTX improve

selloriwoo:
## Summary

Reworks the `pptx` coworker skill from "the model writes python-pptx code with hand-computed coordinates" to "the model authors each slide in **HTML/CSS** (charts via Chart.js / SVG / images), then converts to an **editable** `.pptx` and verifies it in LibreOffice."

Text and simple shapes round-trip as **native, editable** PowerPoint objects. only chart/image internals stay raster. Design fidelity is exactly what Chromium renders.

BEFORE:  model → outline.md → python-pptx code (manual x/y/w/h) → .pptx → verify
AFTER:   model → HTML/CSS → html2pptx.py (Chromium → native objects) → editable .pptx → verify + contact-sheet

## What's in here

**New converter — `script/html2pptx.py`**
- Renders each `.slide` (1280×720, 96px/in) in headless Chromium and decomposes it into native objects: text boxes, autoshapes (boxes/borders/bullet markers), one picture per chart/img/svg, and a native slide-background fill.
- **Frozen line breaks**: text is emitted `word_wrap=off` with hard `<a:br/>` at the browser's exact break positions, so soffice/PowerPoint can't re-wrap CJK differently from what was authored.
- Line grouping uses a **vertical-overlap fraction** so a big number + a small unit span sharing a baseline stays on one line, while genuine wraps are kept.
- Faint `rgba()` fills keep their transparency (`<a:alpha>`) so a light tint stays light instead of rendering opaque and hiding the text on it.
- Inline `<svg>` is detected as media. Auto-saves a self-contained `<out>.source.html` sidecar for review.

**Skill guidance — `SKILL.md`, `components.css`, `template.html`**
- Per-deck **palette composition** (fixed role structure, hex values derived from the subject) instead of a fixed gallery, plus a distinctiveness self-check and a subject-grounding step.
- Type scale is composed per deck too — four `--fs-*` size variables sized to the deck's density and locked, mirroring the palette; `components.css` and its helper classes consume them.
- `components.css` is a role-variable system (type scale, grid, shape vocabulary) so decks read "designed, not assembled."
- Authors `deck.html` straight from the source (no intermediate outline file). bookend/takeaway-title rules folded into the authoring step.
- Design guidance to **fill the canvas**: use the space at scale, carry a sparse slide with a chart/table instead of white space, and anchor two-column cards and exec-summary stat rows so the lower half isn't dead. (See before/after.)
- A **completion guard** — the deliverable is the `.pptx`, not the HTML. the run isn't done until the file exists and verify passes.
- Requires the full toolchain install up front (incl. `playwright install-deps`) so the first conversion can't fail on missing browser libraries.

**Verification — `script/verify_pptx.py`, `script/contact_sheet.py`**
- Five checks (fonts/sizes/page_numbers/overlap/palette-discipline).
- `contact_sheet.py` renders the soffice-produced PDF with **pypdfium2** (writing a `slide-N.png` per page) and montages them into one sheet for cheap visual QA
  — `read` the sheet once, deep-read only suspect slides.

**Harness — `src/bin/test_case.rs`**
- Per-turn token-usage logging; startup sweep of orphaned `ailoy-*` sandbox dirs.

## Dependencies / licensing

Runtime tools are installed in the sandbox, not vendored. All permissive — no strong copyleft: python-pptx (MIT), playwright + Chromium (Apache-2.0 / BSD), pypdfium2 / PDFium (Apache-2.0 / BSD), Pillow (HPND). The image bakes in LibreOffice (MPL-2.0) and Noto/Liberation fonts (OFL). poppler (GPL) was dropped in favor of pypdfium2.

## Results (case 20, billable tokens / LM calls)

| Model  | Old (python-pptx) | New (HTML-authored) |
|--------|-------------------|---------------------|
| claude | 1,919,338 / 39    | 484K–1.14M / 14–28  |
| gemini | 2,107,522 / 45    | 899K–1.07M / 20–32  |
| openai | 1,213,234 / 37    | 726,029 / 27        |

~40–75% fewer tokens than the old python-pptx skill across all three models (run-to-run variance from QA-fix loops). Decks verified clean and rendered correctly in LibreOffice for claude, gemini, and openai. The design guidance (fill-the-canvas, type scale) lands well on stronger models; faster/weaker models (gemini-flash) follow it partially — it's prose guidance, not a hard gate.

<details>
<summary><h3>Image</h3></summary>

-gemini-
-before-
스크린샷 2026-06-23 오전 10 30 47
-after-
스크린샷 2026-06-23 오전 10 30 55

-gpt-
-before-
스크린샷 2026-06-23 오전 10 35 19
스크린샷 2026-06-23 오전 10 37 05
-after-
스크린샷 2026-06-23 오전 10 35 51

-claude-
-before-
스크린샷 2026-06-23 오전 10 41 19
-after-
스크린샷 2026-06-23 오전 10 41 01

</details>

Resolved #195

### DIFF
--- a/agent-k/src/agents/skill/pptx/script/html2pptx.py
+++ b/agent-k/src/agents/skill/pptx/script/html2pptx.py
@@ -122,7 +122,9 @@ SCRAPE_SLIDE_JS = r"""
     return true;
   };
   const isSvgInternal = (el) => !!el.ownerSVGElement;        // inside an <svg>
-  const isMedia = (el) => ['CANVAS', 'IMG', 'SVG'].includes(el.tagName) && !el.ownerSVGElement;
+  // NB: an inline <svg> reports tagName "svg" (lowercase, not HTML-uppercased),
+  // so compare case-insensitively or SVG charts are silently dropped.
+  const isMedia = (el) => ['CANVAS', 'IMG', 'SVG'].includes((el.tagName || '').toUpperCase()) && !el.ownerSVGElement;
 
   const pics = [], shapes = [], texts = [];
   let order = 0;
@@ -223,7 +225,15 @@ SCRAPE_SLIDE_JS = r"""
             right: Math.max(bb.right, rr.right), bottom: Math.max(bb.bottom, rr.bottom),
           } : { left: rr.left, top: rr.top, right: rr.right, bottom: rr.bottom };
         };
-        const segs = [];   // ordered {text, top, style}
+        // Two vertical spans are on the same visual line when they overlap by
+        // more than half the shorter span's height. Same-line size mixes overlap
+        // ~fully (~1.0); even tight-leading wraps overlap little (≲0.15).
+        const samELine = (t1, b1, t2, b2) => {
+          const ov = Math.min(b1, b2) - Math.max(t1, t2);
+          const minH = Math.min(b1 - t1, b2 - t2);
+          return minH > 0 && ov > 0.5 * minH;
+        };
+        const segs = [];   // ordered {text, top, bottom, style}
         const addText = (tn, styleEl) => {
           const s = tn.textContent;
           if (!s) return;
@@ -231,20 +241,29 @@ SCRAPE_SLIDE_JS = r"""
           if (col0 && col0.a < 0.1) return;   // fully-transparent text → no text
           const st = mkStyle(styleEl);
           const rg = document.createRange();
-          let lineStart = 0, curTop = null;
+          let lineStart = 0, curTop = null, curBot = null;
           for (let i = 0; i < s.length; i++) {
             rg.setStart(tn, i); rg.setEnd(tn, i + 1);
             const rects = rg.getClientRects();
             if (!rects.length) continue;
             grow(rects[0]);
-            const top = Math.round(rects[0].top);
-            if (curTop === null) curTop = top;
-            else if (Math.abs(top - curTop) > 2) {
-              segs.push({ text: s.slice(lineStart, i), top: curTop, style: st });
-              lineStart = i; curTop = top;
+            const top = rects[0].top, bot = rects[0].bottom;
+            if (curTop === null) { curTop = top; curBot = bot; }
+            // A real wrap drops to a new line, so the glyph barely overlaps the
+            // current line vertically. A different font size on the SAME line
+            // overlaps heavily (the small run sits inside the tall one), so we
+            // split on overlap FRACTION, not on a raw top change — that keeps a
+            // big number + small unit on one line yet still catches tight-leading
+            // wraps that a plain band-overlap test would wrongly merge.
+            else if (samELine(top, bot, curTop, curBot)) {
+              curTop = Math.min(curTop, top); curBot = Math.max(curBot, bot);
+            } else {
+              segs.push({ text: s.slice(lineStart, i), top: curTop, bottom: curBot, style: st });
+              lineStart = i; curTop = top; curBot = bot;
             }
           }
-          segs.push({ text: s.slice(lineStart), top: curTop === null ? 0 : curTop, style: st });
+          segs.push({ text: s.slice(lineStart), top: curTop === null ? 0 : curTop,
+                      bottom: curBot === null ? 0 : curBot, style: st });
         };
         for (const n of el.childNodes) {
           if (n.nodeType === 3) { if (n.textContent.trim()) addText(n, el); }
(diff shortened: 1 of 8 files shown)


agent-k/src/agents/skill/pptx/script/html2pptx.py
```diff
@@ -0,0 +1,682 @@
...
+    if work is None:
+        tmp = tempfile.TemporaryDirectory()
+        work = Path(tmp.name)
+    work.mkdir(parents=True, exist_ok=True)
+
+    try:
+        slides = render_and_scrape(args.html, work, args.scale)
+        if args.dump_json:
+            args.dump_json.write_text(json.dumps(slides, indent=2, ensure_ascii=False),
+                                      encoding="utf-8")
```
nuri-yoo, line 677: `--dump-json` crashes with `FileNotFoundError` when its parent directory doesn't exist yet — reproducible with `--dump-json out/scrape.json` on a fresh tree.

`build_pptx` creates the output dir via `out_path.parent.mkdir(parents=True, exist_ok=True)`, but this `dump_json` write has no equivalent. It also runs *before* `build_pptx`, so pointing both `-o` and `--dump-json` at the same new directory dies here before the pptx is ever written.

```python
if args.dump_json:
    args.dump_json.parent.mkdir(parents=True, exist_ok=True)
    args.dump_json.write_text(json.dumps(slides, indent=2, ensure_ascii=False),
                              encoding="utf-8")
```

agent-k/src/agents/skill/pptx/script/html2pptx.py
```diff
@@ -0,0 +1,682 @@
...
+              const rr = n.getBoundingClientRect(); grow(rr);
+              segs.push({ text: n.textContent, top: Math.round(rr.top),
+                          bottom: Math.round(rr.bottom), style: mkStyle(n) });
+            }
+          }
+        }
+        if (segs.length && bb) {
+          // Group segments into visual lines by the same overlap-fraction test,
+          // then merge adjacent same-style runs. A big number + a small unit
+          // span share a line (heavy overlap); a real wrap starts a new one.
```
nuri-yoo: Behavior is correct (verified by rendering) — "share a line" reads a bit stronger than what actually happens.

This grouping only merges segments *within a single text block*. So a big number and its unit merge into one `line` object (two runs) when they're inline runs under the same parent (`<div><span>$14.2</span><span>M</span></div>`). When they're instead *sibling blocks* — each holding its own text node, so the topmost-text-bearing logic above promotes each to its own `data-pptx-text` — they become separate text boxes, and `samELine` never merges across siblings.

The sibling case still renders on the same baseline because the coordinates are preserved (confirmed in the soffice render), so the result is what you want either way. The guarantee here is really "no spurious `<a:br/>`, same baseline on render," not "collapsed into one `line` object."

agent-k/src/agents/skill/pptx/script/html2pptx.py
selloriwoo: fixed in [de1443f](https://github.com/brekkylab/agent-k/pull/199/commits/de1443f98a7d662046fdb6be2624f45b0375df1c)

agent-k/src/agents/skill/pptx/script/html2pptx.py
selloriwoo: fixed in [de1443f](https://github.com/brekkylab/agent-k/pull/199/commits/de1443f98a7d662046fdb6be2624f45b0375df1c)

agent-k/src/agents/skill/pptx/script/verify_pptx.py
```diff
@@ -554,92 +453,49 @@ def classify(sh, b):
...
-                issues.append((i, text[:40].replace("\n", " ")))
+            top = sh.top / EMU_PX
+            bottom = top + sh.height / EMU_PX
+            has_text = sh.has_text_frame and sh.text_frame.text.strip()
+            if has_text:
+                for para in sh.text_frame.paragraphs:
+                    for run in para.runs:
+                        if run.font.size:
+                            max_font = max(max_font, run.font.size.pt)
+            if (has_text and top < FOOTER_TOP_PX) or sh.shape_type == PICTURE:
```
nuri-yoo: Ran this against a few normal layouts and it false-positives on card/container fills. The condition only counts text boxes and pictures as content, so a card whose body text sits at the top but whose box stretches to the floor isn't recognized — `max_bottom` is computed from the text, not the card.

Concretely, on a two-column-cards slide the card autoshapes run `top=154 → bot=658` (right above the footer), but the text inside ends at `bot=255`, so `dead_band` comes out 385px and the slide is flagged — even though it's visually full. The reference `template.html` slide 2 (the exec-summary stat row + bullets) flags the same way at `content_bottom=378`.

That matters because SKILL.md makes verify a hard gate ("do not surface the `.pptx` until every `verify()` check is clean"), and two-column cards are one of the fill-the-canvas patterns the skill recommends. A model could get stuck trying to "fix" a slide that's already full, since growing the card doesn't move the text-based `max_bottom`.

The hollow-card intent is reasonable, but as written it can't tell "card stretched with text at the top" from "card stretched and empty." Could the bottom of a large autoshape that contains other content count toward `max_bottom`? Or should dead_bottom be advisory (a visual hint) rather than part of the clean-gate, given the contact-sheet read already covers the lower-half-empty case?

selloriwoo: This was something I hadn't fully considered. Trying to detect empty space with a geometric check is hard to get right — decorative shapes, rules, and the like get picked up too, so there's no clean threshold.
So I removed the existing code and moved the intent into SKILL.md more explicitly instead. Across two test runs on different models it seems to work well now. Thanks! [5fd67fd](https://github.com/brekkylab/agent-k/pull/199/commits/5fd67fd54566163114d76820f2be6adab43e666e)

nuri-yoo (approved):
LGTM

grf53 (approved):

PR #200, opened 2026-06-29
feat: live token streaming for agent responses

selloriwoo:
## Summary
Agent responses now stream to the client token-by-token instead of
appearing only when each turn completes. Adopts ailoy's(https://github.com/brekkylab/ailoy/pull/434) `run_stream` agentic loop end-to-end: backend forwards token deltas over the existing WebSocket, the client renders them with a smooth typewriter effect, and an in-progress turn survives a page refresh.

## Changes
**Backend**
- Switch the session run loop (and the `run` CLI) from `Agent::run` to `Agent::run_stream`, forwarding `AgentEvent::Delta` as a new ephemeral `WsEvent::AgentDelta` (not persisted, no seq). The `Message` arm keeps the existing persist / broadcast / history logic unchanged.
- Stop now aborts promptly: cancel drops the stream (cancelling the in-flight model request) instead of draining the turn to completion (a regression the old non-streaming grace-wait caused once every token became a sub-grace delta).
- Preserve the partial answer on stop: accumulate streamed text and persist it with the interrupt note via `synthetic_msgs`, so a mid-turn stop survives a refresh and the agent's history stays consistent.
- Send deltas incrementally: `AgentDelta` carries an incremental `delta` + `cum_len` (UTF-16 running total) rather than resending the full text each tick, keeping the wire cheap with many subscribers.
- Resume in-flight streaming: keep the turn's accumulated text in `ActiveAgentRun.partial`; add `GET /sessions/{id}/active-run` returning the in-flight snapshot, and have the WS subscribe-replay send the partial so a reconnecting client catches up.
- Bump ailoy to `2086bfd` so streaming reports token usage.

**Frontend**
- Consume `agent_delta`: apply each incremental fragment and reveal it with a `requestAnimationFrame` typewriter loop (backlog-proportional, ~33fps throttle) so bursty network chunks read as steady typing.
- Resume on mount: seed from the `active-run` HTTP snapshot (idempotent with the WS replay), dedup across the replay↔live boundary via `cum_len` (skip already-applied, trim slice overlap). The resumed backlog is painted instantly; only subsequent tokens animate.
- Reconciled by the completed `agent_message`.
- Don't trap the reader at the bottom while streaming: every token used to force the view back down whenever it was within 700px of the tail, so scrolling up to re-read was yanked straight back (only a hard scroll escaped). Replace the distance heuristic with direction-based follow detection — the auto-scroll only moves scrollTop down and growing content fires no scroll event, so any decrease in scrollTop is a genuine user scroll-up. Stop following on the slightest upward move and re-engage only once back at the very bottom; scroll instantly (not smoothly) while streaming so the per-token animation no longer fights the user.

## Behavior changes
- **Stop** now halts immediately rather than finishing the current turn.
- A stopped turn persists the partial text (+ `[Interrupted: …]`).
- Refreshing mid-stream resumes the in-progress turn instead of losing it.
- Scrolling up during a live stream now holds position instead of being forced back to the bottom; auto-follow resumes once you return to the bottom.

## Testing
- `cargo build` + `tsc`/`lint` clean; existing unit tests pass.
- Manual: long answer streams smoothly; stop halts at once and the partial answer + interrupt note survive a page refresh; refreshing mid-stream resumes the in-progress turn and keeps animating; scrolling up mid-stream stays put and resumes auto-follow at the bottom.

### DIFF
--- a/app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
+++ b/app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
@@ -211,6 +211,11 @@ function SessionPage() {
   // a mid-stream resume it's the whole backlog, which should appear at once
   // rather than slowly retyping. Subsequent chunks animate. Reset per turn.
   const revealSeededRef = useRef<boolean>(false);
+  // True while a hole-recovery resync (refetch of the server-side partial) is in
+  // flight. After a dropped delta every subsequent delta also reads as a hole,
+  // so this gates the refetch to one at a time — otherwise a fast stream would
+  // fire an HTTP snapshot per token until the first one lands.
+  const resyncingRef = useRef<boolean>(false);
 
   const [composerText, setComposerText] = useState('');
   // Composer layout — false = compact pill (textarea + buttons inline),
@@ -269,6 +274,7 @@ function SessionPage() {
     wsOutputsRef.current.clear();
     maxSeqRef.current = -1;
     pendingDeltaRef.current = '';
+    resyncingRef.current = false;
     optimisticUserIdRef.current = null;
     currentRunIdRef.current = null;
     doneRunIdsRef.current.clear();
@@ -895,6 +901,7 @@ function SessionPage() {
           pendingDeltaRef.current = '';
           revealedTextRef.current = '';
           revealSeededRef.current = false;
+          resyncingRef.current = false;
         }
         currentRunIdRef.current = run_id;
 
@@ -964,6 +971,7 @@ function SessionPage() {
         pendingDeltaRef.current = '';
         revealedTextRef.current = '';
         revealSeededRef.current = false;
+        resyncingRef.current = false;
         if (revealRafRef.current !== null) {
           cancelAnimationFrame(revealRafRef.current);
           revealRafRef.current = null;
@@ -1032,8 +1040,35 @@ function SessionPage() {
         if (cum_len <= applied) return; // already have this text (replay overlap)
         const cumStart = cum_len - delta.length; // length before this delta
         if (cumStart > applied) {
-          // Hole: we missed an earlier delta (e.g. broadcast lag). Can't fill it
-          // safely — leave it; the completed agent_message reconciles the turn.
+          // Hole: an earlier delta was dropped (broadcast lag). Leaving it would
+          // freeze the whole reveal — every later delta reads as a hole too —
+          // until the commit reconciles. Instead refetch the exact server-side
+          // partial and replay it as one catch-up delta (cum_len == length →
+          // cumStart 0 → the append below trims the overlap), so the reveal
+          // recovers now. Guard against a refetch storm: only one resync in
+          // flight, since every post-hole delta lands here until it completes.
+          if (resyncingRef.current) return;
+          resyncingRef.current = true;
+          // Paint the recovered backlog at once instead of slowly retyping it.
+          revealSeededRef.current = false;
+          void getActiveRunSnapshot(sessionId)
+            .then((snap) => {
+              if (snap?.partial && currentRunIdRef.current === run_id) {
+                processEvent({
+                  type: 'agent_delta',
+                  session_id: sessionId,
+                  run_id,
+                  delta: snap.partial,
+                  cum_len: snap.partial.length,
+                });
+              }
+            })
+            .catch(() => {
+              /* best-effort recovery; the commit still reconciles the turn */
+            })
+            .finally(() => {
+              resyncingRef.current = false;
+            });
           return;
         }
         // Slice off any overlap already applied, then append the new tail.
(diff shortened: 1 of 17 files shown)


backend/src/state.rs
ljhh-0611: > AI recommendation:

One WS event per streamed token means a slow subscriber can overflow the 128-slot ring, drop deltas, and freeze its live reveal until commit. Enlarge the buffer.

```suggestion
        // One event per streamed token: a fast model enqueues hundreds per turn,
        // so a small ring lets a slow subscriber lag and drop deltas.
        let (ws_tx, _) = broadcast::channel(1024);
```

ljhh-0611: **+** A slow subscriber can hit the frontend's hole guard — which just returns, so the live text freezes until the turn commits (ironically worse for the fast streams this feature targets). Fix: on a hole refetch the exact server-side partial and replay it as one catch-up delta.

```ts
// app/src/routes/…sessionPrefix.tsx — agent_delta handler, the hole branch
if (cumStart > applied) {
  // Hole: an earlier delta was dropped (broadcast lag). Refetch the exact
  // server-side partial and replay it as one catch-up delta so the reveal
  // recovers instead of freezing until commit.
  void getActiveRunSnapshot(sessionId)
    .then((snap) => {
      if (snap?.partial && currentRunIdRef.current === run_id) {
        processEvent({
          type: 'agent_delta', session_id: sessionId, run_id,
          delta: snap.partial, cum_len: snap.partial.length,
        });
      }
    })
    .catch(() => {});
  return;
}
```

selloriwoo: Done. refetch + replay as suggested. Added an in-flight guard so a hole doesn't fire one snapshot request per subsequent delta (they all read as holes until the first lands); it re-triggers if a gap remains and resets on turn/session boundaries.
[e300b26](https://github.com/brekkylab/agent-k/pull/200/commits/e300b26bad1d3ddd4aa16fcd2118df1387c44221)

app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -991,14 +1173,51 @@ function SessionPage() {
...
       unsubscribe();
       if (wsInactivityTimerRef.current) {
         clearTimeout(wsInactivityTimerRef.current);
         wsInactivityTimerRef.current = null;
       }
+      if (revealRafRef.current !== null) {
+        cancelAnimationFrame(revealRafRef.current);
+        revealRafRef.current = null;
+      }
     };
```
ljhh-0611, line 1246: > AI Change Recommendation Review

Tiny race here: this snapshot request can finish after the user has already switched sessions. In that case, the old effect can still feed events into `processEvent` and show the previous session's live run in the new view. Can we ignore late snapshot results after cleanup?

```suggestion
    let ignoreSnapshot = false;
    void getActiveRunSnapshot(sessionId)
      .then((snap) => {
        if (ignoreSnapshot || !snap) return;
        processEvent({
          type: 'agent_run_started',
          session_id: sessionId,
          run_id: snap.run_id,
          user_message: snap.user_message,
        });
        for (const { seq, output } of snap.outputs) {
          processEvent({ type: 'agent_message', session_id: sessionId, run_id: snap.run_id, seq, output });
        }
        if (snap.partial) {
          processEvent({
            type: 'agent_delta',
            session_id: sessionId,
            run_id: snap.run_id,
            delta: snap.partial,
            cum_len: snap.partial.length,
          });
        }
      })
      .catch(() => {
        /* best-effort seed; the WS replay is the fallback */
      });

    return () => {
      ignoreSnapshot = true;
      unsubscribe();
      if (wsInactivityTimerRef.current) {
        clearTimeout(wsInactivityTimerRef.current);
        wsInactivityTimerRef.current = null;
      }
      if (revealRafRef.current !== null) {
        cancelAnimationFrame(revealRafRef.current);
        revealRafRef.current = null;
      }
    };
```

app/src/routes/_app.projects.$projectSlug.sessions.$sessionPrefix.tsx
```diff
@@ -910,6 +1022,73 @@ function SessionPage() {
...
+          // cumStart 0 → the append below trims the overlap), so the reveal
+          // recovers now. Guard against a refetch storm: only one resync in
+          // flight, since every post-hole delta lands here until it completes.
+          if (resyncingRef.current) return;
+          resyncingRef.current = true;
+          // Paint the recovered backlog at once instead of slowly retyping it.
+          revealSeededRef.current = false;
+          void getActiveRunSnapshot(sessionId)
+            .then((snap) => {
+              if (snap?.partial && currentRunIdRef.current === run_id) {
```
ljhh-0611: Small run-id check missing here. This snapshot is for the session's current active run, so if another run starts before the request resolves, we could replay the new run's partial as if it belonged to this older run.

```suggestion
              if (snap?.partial && snap.run_id === run_id && currentRunIdRef.current === run_id) {
```

ljhh-0611 (approved):
Great job!
 
Could you confirm a manual streamed turn on the latest code — ideally a multi-turn with tool calls? Approving on the assumption that manual streaming still behaves as the PR description claims.

jhlee525 (approved):

PR #201, opened 2026-07-02
Workspace per user & Remove project

grf53:
## Summary
'Project' concept is removed, and 'Workspace' concept is reorganized.

Users always have their own default workspace. Here, we are not assuming anything about existence of the workspaces other than users' defaults.

### Workspace kept as its own entity (not collapsed into the user)

- **Chosen: workspace is an independent entity**
  - It has a `workspaces` table in the DB and also exists independently on disk under `workspaces/{wid}/`.
  - "One workspace per user" is enforced in **app logic, not the schema** — so opening it up to multiple-per-user later is a constraint/endpoint change, **not a data migration**.
  - refer to 'After' in 'Scheme change' section

- ~**Alternative: make the workspace fully subordinate to the user** (could be adopted instead)~
  - Drop the `workspaces` table; `agents`/`sessions` reference `user_id` directly instead of `workspace_id`.
  - On disk it would live at `users/{uid}/workspace/` (currently that path is included via a symlink).
  - Workspace metadata, when needed, would be carried as a file — Currently, the workspace's real datum (`title`) is mirrored to a `.title` file(can be `.metadata.json` or sth) **outside** `files/` (backend-only, invisible to clients/agent), which is not really necessary in 'Chosen' design.
  - refer to 'Alternative' in 'Scheme change' section

- **Trade-off**
  - The independent-entity approach carries the cost of one thin table, but avoids structural changes/migrations when expanding to multiple workspaces later — more flexible.
  - The subordinate approach is simpler now (no table), but a later expansion is costly: reassigning `user_id` ↔ `workspace_id` and moving files around.

### Others
- **Workspace has no `user_id`**
  - the user default workspace's id equals the user's id. (by app logic; no schematic restriction exists currently)
  - the decisions on non-default workspaces are not in this.
- **Default workspace auto-provisioned at signup, healed lazily on login, and not user-deletable.**
  Guarantees every active user always has exactly one reachable workspace.

## Schema change

|                        | Before       | After                                                                          |
| ---------------------- | ------------ | ------------------------------------------------------------------------------ |
| Container table        | `projects`   | `workspaces`                                                                   |
| agents/sessions FK     | `project_id` | `workspace_id`                                                                 |
| user ↔ container       | none         | **no DB FK**; default linked only by the `wid == uid` convention (dashed in diagram) |
| cleanup on user delete | (no link)    | no FK → handled in the app (`delete_user_admin` calls `workspaces.remove(id)`) |

**Before — project-based**

<details>
  <summary>Show diagram</summary>

```mermaid
erDiagram
    users {
        TEXT id PK
        TEXT username UK
        TEXT role
    }
    projects {
        TEXT id PK
        TEXT title
        TEXT created_at
        TEXT updated_at
    }
    agents {
        TEXT id PK
        TEXT project_id FK
        TEXT name
        TEXT spec
    }
    sessions {
        TEXT id PK
        TEXT project_id FK
        TEXT agent_id FK
        TEXT spec
    }
    messages {
        TEXT session_id FK
        INTEGER seq
        TEXT content
    }

    projects ||--o{ agents   : "project_id (CASCADE)"
    projects ||--o{ sessions : "project_id (CASCADE)"
    agents   ||--o{ sessions : "agent_id (nullable, CASCADE)"
    sessions ||--o{ messages : "session_id (CASCADE)"
```

</details>

> `users` was an isolated table — no FK to or from any other table, which was kinda 'incomplete'.

**After — with workspace table**

<details>
  <summary>Show diagram</summary>

```mermaid
erDiagram
    users {
        TEXT id PK
        TEXT username UK
        TEXT role
    }
    workspaces {
        TEXT id PK
        TEXT title
        TEXT created_at
        TEXT updated_at
    }
    agents {
        TEXT id PK
        TEXT workspace_id FK
        TEXT name
        TEXT spec
    }
    sessions {
        TEXT id PK
        TEXT workspace_id FK
        TEXT agent_id FK
        TEXT spec
    }
    messages {
        TEXT session_id FK
        INTEGER seq
        TEXT content
    }

    users      ||..|| workspaces : "default: id == id (convention, not a FK)"
    workspaces ||--o{ agents     : "workspace_id (CASCADE)"
    workspaces ||--o{ sessions   : "workspace_id (CASCADE)"
    agents     ||--o{ sessions   : "agent_id (nullable, CASCADE)"
    sessions   ||--o{ messages   : "session_id (CASCADE)"
```

</details>

**Alternative — without workspace table**
User has agents and sessions. Workspace(as filesystem) is only under `users/{uid}/workspace`.

<details>
  <summary>Show diagram</summary>

```mermaid
erDiagram
    users {
        TEXT id PK
        TEXT username UK
        TEXT role
    }
    agents {
        TEXT id PK
        TEXT user_id FK
        TEXT name
        TEXT spec
    }
    sessions {
        TEXT id PK
        TEXT user_id FK
        TEXT agent_id FK
        TEXT spec
    }
    messages {
        TEXT session_id FK
        INTEGER seq
        TEXT content
    }

    users ||--o{ agents     : "user_id (CASCADE)"
    users ||--o{ sessions   : "user_id (CASCADE)"
    agents     ||--o{ sessions   : "agent_id (nullable, CASCADE)"
    sessions   ||--o{ messages   : "session_id (CASCADE)"
```

</details>

## On-disk layout (`data_root/`)

For two users Alice (`uid = a1b2…`) and Bob (`uid = c3d4…`), whose default workspaces have `wid == uid`:

```
data_root/
├── agent-k.db                         # SQLite (here by default: DATABASE_URL=sqlite://./data/agent-k.db)
│
├── workspaces/
│   ├── a1b2.../                       # Alice's default workspace (wid == Alice uid)
│   │   ├── .title                     # "Alice's workspace"  (backend-only, not served over WebDAV)
│   │   ├── files/                     # Alice's file tree, served at /workspaces/a1b2.../files
│   │   ├── notion/                    # Alice's other sources(maybe)
│   │   ├── gdrive/
│   │   └── s3/
│   └── c3d4.../                       # Bob's default workspace
│       ├── .title                     # "Bob's workspace"
│       └── files/
│
└── sessions/
    └── <session_id>/
        └── sandbox.tar.zst            # only for sessions that have a runenv
```

- The workspace's substance lives in `workspaces/{wid}/`: `.title` (title mirror) + `files/` (actual files). WebDAV serves only `files/`, so `.title` is never exposed to clients/agents.
- ~`users/{uid}/workspace` is just a convenience symlink (relative). For a default workspace it points back at `workspaces/{uid}`.~
- Access isolation: a caller (token `sub`) may reach only the workspace whose `id == sub`; any other id returns `404`.

## Ownership / access control

- Single gate `require_owned_workspace`: `wid == auth.id` (else `404`), reused by workspace CRUD, agents, sessions, and messages.
- WebDAV (`/workspaces/{wid}/files`) and the messages WebSocket both authenticate via `?token=` and enforce the same `id == sub` rule.
- Default workspace is auto-provisioned at signup and for the bootstrap admin (`create_default(&user)`), with title `"{username}'s workspace"`.

## API changes

- `/projects` CRUD → `/workspaces` (`GET` list, `GET/PATCH/DELETE /workspaces/{id}`). No `POST` — the single default is created at signup; ownerless multi-create is out of scope.
- WebDAV mount: `/projects/{pid}/workspace[/…]` → `/workspaces/{wid}/files[/…]`.
- `agents` / `sessions` request & response fields: `project_id` → `workspace_id`.

## Migration / ops notes

- **Dev DB reset required**: `0001_initial.sql` was rewritten, so an already-migrated DB will fail the checksum check on startup — delete the existing `agent-k.db` and re-create.
- Old `data_root/projects/` directories are orphaned by the new layout; clean up manually if present.

## Testing

- `cargo test -p agent-k-backend` — 9 passing. Tests added in this change:
  - workspace state: CRUD round-trip, `create_default` (`wid == uid` + `.title` mirror + ~`users/{uid}/workspace` symlink~), and `get_for_user` access rule (owner sees it, other users get `None`, no existence leak).
  - `authenticate` gate: active user passes; deactivated and non-existent subjects are rejected (guards the WebDAV/WS bypass).
  - agent state: CRUD round-trip and `(workspace_id, name)` uniqueness conflict (renamed from the previous project-scoped tests).
- `cargo build` / `cargo clippy` clean — no new warnings (3 pre-existing baseline warnings unrelated to this change).


backend_v2/src/router/session.rs
```diff
@@ -131,10 +137,10 @@ pub(super) async fn create_session(
                 .get(agent_id)
                 .await?
                 .ok_or_else(|| err(StatusCode::NOT_FOUND, "agent not found"))?;
-            if agent.project_id != payload.project_id {
+            if agent.workspace_id != payload.workspace_id {
                 return Err(err(
                     StatusCode::BAD_REQUEST,
-                    "agent does not belong to the given project",
+                    "agent does not belong to the given workspace",
```
nuri-yoo: `require_owned_workspace(payload.workspace_id)` at line 123 has already confirmed the caller owns the target workspace. The agent is then fetched with the *unscoped* `state.agents.get(agent_id)` (line 137), so the two error branches split by whether the id exists anywhere, not by ownership:

- foreign but existing agent id -> `agent.workspace_id != payload.workspace_id` -> `400 "agent does not belong to the given workspace"`
- nonexistent agent id -> `404 "agent not found"`

`GET /agents/{id}` on that same foreign id returns `404` (via `require_owned_agent`), so this endpoint is the one place that reveals cross-tenant existence: `400` means "exists in someone else's workspace", `404` means "does not exist". That contradicts the design note on `require_owned_workspace` ("reported as 404 so it can't be probed").

Reproduced live: bob POSTs `/sessions` with his own `workspace_id` + alice's real agent id -> `400`; same call with a random id -> `404`.

Suggested fix: resolve the agent through `require_owned_agent(&state, &auth, agent_id)` (which 404s for both missing and foreign), then only the active/spec checks remain.

backend_v2/src/router/workspace.rs
```diff
@@ -1,351 +1,174 @@
...
+    state: &AppState,
+    auth: &AuthUser,
+    aid: Uuid,
+) -> Result<Agent, ApiError> {
+    let agent = state
+        .agents
+        .get(aid)
+        .await?
+        .ok_or_else(|| err(StatusCode::NOT_FOUND, "agent not found"))?;
+    require_owned_workspace(state, auth, agent.workspace_id).await?;
```
nuri-yoo: Status codes match (both `404`), but the body text differs by whether the resource exists:

- missing agent -> `404 {"error":"agent not found"}` (line 76)
- foreign but existing agent -> `require_owned_workspace` fails on line 77 -> `404 {"error":"workspace not found"}`

`ErrorBody.error` is serialized into the response (`router/error.rs`), so a client reads the difference. This defeats the "can't be probed" property for every route going through `require_owned_agent`: `GET/PATCH/DELETE /agents/{id}`.

Reproduced live: bob `GET /agents/{alice's real id}` -> `404 {"error":"workspace not found"}`; random id -> `404 {"error":"agent not found"}`.

Suggested fix: return the same message on both legs (e.g. always `"agent not found"`). The WS path in `message.rs` already collapses both to one message, so this only aligns the HTTP helpers with that behavior.

backend_v2/src/router/workspace.rs
```diff
@@ -1,351 +1,174 @@
...
+    state: &AppState,
+    auth: &AuthUser,
+    sid: Uuid,
+) -> Result<Session, ApiError> {
+    let session = state
+        .sessions
+        .get(sid)
+        .await?
+        .ok_or_else(|| err(StatusCode::NOT_FOUND, "session not found"))?;
+    require_owned_workspace(state, auth, session.workspace_id).await?;
```
nuri-yoo: A foreign but existing session yields `404 "workspace not found"` here, while a missing one yields `404 "session not found"` (line 91) — probeable across tenants. Affects everything behind `require_owned_session`: `GET/DELETE /sessions/{id}`, `GET/POST /sessions/{id}/messages`, and the run/stop routes.

Suggested fix: collapse both legs to a single `"session not found"` message (matching the WS handler in `message.rs`, which already does this).

backend_v2/src/router/webdav.rs
```diff
@@ -0,0 +1,356 @@
...
+            return (StatusCode::INTERNAL_SERVER_ERROR, "internal error").into_response();
+        }
+    }
+
+    // The filesystem (and its side-processing) lives in `state.workspaces`;
+    // here we only wrap it in the WebDAV protocol (see [`DavFs`]).
+    let dav = DavHandler::builder()
+        .filesystem(Box::new(DavFs(state.workspaces.get_fs(wid))))
+        .locksystem(FakeLs::new())
+        .strip_prefix(format!("/workspaces/{wid}/files"))
```
nuri-yoo, line 76: `parse_wid` (line 82) uses `Uuid::parse_str`, which accepts uppercase hex and the 32-char no-hyphen "simple" form; both parse to a `Uuid` that compares equal, so auth and `get_for_user` pass. But this `strip_prefix` is built from `format!("/workspaces/{wid}/files")`, and `{wid}` renders via `Display` as the canonical lowercase-hyphenated form. dav-server does a byte-exact prefix match, so a non-canonical request path never matches -> `PrefixMismatch` -> `502 BAD_GATEWAY`.

Reproduced live (owner's own workspace, valid token):
- lowercase `PROPFIND /workspaces/15fc016b-.../files` -> `207`
- same value uppercased `.../15FC016B-.../files` -> `502`
- 32-hex no-hyphen form -> `502`

Pre-existing shape (old `parse_pid` + `/projects/{pid}/workspace` did the same), but this file was rewritten in the PR, so worth fixing now. Suggested fix: derive the strip prefix from the actual request path (up to and including `/files`) rather than re-formatting the parsed uuid; or normalize `wid.to_string()` and reject a non-canonical segment up front.

backend_v2/src/router/workspace.rs
```diff
@@ -1,351 +1,174 @@
...
+    require_owned_workspace(&state, &auth, id).await?;
+    // The default workspace (id == user id) is not user-deletable; it is only
+    // removed when the account itself is deleted.
+    if id == auth.id {
+        return Err(err(
+            StatusCode::FORBIDDEN,
+            "cannot delete your default workspace",
+        ));
+    }
+    state.workspaces.remove(id).await?;
```
nuri-yoo: `require_owned_workspace` at line 139 only returns `Ok` when `id == auth.id` (for any other id `get_for_user` returns `None` -> `404`). The guard on line 142 then rejects exactly `id == auth.id` with `403 "cannot delete your default workspace"`. So the case analysis is: `id != auth.id` -> `404`; `id == auth.id` -> `403`; the row-missing subcase -> `404`. Line 148 is dead.

Reproduced live: own wid -> `403`; another user's wid -> `404`; random -> `404`. No admin bypass (this route sits under plain `auth_required`, and neither the handler nor `get_for_user` reads `role`). `WorkspacesState::remove` has no other production caller.

The route is still registered with `.delete(...)`, so it appears in the OpenAPI surface and clients generate a delete op that always errors. Suggested fix: drop the `.delete` on `/workspaces/{id}` (axum will answer `405`) until non-default workspaces exist — the branch only becomes reachable then, which is when it should be added.

backend_v2/src/state/workspace/mod.rs
```diff
@@ -1,39 +1,340 @@
...
+        if tokio::fs::symlink_metadata(&link).await.is_ok() {
+            tokio::fs::remove_file(&link).await.ok();
+        }
+        // From `users/{uid}/` the workspace dir is `../../workspaces/{uid}`.
+        let target = PathBuf::from("..")
+            .join("..")
+            .join("workspaces")
+            .join(user_id.to_string());
+        #[cfg(unix)]
+        tokio::fs::symlink(&target, &link).await?;
```
nuri-yoo: This is `symlink_metadata` check -> `remove_file` (`.ok()`'d) -> `symlink` (`?`'d), with no atomicity. Two concurrent `create_default` calls for the same user both observe no existing link, so the loser's `symlink` fails with `EEXIST`, which propagates as `StateError::Io` -> `500`. The trigger is narrow but real: the login heal (`auth.rs:125`) runs `create_default` only when the workspace *row* is missing, so two simultaneous logins in that state (partial signup, wiped workspace) race here.

Reproduced live: with the workspace row deleted but the user kept, 8 concurrent logins -> 2 returned `500`, and the server log shows `state error: io: File exists (os error 17)` — valid credentials, failed login.

Suggested fix: treat `EEXIST` from `symlink` as success (the intent is idempotent replacement), or create the link at a temp name and `rename` it into place.

backend_v2/src/router/user.rs
```diff
@@ -283,11 +283,16 @@ pub(super) async fn delete_user_admin(
...
-    let deleted = state.users.delete(id).await?;
-    if !deleted {
+    if state.users.get(id).await?.is_none() {
         return Err(StateError::NotFound.into());
     }
 
+    // Remove the default workspace's files first; only if that succeeds do we
+    // delete the user and workspace rows atomically. A filesystem failure aborts
+    // before any row is removed, and the two row deletes never diverge.
+    state.workspaces.remove_files(id).await?;
```
nuri-yoo: Two issues on this path. First, non-atomic across DB and filesystem, and resurrectable: `remove_files` wipes `data_root/workspaces/{uid}` (and the symlink) first; `delete_with_default_workspace` then deletes the rows in a separate transaction. If that tx fails (e.g. `SQLITE_BUSY`), the request 500s with the user still active but their files gone — and the login heal won't re-provision, because it checks the workspace *row* (`auth.rs:125`), which still exists. Separately, `workspaces` has no FK to `users`, so a victim's in-flight `PATCH /me/workspace` that committed just after the row delete re-inserts an orphan row + directories that no API can remove (the user is gone -> `delete_user_admin` 404s; `DELETE /workspaces/{id}` can't succeed). An admin retry does recover the first case (`remove_files` is idempotent), so severity is bounded.

Second, session cleanup is bypassed: agent/session/message rows cascade via FK, but `SessionsState::remove` — the only code that cancels an in-flight run, deletes `data_root/sessions/{sid}/`, and closes the event channel — is never called on this path. So: a running session keeps executing until its next message INSERT hits the FK, an attached WS client never wakes with `RecvError::Closed`, and the `sandbox.tar.zst` archive leaks.

Proven with a real microsandbox VM: booted a sandbox, archived a runenv session (2,078,142-byte `sandbox.tar.zst` on disk), then ran this exact delete path — the session row cascaded away but the archive remained and `SessionsState::remove` now returns `NotFound`, so it's unreclaimable via the API.

Minor: the `users.get(id)` pre-check on line 286 duplicates the bool `delete_with_default_workspace` already returns for exactly this purpose (and adds a TOCTOU gap).

Suggested direction: enumerate and tear down the workspace's sessions through `SessionsState::remove` (cancel + dir + channel) before/within the delete, and either wrap files+rows so a failure can't leave them diverged or make the heal detect a missing files root.

backend_v2/src/router/agent.rs
```diff
@@ -90,11 +97,15 @@ pub struct UpdateAgentRequest {
...
 ) -> Result<Json<AgentListResponse>, ApiError> {
-    let agents = match query.project_id {
-        Some(pid) => state.agents.list_by_project(pid).await?,
-        None => state.agents.list().await?,
+    let agents = match query.workspace_id {
+        Some(wid) => {
+            require_owned_workspace(&state, &auth, wid).await?;
+            state.agents.list_by_workspace(wid).await?
+        }
+        None => state.agents.list_by_workspace(auth.id).await?,
```
nuri-yoo, line 108: `auth.id` is passed straight in as a workspace id, re-encoding the default-workspace identity outside `get_for_user`, which its own doc calls "the single definition of the workspace access rule". It also under-delivers the query-param doc right above ("Omit to list every agent across all of the caller's workspaces") — this only scans the default workspace, correct solely while one-per-user holds.

The same convention is repeated at `session.rs` list (`list_by_workspace(auth.id)`), `workspace.rs` `list_workspaces` (`get(auth.id)` shaped as a list), the `delete_workspace` default guard (`id == auth.id`), and the login heal. When a second workspace arrives — the PR's stated direction — updating `get_for_user` fixes none of these; each silently returns the wrong scope (empty/partial) with no compile error, and `workspaces` has no owner column, so a real list-by-owner can't even be written without the schema change the design says it avoids.

Not a bug today (the identity holds), so this is design-debt, not correctness. Consolidating through a `default_workspace_id(user)` / `list_for_user(user)` helper beside `get_for_user` (and an owner column when multi-workspace lands) would remove the scattering.

backend_v2/src/state/workspace/mod.rs
```diff
@@ -1,39 +1,340 @@
...
+        .bind(&item.title)
+        .bind(item.created_at.to_rfc3339())
+        .bind(item.updated_at.to_rfc3339())
+        .execute(&self.db)
+        .await?;
+
+        // Provision the file root and mirror the title to `.title` on disk, so
+        // the workspace is fully described without a DB read.
+        tokio::fs::create_dir_all(self.get_root(id)).await?;
+        tokio::fs::write(self.title_path(id), &item.title).await?;
```
nuri-yoo: It's written on every `upsert` — signup, login heal, and every title `PATCH` — but no production code reads it back (grep finds reads only in this module's tests). The doc says the mirror makes the workspace "fully described without a DB read", but nothing performs that read.

It also adds a fallible fs write *after* the DB row is already updated, so on `PATCH /workspaces/{id}` an fs error (perms, disk full) turns a committed title change into a `500` with the DB row updated and the mirror stale/missing. The PR body itself notes the mirror "is not really necessary in 'Chosen' design".

Suggested fix: drop `title_path()`, the write, and the two test assertions; the `workspaces.title` column is the single source of truth.

backend_v2/src/state/workspace/mod.rs
```diff
@@ -1,39 +1,340 @@
...
+            .join("users")
+            .join(user_id.to_string())
             .join("workspace")
     }
+
+    /// (Re)create the `users/{uid}/workspace` → `workspaces/{uid}` symlink. The
+    /// default workspace's id equals the user's id, so the target is
+    /// `workspaces/{uid}`. The link is *relative* so the whole `data_root` can
+    /// be relocated without breaking it.
+    async fn link_user_default(&self, user_id: Uuid) -> StateResult<()> {
```
nuri-yoo: Created here (via `create_default`) and removed in `remove_files`, but no code path ever resolves or traverses it — WebDAV goes straight to `data_root/workspaces/{wid}/files` (`get_root`), and sessions use `data_root/sessions`. Confirmed live: after signup the link exists (`workspace -> ../../workspaces/{uid}`) but nothing reads it.

It's ~45 lines total (`link_user_default` with its idempotent-replace + `cfg(not(unix))` warn arm, `user_default_link`, the removal branch, tests) plus a whole `data_root/users/` tree that exists only to hold the link — and it's the source of the `EEXIST` race flagged on `link_user_default`. Consider introducing it in the PR that actually consumes a user-centric path.

backend_v2/src/router/workspace.rs
```diff
@@ -1,351 +1,174 @@
...
-    fn metadata(&self) -> FsFuture<'_, Box<dyn DavMetaData>> {
-        let meta = self
-            .0
-            .metadata()
-            .map(|m| Box::new(DavMetaAdapter(m)) as Box<dyn DavMetaData>)
-            .map_err(to_dav_err);
-        Box::pin(ready(meta))
-    }
+/// `PATCH /me/workspace` — update the caller's default workspace.
+pub(super) async fn update_my_workspace(
```
nuri-yoo, line 190: `update_my_workspace` (and `get_my_workspace` above) are line-for-line copies of `update_workspace`/`get_workspace` with `auth.id` substituted for `Path(id)`. The two PATCH bodies must be kept in sync by hand — a later change (title validation, event emit, mirroring behavior) applied to one silently skips the other.

Suggested fix: make the `/me` handlers one-line delegates into the generic handlers with `Path(auth.id)`; or drop the `/me/workspace` pair entirely (clients call `GET /workspaces` and pick — a singular "my workspace" alias also breaks the moment a user has more than one).

backend_v2/src/router/message.rs
```diff
@@ -108,11 +109,26 @@ pub(super) async fn stream_messages(
...
+    let session = state
+        .sessions
+        .get(sid)
+        .await?
+        .ok_or_else(|| err(StatusCode::NOT_FOUND, "session not found"))?;
+    // Same access gate as the HTTP routes: the session's workspace must be one
+    // the caller can reach.
+    if state
+        .workspaces
+        .get_for_user(user.id, session.workspace_id)
```
nuri-yoo: Lines ~117-133 inline `sessions.get -> 404 -> get_for_user -> 404`, duplicating `require_owned_session`, which is already imported in this file and used by `list_messages`/`start_run`/`stop_run`. After `authenticate` returns an `AuthUser`, `require_owned_session(&state, &user, sid).await?` is a drop-in.

Important caveat: the divergence is currently *beneficial*. This path returns `"session not found"` on both the missing-session and foreign-workspace legs, so it does not leak existence via message text — whereas `require_owned_session` does (see that comment). So don't dedupe naively: harden the helper first (single message on both legs), then have this handler call it. Flagging so the two don't silently drift.

backend_v2/src/state/workspace/fs.rs
```diff
@@ -190,31 +190,31 @@ impl File {
...
-    pub(super) fn new(root: PathBuf, pid: Uuid) -> Self {
-        Self { root, pid }
+    pub(super) fn new(root: PathBuf, wid: Uuid) -> Self {
+        Self { root, wid }
     }
 
     /// Absolute on-disk path of `rel_path` (a workspace-relative path such as
-    /// `/knowledge/foo.txt`) inside this project's workspace.
+    /// `/knowledge/foo.txt`) inside this workspace.
     fn resolve(&self, rel_path: &str) -> PathBuf {
```
selloriwoo, line 218: `resolve` joins `rel_path` onto `self.root` without any containment check:

```rust
    self.root.join(rel_path.trim_start_matches('/'))
}
```

`join` doesn't normalize, so a `..` in `rel_path` escapes the workspace root. e.g `/../<other_wid>/files/...` or `/../../../etc/passwd` reaches other tenants or the host FS.

Not exploitable today: the only caller is WebDAV, and `dav-server`'s `DavPath` already strips `..`. But this module bills itself as the reusable entry point for *all* fs ops, so the guard shouldn't live only in the protocol layer. A future non-WebDAV caller (e.g. `knowledge/` ingestion) passing an untrusted path would make this a live traversal.

Suggested: enforce containment here, lexically

```rust
use std::path::Component;

fn resolve(&self, rel_path: &str) -> FsResult<PathBuf> {
    let mut out = self.root.clone();
    for comp in Path::new(rel_path.trim_start_matches('/')).components() {
        match comp {
            Component::Normal(c) => out.push(c),
            Component::CurDir => {}
            _ => return Err(FsError::Forbidden), // reject `..`, absolute, prefix
        }
    }
    Ok(out)
}
```

backend_v2/src/state/workspace/mod.rs
grf53: Codes related to symlinks doesn't exist, so let me mark this resolved.

backend_v2/src/state/workspace/mod.rs
grf53: Codes related to symlinks doesn't exist, so let me mark this resolved.

backend_v2/src/router/agent.rs
```diff
@@ -159,8 +171,10 @@ pub(super) async fn update_agent(
 
 pub(super) async fn delete_agent(
     State(state): State<Arc<AppState>>,
+    Extension(auth): Extension<AuthUser>,
     Path(id): Path<Uuid>,
 ) -> Result<StatusCode, ApiError> {
+    require_owned_agent(&state, &auth, id).await?;
     state.agents.remove(id).await?;
```
selloriwoo: `delete_agent` removes the agent row, which cascades to its sessions via `agent_id ... ON DELETE CASCADE`. But the cascade is pure SQL — it never calls `SessionsState::remove`, the only code that cancels an in-flight run, deletes `data_root/sessions/{sid}/`, and closes the event channel. Same gap flagged on `delete_user_admin`, just via a more common route.

So deleting an agent that has a live or archived session leaves:
- the run task executing until its next message INSERT hits the FK and errors out
- the `sandbox.tar.zst` archive orphaned on disk — unreclaimable, since the session row is already gone
- attached WS clients never woken with `RecvError::Closed`

Suggested: enumerate the agent's sessions and tear each down through `SessionsState::remove` (cancel + dir + channel) before deleting the agent row — same direction as the `delete_user_admin` fix. More broadly, every cascade-based delete path (user / agent / workspace) needs an app-level teardown hook; the FK alone can't reach the in-memory run map, the sandbox archives, or the event channels.

backend_v2/src/router/workspace.rs
grf53: `DELETE /workspaces/{id}` endpoint is commented out, and the handler is marked as 'dead code'.

selloriwoo (changes requested):

backend_v2/src/state/workspace/fs.rs
selloriwoo: I accidentally clicked it, so it says “requested,” but I don't think you need to worry about it.

backend_v2/src/router/webdav.rs
```diff
@@ -0,0 +1,356 @@
...
+    Router::new().fallback(handle).with_state(state)
+}
+
+async fn handle(State(state): State<Arc<AppState>>, req: Request) -> Response<Body> {
+    let wid = match parse_wid(req.uri().path()) {
+        Some(w) => w,
+        None => return (StatusCode::BAD_REQUEST, "invalid workspace id").into_response(),
+    };
+
+    let token = req.uri().query().and_then(extract_token);
```
selloriwoo, line 51: WebDAV auth takes the JWT from the URL (`?token=`), which lands in server and proxy access logs (and browser history if navigated to). Anyone with log access then has a valid token until it expires. And this is the full session JWT, so a leak is account-wide, not just file access — CWE-598/532. (No CSRF concern though — it's an explicit token, not a cookie.)

Suggested: prefer an `Authorization: Bearer` header (the raw `Request` is available here, so it's the same read `auth_required` already does) and keep `?token=` only as a fallback; at minimum mask the `token` param in logging. If URL-based access needs to stay for ergonomics, issue a short-lived, workspace-scoped token for it rather than the session JWT.

The WS route can keep its query token — browser WebSocket can't set headers.

backend_v2/src/router/user.rs
grf53: ### Treatments
- Session Leak: Cancel run, delete folder, and close channel (`discard_artifacts`) for each session upon user deletion
- Incomplete Recovery: Change order from deleting rows first to deleting files later (data remains intact even if it fails)
- Orphan Reinsertion: Foreign keys already prevent this → No action taken
- Duplicate Check: Remove dictionary `users.get`, replace deletion return with a boolean

backend_v2/src/router/agent.rs
grf53: Ownership access is already unified in `get_for_user`, and the owner column already exists.
What's left is just the scattered "list scope / default-workspace identity" convention.
It's correct today and only breaks once a user has multiple workspaces.
So we're leaving it as accepted design-debt, to be paid alongside that future work.

backend_v2/src/state/workspace/mod.rs
grf53: `.title` was like a demo of the alternative design(filesystem-as-truth). Now it has no meaning, and it can cause problems. So it will be dropped.

nuri-yoo (approved):
LGTM

backend_v2/src/router/agent.rs
grf53: I mistakenly assumed that the considerations for user deletion were naturally applied to agent deletion as well. To ensure this happens, I implemented a unified deletion logic for each component.

backend_v2/src/router/webdav.rs
grf53: I added no modification on that way to set access token, and there is an explanation for that in the comment.
I don't have a strong opinion for this and it is not really in the scope of this PR. So I just suggest you to make this a separated Issue/PR.
Let me know your opinion, please.

backend_v2/src/state/workspace/fs.rs
grf53: I considered `resolve` as something like a helper function to find the absolute path for a relpath, so I am not sure it is a best place to put that kind of refusal. And also your suggestion blocks too much: any relpath including `..` will be refused.

ljhh-0611 (approved):
There is an follow-up PR for future work (not based on this PR)

Deleting a session (or via account/workspace/agent cascade) while its run is in flight leaks an orphan `sessions/{id}/`: `remove`/`discard_artifacts` cancel-then-`remove_dir_all` without awaiting the run task, whose post-cancel `sandbox.archive` re-creates the dir afterwards (never reclaimed). Root cause pre-exists on `backend_v2` — this PR only extends it to the new cascade paths, so it's a follow-up, not a blocker.

backend_v2/src/state/workspace/fs.rs
selloriwoo: Fair. and scratch my snippet, blanket-rejecting `..` was wrong (`/a/b/../c` is legit).

No live exposure today since DavPath already strips `..`, so happy to keep this out of the PR — just worth a follow-up if a non-WebDAV caller ever feeds resolve untrusted paths. Resolving the thread.

backend_v2/src/router/webdav.rs
selloriwoo: Agreed. the `?token=` transport is carried over, not introduced here, and a real fix (header + frontend change, or a scoped short-lived token) is bigger than this PR's scope. Makes sense to track it as a separate issue rather than fold it in here. Thanks — resolving the thread.

selloriwoo (approved):

PR #203, opened 2026-07-10
Backend v2

grf53:
## Description
Getting backend v2 into main branch

## Suggestion(Not yet applied)
- `backend` -> `backend-v1`
- `backend-v2` -> `backend`


ljhh-0611 (approved):
!Let's F Go!

agent-k/src/knowledge_base/helper.rs
```diff
@@ -73,59 +72,66 @@ pub(super) trait HelperAgent {
...
-                        }
+        let spec = AgentSpec::new(model).instruction(Self::INSTRUCTION);
+        let mut agent = Agent::try_new(spec)?;
+        let query = Self::build_query(&input);
+
+        let mut text_parts: Vec<String> = Vec::new();
+        {
+            let mut stream = agent.run(query);
+            while let Some(result) = stream.next().await {
+                let output = result?;
```
nuri-yoo: **This breaks the metadata-fallback contract (`generate()` regression)**

`result?` here propagates a call/stream error straight up, so a transient LLM error now aborts ingest.

On base (origin/main), `generate()` tried each model in `MODELS` order, advanced to the next on a call/stream error, and returned `Ok(Self::fallback(...))` even if all of them failed. The documented invariant was "a failed metadata call never blocks ingest."

In this PR, `generate()` picks the first registered model and makes a single call, propagating any stream error via this `result?`. Now, if a provider is registered and that call errors (429 / timeout / network), `generate()` returns `Err`. Fallback only remains for the `process()`-returns-`None` and no-provider-registered cases.

Effect at the call sites (`knowledge_base/mod.rs`):

- single `add`: the `?` on `tokio::try_join!(get_title, get_purpose)?` propagates, failing ingest of that document entirely.
- batch `add_many`: the item is pushed to `failed`, and when `new_corpus` the corpus artifact is removed, so the document is dropped from the index.

Previously both paths indexed silently with fallback metadata ("Untitled" / ""). Combined with ailoy's 429 retry (which doesn't distinguish permanent from transient and fails after a wasted backoff), this makes ingest failures more likely under rate limiting.

Provenance: this change originates in `origin/revise` at `c149982` ("WIP"), and this PR's `725e3367` ("Apply revise ... onto origin/main") pulled helper.rs in wholesale. `725e3367` is not an ancestor of `origin/main`, so merging this PR is what first lands the regression on main.

Either restore the original behavior (advance to the next model on error, fall back only when all fail), or — if the single call is intentional — absorb a call/stream `Err` into `Ok(fallback)` after a warn so the "never blocks ingest" contract holds. If the simplification is intended, the doc comments on `generate()` and its callers should be updated to match.

grf53: Cuz I wrote the existing codes, so I restored the fallback that you pointed with my own judgement.
Thank you for pointing this.

nuri-yoo (approved):
LGTM

PR #204, opened 2026-07-10
External source connection

grf53:
## Description

> [!NOTE]
> Most of the core features were taken from https://github.com/brekkylab/ailoy/pull/432.

Mount **external sources (S3/Notion)** into a workspace and expose them to both the **browser (WebDAV)** and the **sandboxed agent (in-guest FUSE)**. Local files and external sources converge into a single unified tree (`/mnt/workspace`), so what the browser sees over WebDAV and what the guest agent sees as a filesystem are the same view.

Introduces the new `backend_v2/src/vfs/` subsystem and the guest-side `crates/ailoy-vfs-forwarder` crate.

## Changes

**External source mounting (data layer)**
- Vendored ailoy's frontend-agnostic VFS core (`61c4c43`) into `backend_v2/src/vfs` — read-only S3+Notion with a typed `VfsError`.
- Turned `WorkspaceFs` into a router: top-level mount prefixes resolve to the provider, everything else stays on local disk. Abstracted metadata into a local `Stat` so local + external share one path.
- Mounts persist per workspace in a new `workspace_mounts` table (migration `0002`), with a credential-masked CRUD API (`GET/POST /workspaces/{wid}/mounts`, `DELETE .../{id}`).

**Sandboxed-agent exposure (in-guest FUSE)**
- Host side: `VfsForward`, an HTTP/1.1 forward server. Guest side: a dependency-free static FUSE forwarder (`crates/ailoy-vfs-forwarder`), embedded via `include_bytes!`.
- runenv sessions restore into a host-egress sandbox so the guest reaches the host at `host.microsandbox.internal`.
- Generalized the forward server over a `ForwardFs` trait and mounted the unified `WorkspaceFs` at `/mnt/workspace` (`files/` local + provider mounts as sibling dirs). Removed the provider-only `/mnt/vfs` view and the `/workspace/files` bind mount after unification, so the guest view is a single tree.

**Write support**
- Implemented writes on the unified mount — local `files/` go to disk (so knowledge ingestion still fires); providers are capability-gated (S3 writable, Notion read-only, cross local/provider rename rejected).

**Metadata fidelity**
- Propagate atime/ctime end-to-end (FileStat/FwdStat/JSON/cache) so `ls -l` shows real timestamps. 

| Source           | mtime              | atime         | ctime          |
| ---------------- | ------------------ | ------------- | -------------- |
| **Local** (disk) | real mtime         | real atime    | real ctime     |
| **S3**           | `LastModified`     | — → **mtime** | — → **mtime**  |
| **Notion**       | `last_edited_time` | — → **mtime** | `created_time` |

**Dependencies / ailoy pin**
- Swapped the temporary `[patch] path="../ailoy"` for a pinned git rev.
- Migrated the boolean host-egress flag to the graded `SandboxNetwork` enum API (`Public`).
- Added `object_store` / `reqwest` / `futures`.

## Testing

- `cargo build` + 28 unit tests pass (verified across the commit history).
- In-VM e2e: reading a local file and real Notion under the unified `/mnt/workspace`, plus an archive + host-egress roundtrip.

## Notes / Follow-ups

- ⚠️ **Provider config is stored as plaintext JSON.** Credential encryption is a follow-up.
- Notion is read-only (no write support yet). Follow-up if needed.

## Related PR
- https://github.com/brekkylab/ailoy/pull/436: These changes are written based on this network egress feature in ailoy.


backend/src/vfs/resource/notion.rs
```diff
@@ -0,0 +1,601 @@
...
+    async fn stat(&self, path: &VPath) -> VfsResult<FileStat> {
+        let segs = segments(path);
+        match segs.as_slice() {
+            [] | [_] => Ok(FileStat {
+                kind: FileKind::Dir,
+                ..Default::default()
+            }),
+            [p, rest @ ..] if p == "pages" && !rest.is_empty() => {
+                if rest.last().map(String::as_str) == Some("page.json") {
+                    let id = page_id(&rest[rest.len() - 2]);
```
nuri-yoo: `stat` on `/pages/page.json` (a `pages/` path with no page-dir segment) matches this arm with `rest = ["page.json"]`, then `rest.last() == "page.json"` is true and `rest[rest.len() - 2]` evaluates `rest[1 - 2]` — a `usize` underflow. Debug panics on overflow, release wraps to `usize::MAX` and panics on the out-of-bounds index; I reproduced both. `read_bytes` guards this with `segs.len() >= 3` and `readdir` returns `NotFound` for a trailing `page.json`, but `stat` doesn't. It's reachable from a cold or TTL-expired cache (the negative-cache short-circuit needs `/pages` freshly listed), so a `PROPFIND` on `…/pages/page.json` or a guest `stat` panics the request task.

Requiring the page-dir segment before indexing closes it; the single-segment and short cases already fall through to `NotFound`:

```suggestion
            [p, rest @ ..] if p == "pages" && rest.len() >= 2 => {
                if rest.last().map(String::as_str) == Some("page.json") {
                    let id = page_id(&rest[rest.len() - 2]);
```

<details>
<summary>Regression test (no network — exercises the guarded match on <code>segments</code>)</summary>

```rust
// in `mod tests` of notion.rs
#[test]
fn stat_of_pages_page_json_does_not_underflow() {
    // `/pages/page.json` has no page-dir segment; the arm must not index
    // rest[rest.len() - 2]. This mirrors the guarded stat() match.
    for path in ["/pages/page.json", "/pages"] {
        let segs = segments(&VPath::new(path));
        if let [p, rest @ ..] = segs.as_slice() {
            if p == "pages" && rest.len() >= 2
                && rest.last().map(String::as_str) == Some("page.json")
            {
                let _ = page_id(&rest[rest.len() - 2]); // only reached when safe
            }
        }
        // reaching here without a panic is the assertion
    }
}
```
</details>

crates/ailoy-vfs-forwarder/src/main.rs
```diff
@@ -0,0 +1,940 @@
...
+            &format!("path={}", pct(&path)),
+            Some(&merged),
+        ) {
+            // C4: a `.cmd/<op>` write returns the command result — cache it so a
+            // read of that path returns it (e.g. the new page id).
+            if path.contains("/.cmd/") {
+                self.cmd_results.lock().unwrap().insert(path, body);
+            }
+        }
+        true
```
nuri-yoo: `put()` returns `bool`, and `flush`/`release` turn it into `reply.ok()` vs `reply.error(EIO)`. This block only enters on `Ok((200, _))` and then falls through to `true` unconditionally, so a PUT that returns `Err` (a transient failure at close time) or `Ok((500, _))` (a host-side `write_bytes` that rejects, e.g. a read-only provider) still reports success — the guest's `close()` returns 0 while nothing was written. The R1 guard only covers the preload GET, not the write itself.

Propagating the PUT outcome makes `flush`/`release` surface `EIO` on failure:

```suggestion
        match http(
            "PUT",
            "/write",
            &format!("path={}", pct(&path)),
            Some(&merged),
        ) {
            Ok((200, body)) => {
                // C4: a `.cmd/<op>` write returns the command result — cache it so a
                // read of that path returns it (e.g. the new page id).
                if path.contains("/.cmd/") {
                    self.cmd_results.lock().unwrap().insert(path, body);
                }
                true
            }
            // A non-200 or transport error must not read back as a successful
            // write — the caller maps `false` to EIO.
            _ => false,
        }
    }
```

workspace/src/vfs/sandbox/forward.rs
```diff
@@ -0,0 +1,313 @@
...
+) -> anyhow::Result<()> {
+    let req = read_request(&mut stream).await?;
+
+    if req.token.as_deref() != Some(token.as_str()) {
+        return respond(&mut stream, 403, "text/plain", b"forbidden".to_vec()).await;
+    }
+
+    let params = parse_query(&req.query);
+    let path = params
+        .get("path")
```
nuri-yoo, line 165: `path` flows straight into `WorkspaceFs::{read,write,…}` → `guest_local_rel` → `resolve()`, which is `self.root.join(rel.trim_start_matches('/'))`. `PathBuf::join` doesn't collapse `..` (verified), so `path=/files/../../../etc/passwd` resolves outside the workspace at the syscall level. WebDAV is safe because `dav_server::DavPath` rejects `..`, but this server bypasses `dav_server` and has no equivalent guard.

Access through the FUSE mount is fine — the guest kernel resolves `..` before it reaches the forwarder. The gap is a direct HTTP request to this server: a shell in the guest can read `VFS_TOKEN` from the forwarder's `/proc/<pid>/environ` and reach `host.microsandbox.internal:<port>`, so the token isn't a boundary against the sandboxed agent itself. Rejecting `..`/`.` components at this edge (and in `resolve()` as defense in depth) closes it. The S3/Notion side is already safe — `object_store` encodes `..` to `%2E%2E` — so this is specifically the local `files/` subtree.

<details>
<summary>Test — currently fails, should pass after the fix</summary>

```rust
// in `mod tests` of state/workspace/fs.rs
#[tokio::test]
async fn forward_fs_rejects_parent_traversal() {
    use crate::vfs::sandbox::ForwardFs;
    let tmp = tempfile::tempdir().unwrap();
    std::fs::write(tmp.path().join("secret"), b"top-secret").unwrap(); // sibling above root
    let files = tmp.path().join("ws");
    std::fs::create_dir_all(&files).unwrap();
    let fs = WorkspaceFs::new(files, Uuid::new_v4());

    let r = ForwardFs::read(&fs, "/files/../secret", None, None).await;
    assert!(r.is_err(), "traversal escaped the workspace root");
    assert!(!ForwardFs::stat(&fs, "/files/../secret").await.unwrap().exists);
}
```
</details>

workspace/src/vfs/accessor/notion.rs
```diff
@@ -0,0 +1,173 @@
...
+    pub async fn create_page(&self, body: Value) -> anyhow::Result<Value> {
+        self.send(self.client.post(format!("{API}/pages")).json(&body))
+            .await
+    }
+
+    pub async fn append_blocks(&self, block_id: &str, children: Value) -> anyhow::Result<Value> {
+        let body = json!({ "children": children });
+        self.send(
+            self.client
+                .patch(format!("{API}/blocks/{block_id}/children"))
```
nuri-yoo, line 219: `format!("{API}/blocks/{block_id}/children")` is parsed by reqwest via the `url` crate, which resolves dot-segments and honors `?`/`#`. With `block_id = "../pages/<id>?"` the effective request becomes `PATCH …/v1/pages/<id>` — arbitrary Notion PATCH endpoints (archive page, update properties/database) reachable with the mount token. I confirmed the rewrite with the `url` crate directly. `page_id()` → `get_page`/`list_children` have the same shape.

This isn't reachable in the current PR — I traced it: the host never dispatches a `.cmd/` write to `Resource::command` (writes hit `write_bytes`, `Unsupported` for Notion), so `append_blocks` is dead for now. But the guest forwarder's `.cmd` result caching and the provider prompt already anticipate wiring it. Validating ids against the Notion UUID shape before building the URL keeps this from going live the moment that dispatch lands — worth doing in this PR while the code is fresh.

<details>
<summary>Test — id validation rejects traversal-shaped block ids</summary>

```rust
// assumes a helper like `fn valid_notion_id(s: &str) -> bool`
// (32 hex chars, optionally hyphenated) added alongside append_blocks.
#[test]
fn notion_id_validation_rejects_url_escapes() {
    assert!(valid_notion_id("22222222222222222222222222222222"));
    assert!(valid_notion_id("22222222-2222-2222-2222-222222222222"));
    for bad in ["../pages/2222?", "..%2Fpages", "abc/def", "x?y", "x#y", ""] {
        assert!(!valid_notion_id(bad), "should reject {bad:?}");
    }
}
```
</details>

backend/src/vfs/resource/s3.rs
```diff
@@ -0,0 +1,204 @@
...
+
+impl S3Resource {
+    pub fn new(config: &S3Config) -> anyhow::Result<Self> {
+        Ok(Self {
+            accessor: S3Accessor::new(config)?,
+        })
+    }
+
+    fn os_path(&self, path: &VPath) -> OsPath {
+        OsPath::from(self.accessor.key(path))
```
nuri-yoo: `OsPath::from(key)` percent-encodes `%`, `#`, `?`, `[`, `]`, `~` (and `..` → `%2E%2E`) in `object_store` 0.14 — I verified `Path::from("50%off.txt")` → `50%25off.txt`. `readdir` surfaces the raw key from `obj.location.filename()`, so a real object `50%off.txt` shows up in `ls`, but a following `stat`/`read`/`unlink` addresses `50%25off.txt` — a different object → `NotFound`, and `write_bytes` silently creates a differently-named object. Round-trip is broken for any key with those characters.

One caveat before changing this: the same encoding is what neutralizes `..` on the provider side (it can't escape `key_prefix`). If you move to `OsPath::parse` for fidelity, keep an explicit `.`/`..` component rejection.

<details>
<summary>Test — currently fails, demonstrates the divergence</summary>

```rust
// new `#[cfg(test)] mod tests` in s3.rs (os_path/key must be reachable from here)
#[cfg(test)]
mod tests {
    use super::*;
    use crate::vfs::accessor::S3Config;

    fn cfg() -> S3Config {
        S3Config {
            bucket: "b".into(), region: "us-east-1".into(),
            access_key_id: "k".into(), secret_access_key: "s".into(),
            endpoint: None, key_prefix: None,
        }
    }

    #[test]
    fn special_char_key_round_trips() {
        let r = S3Resource::new(&cfg()).unwrap();
        for name in ["50%off.txt", "a#b.txt", "note~1"] {
            let vp = VPath::new(&format!("/{name}"));
            // The key readdir would surface vs. the key stat/read address.
            assert_eq!(
                r.os_path(&vp).as_ref(),
                r.accessor.key(&vp),
                "round-trip broken for {name:?}"
            );
        }
    }
}
```
</details>

workspace/src/vfs/cache.rs
```diff
@@ -0,0 +1,391 @@
...
+    async fn readdir(&self, path: &VPath) -> VfsResult<Vec<DirEntry>> {
+        if let Some(entries) = self.cache.list_dir_entries(path.as_str()) {
+            return Ok(entries);
+        }
+        let entries = self.inner.readdir(path).await?;
+        self.cache.set_dir(path.as_str(), &entries);
+        Ok(entries)
+    }
+
+    async fn stat(&self, path: &VPath) -> VfsResult<FileStat> {
```
nuri-yoo: Three interacting issues let a deleted object keep stat'ing as existing until process restart. `get()` (line 63) reads `entries` with no TTL check — only `is_listed`/`list_dir_entries` honor `expiry` — so a `Some(e)` here is served even after its listing expired. The `Some(e)` fast-path also returns before the negative-cache check in the `None` arm, so a fresh parent listing proving non-existence never overrides a stale positive. And `set_dir` inserts new children without removing entries for children that vanished, while `invalidate_dir` drops only direct children, so a subtree `rmdir`/`rename` leaves grandchild entries behind.

I reproduced the ordering: with `/a/b` freshly re-listed but `/a/b/c.txt` still in `entries`, `stat("/a/b/c.txt")` returns the stale `FILE` instead of `NotFound`. The same "never evicted" property means `entries`/`children`/`expiry` grow unbounded (no TTL sweep, no cap), so crawling a large bucket leaks metadata for the process lifetime. Direction: check the owning listing's `expiry` in `get()` (drop expired lazily), consult the negative cache before returning a positive whose listing is stale, and have `invalidate_dir` drop entries by path-prefix rather than the direct-child vec.

<details>
<summary>Tests — both currently fail</summary>

```rust
// in `mod tests` of cache.rs
#[test]
fn stale_grandchild_does_not_survive_subtree_invalidation() {
    let c = IndexCache::new(Duration::from_secs(600));
    c.set_dir("/a/b", &[de("c.txt", FileKind::File, 1234)]); // old listing
    c.set_dir("/a", &[]);      // /a re-listed, /a/b gone
    c.invalidate_dir("/a/b");  // subtree dropped
    assert!(
        c.get("/a/b/c.txt").is_none(),
        "stale grandchild entry survived subtree invalidation"
    );
}

#[test]
fn expired_entry_is_not_served_as_fresh() {
    let c = IndexCache::new(Duration::from_millis(0)); // immediately expired
    c.set_dir("/", &[de("gone.txt", FileKind::File, 10)]);
    assert!(!c.is_listed("/"));
    assert!(
        c.get("/gone.txt").is_none(),
        "expired entry still served via the no-TTL get() fast path"
    );
}
```
</details>

backend/src/vfs/accessor/notion.rs
```diff
@@ -0,0 +1,173 @@
...
+        // Content-Type is set by `.json(...)` on the builder; do not add a
+        // second one here (duplicate Content-Type makes Notion ignore the body).
+        let resp = req
+            .header("Authorization", format!("Bearer {}", self.api_key))
+            .header("Notion-Version", NOTION_VERSION)
+            .send()
+            .await?;
+        let status = resp.status();
+        let body = resp.text().await.unwrap_or_default();
+        if !status.is_success() {
```
nuri-yoo: `send()` treats every non-2xx as a hard `bail!` — no retry, no `Retry-After`. Rendering one `page.json` is `get_page` plus `list_block_tree`, which issues one `list_children` per block-with-children, sequentially, to depth 10; against Notion's ~3 req/s limit a single medium page can trip 429 and fail the whole read. A bounded retry that honors `Retry-After` on 429 (and optionally 5xx) would make reads resilient.

Related: `NotionResource::stat` of a page dir (resource/notion.rs:173) runs the full render and maps every error — including 429 and transient failures — to `NotFound`, so under rate-limiting an existing page reports 404. Once retry is in place, consider a lighter existence check there and distinguishing "missing" from "temporarily failed."

nuri-yoo (commented):
Review of External source connection. The overall direction is sound, and the unified tree plus the forwarder's hang-proofing are well done. Two of the inline points want handling before merge: the Notion `stat` underflow panic and the forwarder reporting a failed write as success. The rest — the `..` traversal on the forward path, S3 special-character keys, cache eviction, and Notion 429 handling — would be good to address in the same cycle. The `block_id` validation isn't reachable today, but it's worth landing before the `.cmd` write dispatch is wired. Details are in each inline comment.

backend/src/vfs/resource/notion.rs
grf53: Good to cover this logical defect even though there is no real notion page corresponding to `/pages/page.json`.
I'd like to accept your suggestion, but there would be a bug that is: `/pages/<page-dir>` would be lead to `NotFound` instead of `FileStat` with `FileKind::Dir`.
So let me use `rest.iter().nth_back(1)` as index instead.

workspace/src/vfs/sandbox/forward.rs
grf53: This points out the same issue with https://github.com/brekkylab/agent-k/pull/206#discussion_r3570036800.
This would be covered in a separate PR by @selloriwoo, as I understand.

workspace/src/vfs/accessor/notion.rs
grf53: Added the `valid_notion_id()` function using `UUID:try_parse()`. Thanks.

workspace/src/vfs/sandbox/forward.rs
selloriwoo: This part is also resolved! The test passes in that PR [(#209).](https://github.com/brekkylab/agent-k/pull/209)

workspace/src/vfs/cache.rs
grf53: Here're the fixes.

1. **`get()` checks its owning listing's TTL** — an entry is served only while its parent listing is fresh; a stale/expired one is dropped on access and reported absent.
2. **`set_dir()` drops vanished children on re-listing** — a child gone from the new listing has its subtree purged, so a deleted object stops stat'ing as present.
3. **`invalidate_dir()` purges by path prefix (`purge_prefix`)** — invalidating a directory removes its whole subtree, not just direct children, so `rmdir`/`rename` leaves no grandchild entries behind.

backend/src/vfs/accessor/notion.rs
grf53: Now it retries Notion `send()` on 429/5xx twice (0.5s→2s, honoring `Retry-After`, capped at 3s) instead of failing on every non-2xx, and propagate `stat`'s page-dir render failure as `Backend` rather than `NotFound` so a transient rate-limit isn't misreported as a missing page.

nuri-yoo (approved):
LGTM. All the inline points landed with regression tests.

The local `files/` `..` traversal is the one carryover, fixed in #209 — but that fix does not carry over to this branch automatically. This branch moves `WorkspaceFs` to `workspace/src/fs.rs`, and a merge simulation (`git merge-tree`) reports a modify/delete conflict on `backend/src/state/workspace/fs.rs` instead of following the move, so #209's change (the lexical fold in `resolve` plus its `resolve_tests`) needs to be ported into `workspace/src/fs.rs` before this merges.

The port matters more here than on main: the guest forward path reaches `resolve` with no pre-normalization anywhere in between (`?path=/files/../../…` → `guest_local_rel` → `resolve`; `VPath` only fixes slashes), so merging without the port opens that traversal to the sandbox guest. When porting, consider folding the rel path once and reusing it for the `FsEvent` paths as well, so hook classification and the on-disk path can't disagree on inputs like `/knowledge/../x`.

jhlee525 (approved):

ljhh-0611 (approved):

PR #205, opened 2026-07-10
Workspace in browser

grf53:
## Description

This PR makes the workspace file trees browsable from the frontend over WebDAV, and optimize it so listings and file reads are correct and fast — including render-on-read providers like Notion.

WebDAV becomes the single browsing surface: a standard WebDAV client can list and read the tree, so the frontend needs no bespoke server API. The view matches the guest exactly — local files under `files/`, each provider mount as a sibling (`notion/`, `s3-…/`).

## Changes

**WebDAV serves the unified tree at `/sources`(renamed from `/files`)**
- Mounted at `/workspaces/{wid}/sources`, serving the same `ForwardFs` projection the guest sees (previously WebDAV exposed the inherent layout — local files at the root — so the views diverged).
- Listings/metadata come from the projection; `open`/create/remove/rename translate back to the inherent path (`WorkspaceFs::unified_to_inherent`), reusing the workspace's streaming, seek, and `knowledge/` write hooks.
- Authenticates from the `Authorization: Bearer` header (the frontend `webdav` client sets it directly), replacing the earlier `?token=` query param — which leaked the token into URLs/logs and which no standard WebDAV client appends. (This is mentioned in https://github.com/brekkylab/agent-k/pull/201#discussion_r3519305713 before.)

```
{Backend URL}/workspaces/{wid}/sources/       ← WebDAV root (ForwardFs integrated projection)
│
├── files/                                    ← local workspace files (RW)
│   │                                            host disk: data_root/workspaces/{wid}/files
│   ├── notes.txt
│   ├── report.pdf
│
├── notion/                                   ← Notion mount
│   └── pages/
│       └── <title>__<page-id>/               ← one page = directory
│           ├── page.json                     ← rendered content of page
│           └── <child-title>__<child-id>/    ← subpages
│               ├── page.json
│               └── …
│
└── s3-prod/                                  ← S3 mount — bucket key is the tree root
    ├── image.png
    └── data/
        └── q3.csv
```

**Correctness — render-on-read sizing**
- *Before*: Notion renders `page.json` only on read, so `stat`/`readdir` reported size 0; WebDAV took that as `Content-Length` and served an empty body. (FUSE and the old JSON API read to EOF, so only WebDAV broke.)
- *After*: `stat`/`readdir` render once through the content cache to get the real size, so listings/`stat`/`HEAD` are correct and the following read hits the cache — still one render.

**Performance — smooth browsing**
- Per-workspace `Vfs` cached across requests (provider clients + metadata cache persist instead of being rebuilt each request); evicted on mount change.
- Whole-object **content cache validated by backend version** (S3 ETag / Notion `last_edited_time`), so external edits refetch without a TTL wait; directory listings behind a shorter TTL(300s).
- Sequential reads serve continuation chunks (offset > 0) from the cached snapshot, skipping the per-chunk revalidation `stat` — a Notion `get_page` per 16 KiB chunk was ~250 ms, so a large page took seconds to open.

## Testing

- `cargo test -p agent-k-backend` — 30 passed, 0 failed.
- [webdav-verify.html](https://github.com/user-attachments/files/29889629/webdav-verify.html)

## Notes
- Content cache is capped at 8 MiB; a render-on-read object larger than that re-renders on each read.

### DIFF
--- a/backend/src/state/workspace/mod.rs
+++ b/backend/src/state/workspace/mod.rs
@@ -1,5 +1,6 @@
+use std::collections::HashMap;
 use std::path::PathBuf;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use ::workspace::{FsConfig, FsEvent, FsHook, WorkspaceFs};
 use chrono::{DateTime, Utc};
@@ -72,11 +73,42 @@ impl Workspace {
 pub struct WorkspacesState {
     db: SqlitePool,
     data_root: PathBuf,
+    /// Per-workspace assembled [`WorkspaceFs`], cached so provider clients and
+    /// their metadata `CachedResource` survive across requests instead of being
+    /// rebuilt each [`Self::get_fs`]. Cloning shares the mounts' `Arc<Resource>`,
+    /// so cache hits reuse the same caches. Evicted on mount change or removal.
+    fs_cache: Mutex<HashMap<Uuid, WorkspaceFs>>,
 }
 
 impl WorkspacesState {
     pub fn new(db: SqlitePool, data_root: PathBuf) -> Self {
-        Self { db, data_root }
+        Self {
+            db,
+            data_root,
+            fs_cache: Mutex::new(HashMap::new()),
+        }
+    }
+
+    /// The workspace's assembled [`WorkspaceFs`], built once and cached (see
+    /// [`Self::fs_cache`]). The returned handle shares the cached mounts.
+    async fn fs_for(&self, wid: Uuid) -> StateResult<WorkspaceFs> {
+        if let Some(fs) = self.fs_cache.lock().unwrap().get(&wid).cloned() {
+            return Ok(fs);
+        }
+        // Build outside the lock; a concurrent miss builds twice, first wins.
+        let built = self.build_fs(wid).await?;
+        Ok(self
+            .fs_cache
+            .lock()
+            .unwrap()
+            .entry(wid)
+            .or_insert(built)
+            .clone())
+    }
+
+    /// Evict `wid`'s cached [`WorkspaceFs`] so the next [`Self::get_fs`] rebuilds it.
+    pub(super) fn invalidate_fs(&self, wid: Uuid) {
+        self.fs_cache.lock().unwrap().remove(&wid);
     }
 
     pub async fn get(&self, id: Uuid) -> StateResult<Option<Workspace>> {
@@ -129,6 +161,7 @@ impl WorkspacesState {
     /// are handled by the database's foreign keys.
     pub async fn remove(&self, id: Uuid) -> StateResult<Workspace> {
         let existing = self.get(id).await?.ok_or(StateError::NotFound)?;
+        self.invalidate_fs(id);
         self.remove_files(id).await?;
         sqlx::query("DELETE FROM workspaces WHERE id = ?")
             .bind(id.to_string())
@@ -178,8 +211,8 @@ impl WorkspacesState {
     /// workspace's external-provider mounts attached (paths under a mount prefix
     /// route to the provider; everything else stays local).
     pub async fn get_fs(&self, wid: Uuid) -> StateResult<WorkspaceFs> {
-        // `build_fs` assembles the full mount table (local `/files` + providers).
-        let fs = self.build_fs(wid).await?;
+        // Cached assembly (local `/files` + providers); the hook is per-call.
+        let fs = self.fs_for(wid).await?;
         Ok(fs.with_hook(knowledge_hook(wid)))
     }
 
--- a/workspace/src/vfs/cache.rs
+++ b/workspace/src/vfs/cache.rs
@@ -5,7 +5,7 @@
 //! [`Resource`] so both FUSE frontends benefit transparently.
 
 use std::{
-    collections::HashMap,
+    collections::{BTreeMap, HashMap},
     ops::Range,
     sync::{Arc, Mutex},
     time::{Duration, Instant},
@@ -19,10 +19,11 @@ use crate::vfs::{
     resource::{DirEntry, FileKind, FileStat, Resource},
 };
 
-/// Default listing TTL (600s).
-const DEFAULT_TTL: Duration = Duration::from_secs(600);
+/// Directory-listing TTL. Listings have no cheap version to revalidate against
+/// (unlike file content), so they fall back to a short expiry.
+const LISTING_TTL: Duration = Duration::from_secs(300);
 
-#[derive(Clone, Copy)]
+#[derive(Clone)]
 struct Entry {
     is_dir: bool,
     size: u64,
@@ -32,6 +33,9 @@ struct Entry {
     mtime: Option<std::time::SystemTime>,
     atime: Option<std::time::SystemTime>,
     ctime: Option<std::time::SystemTime>,
+    /// Strong version tag (S3 `ETag`) carried from the listing, so the stat
+    /// fast-path can hand `open` a pin-able snapshot (`If-Match`).
+    etag: Option<String>,
 }
 
 #[derive(Default)]
@@ -71,7 +75,7 @@ impl IndexCache {
             inner.entries.remove(path);
             return None;
         }
-        inner.entries.get(path).copied()
+        inner.entries.get(path).cloned()
     }
 
     /// Whether `path`'s listing is present and unexpired (basis for negative
@@ -106,6 +110,7 @@ impl IndexCache {
                     atime: e.atime,
                     ctime: e.ctime,
                     created: None,
+                    etag: e.etag.clone(),
                 })
             })
             .collect();
@@ -150,6 +155,7 @@ impl IndexCache {
                     mtime: e.mtime,
                     atime: e.atime,
                     ctime: e.ctime,
+                    etag: e.etag.clone(),
                 },
             );
         }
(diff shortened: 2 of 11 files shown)


backend/src/vfs/cache.rs
```diff
@@ -155,33 +156,174 @@ impl IndexCache {
...
+type ContentEntry = (Version, Arc<Vec<u8>>);
+
+/// Wraps a provider [`Resource`] with an [`IndexCache`] for listings (short TTL)
+/// plus a **validated** content cache: a read probes the object's cheap version
+/// via `stat` and serves cached bytes only while it matches, so external edits
+/// are refetched without a TTL wait. Mutations invalidate both.
 pub struct CachedResource {
     inner: Arc<dyn Resource>,
     cache: IndexCache,
+    /// path -> (validator, full bytes). Held until the validator changes or the
```
selloriwoo: The description notes the content cache is "capped at 8 MiB", but that's a per-entry cap — there's no limit on the number of entries. Combined with Vfs now persisting for the process lifetime (vfs_cache, evicted only on mount change), the total footprint grows with every distinct file read and is never reclaimed during normal operation: an agent reading across a mounted S3 source pins each object ≤ 8 MiB in server memory until restart (~8 MiB × distinct paths × workspaces, shared across all users in the process).

Not a crash — a slow climb whose impact scales with file count and deployment size. If unbounded retention is intentional, a one-line note on the assumption would help; otherwise bound it with a total-size (or entry-count) LRU, or drop entries not read in a while. Eviction is lossless here since every read already revalidates against the backend version — a dropped entry just re-fetches on the next read.

workspace/src/vfs/cache.rs
```diff
@@ -155,33 +156,174 @@ impl IndexCache {
...
+        // first chunk cached, skipping the per-chunk provider stat (a Notion
+        // `get_page` is a network round-trip — an N-chunk read would be N stats).
+        if matches!(&range, Some(r) if r.start > 0)
+            && let Some(full) = self.content_peek(key)
+        {
+            return Ok(slice(&full, &range));
+        }
+        // First chunk / whole read / uncached: validate the cheap version and
+        // serve cached bytes only while it matches.
+        let st = self.inner.stat(path).await?;
```
selloriwoo, line 406: Large files re-stat the provider on every chunk.

Files are read in 256 KiB chunks, and each chunk calls read_bytes, which does a stat first to get the version it validates the cache against. A later chunk can skip that stat only if the file's bytes are already in content — but files > 8 MiB are never cached, so the lookup always misses and every chunk stats again.

The thing is, stat returns whole-file metadata (size, version) — it's the same for every chunk and doesn't need re-checking as we move through the file. Only the bytes differ per chunk. So for a 100 MB S3 object we go from N GetObject (before this PR) to N HeadObject + N GetObject. Notion is worse: it has no byte-range, so each chunk re-renders the entire page (get_page + full list_block_tree).

Root cause: "have I already validated this file?" is remembered only through the byte cache, so the memory is lost exactly when the file is too big to cache. Could we grab the version once when the file is opened and reuse it for the rest of the chunks, regardless of size? Then it's one stat per read instead of one per chunk. (For S3 the GetObject response already includes the ETag, so we could drop the upfront stat entirely.) Bonus: this also closes the offset > 0 stale-read gap, since the whole read would share one consistent version.
<br/>
<br/>

BEFORE (now) - every chunk re-stats:
  chunk1  : stat + read
  chunk2  : stat + read
  ...     : ...
  chunk400: stat + read
  => total: stat x400 (HeadObject) + read x400 (GetObject)

AFTER (grab the version once at open):
  open    : stat           <- once
  chunk1  :        read
  chunk2  :        read
  ...     :        read
  chunk400:        read
  => total: stat x1 + read x400   (read count same - bytes differ per chunk)

backend/src/vfs/cache.rs
grf53: Thanks. the content cache is now a total-byte LRU (128 MiB per provider).

workspace/src/vfs/cache.rs
grf53: Now it skips stat check regardless of caching.

selloriwoo (approved):

backend/src/state/workspace/mod.rs
```diff
@@ -72,11 +75,40 @@ impl Workspace {
 pub struct WorkspacesState {
     db: SqlitePool,
     data_root: PathBuf,
+    /// Per-workspace [`Vfs`], cached so provider clients and their metadata
+    /// `CachedResource` survive across requests instead of being rebuilt each
+    /// [`Self::get_fs`]. `None` = no mounts; evicted on mount change or removal.
+    vfs_cache: Mutex<HashMap<Uuid, Option<Arc<Vfs>>>>,
```
ljhh-0611: ~I think that it'd better to give vfs (specifically, CachedResource in Mount) some limits like length, lifetime or somthing for blocking infinite caching problem.~
Aha, Already done in https://github.com/brekkylab/agent-k/pull/205#discussion_r3570001235 ! Sorry

workspace/src/vfs/cache.rs
```diff
@@ -155,33 +156,250 @@ impl IndexCache {
...
+}
+
+impl Version {
+    fn of(s: &FileStat) -> Self {
+        if let Some(e) = &s.etag {
+            Version::Tag(e.clone())
+        } else if let Some(v) = &s.version {
+            Version::Tag(v.clone())
+        } else if let Some(t) = s.mtime {
+            Version::Time(t)
```
ljhh-0611, line 210: Notion validates by `last_edited_time`, which Notion truncates to the minute — a same-minute re-edit serves stale until a cross-minute edit or LRU eviction. Low-impact, but worth a short content-TTL backstop.

backend/src/vfs/cache.rs
```diff
@@ -155,33 +156,250 @@ impl IndexCache {
...
 
 #[async_trait]
 impl Resource for CachedResource {
     async fn read_bytes(&self, path: &VPath, range: Option<Range<u64>>) -> VfsResult<Vec<u8>> {
-        self.inner.read_bytes(path, range).await
+        let key = path.as_str();
+        // Continuation chunk (offset > 0): skip the per-chunk stat — the version
+        // was validated on the first chunk. Serve from cache, else read the range
+        // straight through (large files aren't cached and need no recheck).
+        if matches!(&range, Some(r) if r.start > 0) {
```
nuri-yoo: This branch serves from `content_peek` with no version check, and content entries have no TTL (only LRU pressure or a write-through drops them). The "continuation" assumption breaks for plain HTTP Range requests: nothing between an external edit and a ranged read revalidates, so a `Range: bytes=6-11` GET keeps returning the old bytes indefinitely. Reproduced end-to-end against the real `DavHandler` with an S3-like mock (stat reports real size + etag): a full GET caches v1, the object is edited behind the cache, then `Range: bytes=6-11` returns 206 with the v1 slice; an offset-0 GET repairs it.

The same mechanism allows a torn response when two sequential reads interleave around an edit: client A's continuation chunks pick up the bytes client B just refetched, so A gets half v1, half v2. Notion-style mounts are mostly shielded by accident (their size-0 stat routes through `resolve_size`, which refreshes the cache before the read), so the practical exposure is S3-type mounts.

`WorkspaceFs::open` already stats the object, so `MountFile` could carry that stat's version and continuation reads could call `content_get` with it instead of `content_peek`. That keeps one stat per read and makes every chunk of a response come from a single validated snapshot.

workspace/src/vfs/cache.rs
```diff
@@ -190,7 +408,23 @@ impl Resource for CachedResource {
         if let Some(entries) = self.cache.list_dir_entries(path.as_str()) {
             return Ok(entries);
         }
-        let entries = self.inner.readdir(path).await?;
+        let mut entries = self.inner.readdir(path).await?;
+        // Eagerly size (and content-cache) files the backend listed as 0 because
+        // sizing needs a render (Notion page.json): render once here so the entry
+        // shows a real size AND a following read is a cache hit. This adds that
+        // render to the listing's latency — the trade for correct sizes up front.
+        for e in entries.iter_mut() {
```
nuri-yoo, line 473: The sizing loop awaits each render in sequence, so a cold listing costs N renders back to back. Measured end-to-end (real `DavHandler`, mock provider with a 100 ms render): a depth-1 PROPFIND over 8 size-0 files took 817 ms, where parallel rendering would finish in roughly one render. A Notion `pages/` directory with a few dozen pages puts a cold PROPFIND into client-timeout territory.

Each iteration only touches its own child path, so this looks safe to run with bounded concurrency (`futures::stream::iter(..).map(..).buffer_unordered(k)` or `for_each_concurrent`), cutting the cold listing to about ceil(N/k) render latencies without changing cache semantics.

workspace/src/vfs/cache.rs
grf53: Thank you for checking this! It would not be easily found, I guess.
Good to have content-TTL separate from the listing-TTL and it would be good to be short to provide realtime-ish content, but it would raise the burden of updating content, let me apply the same TTL as the others for now.

workspace/src/vfs/cache.rs
grf53: Calculating size via rendering applies only when no separate metadata is provided for the content, and is currently limited to Notion.

In Notion, each 'page' is reduced to a single directory containing only a single file with the page's content. Therefore, size calculation via rendering occurs only once at a specific location.

The code is written so that if the size of a new source is not provided in the metadata, it can be calculated via rendering without any separate configuration. In this case, multiple renders like the one you mentioned can occur, and this allows you to prepare for this in advance.

In the current situation, I believe it is better to apply YAGNI and avoid reflecting this in advance.
How do you think about this?

backend/src/vfs/cache.rs
grf53: Reads now pin the version once at open and validate every chunk against it (new `read_bytes_pinned`, with `MountFile` carrying the open-time stat), removing the unvalidated path that let ranged reads serve stale or torn bytes — still one stat per read (7d00e15).

While applying this I found that large (>8 MiB, uncached) S3 objects, having no snapshot to pin, could still tear if replaced mid-stream, so I overrode S3's `read_bytes_pinned` to send `If-Match(etag)` — a mid-stream change returns 412 and the read fails cleanly instead of stitching two versions (745e683).

workspace/src/vfs/cache.rs
nuri-yoo: Agreed on YAGNI.  Resolving.

backend/src/vfs/cache.rs
nuri-yoo: Verified on 745e683: the stale case is closed. A fresh ranged GET now validates against the open-time stat, `content_peek` is gone, and it is still one stat per read (`SeekFrom::End` even drops one). Two gaps remain, both reproduced with unit tests on the PR head (attached below).

The interleaved tear still reproduces. The pin only holds while the pinned snapshot is in the content cache; once it is gone (replaced by another client's refetch, or LRU-evicted under the 128 MiB budget), the miss path in `read_bytes_pinned` falls back to `inner.read_bytes(path, None)`, which reads the *current* bytes with no validation, so A's continuation chunk returns v2 under a v1 pin. It then runs `content_put(key, v1, v2-bytes)`, mislabeling the entry so later v1-pinned chunks keep hitting it. This is the same A/B interleave from my original comment; `pinned_read_interleaved_with_refetch` below fails with "A got v2 bytes under a v1 pin". Minimal fix: delegate the miss-path fetch to `self.inner.read_bytes_pinned(path, None, stat)`, so S3 fails cleanly via If-Match, and the `content_put` stays correct because a successful fetch implies the bytes match the pin.

The If-Match pin from 745e683 is inactive on the common path. `CachedResource::stat` rebuilds cached-file stats with `..Default::default()`, dropping `etag`/`version` (and `DirEntry` never carried the etag from the listing in the first place). A WebDAV GET normally follows a PROPFIND within the 300 s listing TTL, so `open` pins a stat with `etag: None` and the S3 override sends an unconditional GET — the >8 MiB protection only engages when the open-time stat happens to miss the metadata cache. `cached_stat_preserves_etag_for_pinning` below fails with `left: None, right: Some("abc123")`. Carrying `e_tag` from the `object_store` listing into `DirEntry` and the metadata cache (or bypassing the cache for the open-time stat) would restore it. A side effect of the same drop: the pinned path validates by `Time(mtime)` while fresh reads validate by `Tag(etag)`, so the two paths keep invalidating each other's content-cache entries.

The two are coupled: the first fix only has teeth on S3 when the pinned stat actually carries an etag, so it needs the second.

<details>
<summary>Repro tests (add to backend/src/vfs/cache.rs tests; on 745e683: 8 passed, 2 failed)</summary>

```rust
// A pins v1, B's fresh read replaces the cache entry with v2, then A's
// continuation chunk must still be v1 (or fail).
#[tokio::test]
async fn pinned_read_interleaved_with_refetch() {
    use std::sync::atomic::{AtomicU64, Ordering};
    use std::time::{Duration as D, UNIX_EPOCH};

    struct Obj {
        ver: AtomicU64,
        data: Mutex<Vec<u8>>,
    }
    #[async_trait]
    impl Resource for Obj {
        async fn read_bytes(&self, _p: &VPath, range: Option<Range<u64>>) -> VfsResult<Vec<u8>> {
            Ok(slice(self.data.lock().unwrap().as_slice(), &range))
        }
        async fn write_bytes(&self, _p: &VPath, _d: Vec<u8>) -> VfsResult<()> {
            Ok(())
        }
        async fn readdir(&self, _p: &VPath) -> VfsResult<Vec<DirEntry>> {
            Ok(vec![])
        }
        async fn stat(&self, _p: &VPath) -> VfsResult<FileStat> {
            Ok(FileStat {
                kind: FileKind::File,
                mtime: Some(UNIX_EPOCH + D::from_secs(self.ver.load(Ordering::SeqCst))),
                ..Default::default()
            })
        }
    }

    let obj = Arc::new(Obj {
        ver: AtomicU64::new(1),
        data: Mutex::new(b"hello world".to_vec()),
    });
    let cached = CachedResource::new(obj.clone());
    let p = VPath::new("/f.txt");

    // Client A opens (pins v1) and reads its first chunk.
    let s1 = obj.stat(&p).await.unwrap();
    assert_eq!(cached.read_bytes_pinned(&p, Some(0..5), &s1).await.unwrap(), b"hello");

    // External edit -> v2.
    *obj.data.lock().unwrap() = b"EDITED text".to_vec();
    obj.ver.store(2, Ordering::SeqCst);

    // Client B opens fresh (pins v2) and reads; this refetches and replaces
    // the cache entry with v2 bytes.
    let s2 = obj.stat(&p).await.unwrap();
    assert_eq!(cached.read_bytes_pinned(&p, Some(0..5), &s2).await.unwrap(), b"EDITE");

    // Client A's continuation chunk, still pinned to v1: must be v1 bytes
    // ("world") or a clean error; anything else is a torn response.
    let a2 = cached.read_bytes_pinned(&p, Some(6..11), &s1).await;
    match a2 {
        Ok(bytes) => assert_eq!(bytes, b"world", "torn response: A got v2 bytes under a v1 pin"),
        Err(_) => {} // clean failure is acceptable
    }
}

// A stat served from the metadata cache (primed by readdir, i.e. the
// PROPFIND-then-GET flow) must keep the provider's etag; that is the token
// the pinned S3 read sends as If-Match.
#[tokio::test]
async fn cached_stat_preserves_etag_for_pinning() {
    struct S3Like;
    #[async_trait]
    impl Resource for S3Like {
        async fn read_bytes(&self, _p: &VPath, range: Option<Range<u64>>) -> VfsResult<Vec<u8>> {
            Ok(slice(b"0123456789", &range))
        }
        async fn write_bytes(&self, _p: &VPath, _d: Vec<u8>) -> VfsResult<()> {
            Ok(())
        }
        async fn readdir(&self, _p: &VPath) -> VfsResult<Vec<DirEntry>> {
            Ok(vec![DirEntry {
                name: "f.bin".into(),
                kind: FileKind::File,
                size: 10,
                mtime: Some(std::time::UNIX_EPOCH),
                atime: None,
                ctime: None,
            }])
        }
        async fn stat(&self, _p: &VPath) -> VfsResult<FileStat> {
            Ok(FileStat {
                kind: FileKind::File,
                size: 10,
                mtime: Some(std::time::UNIX_EPOCH),
                etag: Some("\"abc123\"".into()),
                ..Default::default()
            })
        }
    }

    let cached = CachedResource::new(Arc::new(S3Like));

    // PROPFIND primes the metadata cache ...
    cached.readdir(&VPath::new("/")).await.unwrap();
    // ... then GET opens the file; WorkspaceFs::open pins this stat.
    let pinned = cached.stat(&VPath::new("/f.bin")).await.unwrap();
    assert_eq!(
        pinned.etag.as_deref(),
        Some("\"abc123\""),
        "etag dropped by the cache fast-path: If-Match pin is disabled"
    );
}
```

</details>

grf53: This is reflected to the rebased version.

nuri-yoo (approved):
LGTM

jhlee525 (approved):

ljhh-0611 (approved):

PR #206, opened 2026-07-10
Source indexing

grf53:
## Description

Enables the backend to access workspace sources — local files and mounted remote sources (S3/Notion) — for indexing, and adds the structures and APIs for the app to designate which targets get indexed.

Index membership is a set of **references** under `/knowledge`: each `*.ref` file holds the unified-tree path of a target (local file, mounted object, or directory), and one scanner reads them all through `WorkspaceFs` — making the indexer a **third client** of the unified tree alongside the WebDAV (browser) and in-guest FUSE (sandbox) frontends.

Indexing delegates to the existing `agent_k::knowledge_base::Store`. Because document ids are a content hash, change detection falls out of a resync/reconcile — **no fingerprint table, no schema change.**

## Reference format (Alternative suggestions are welcome.!)

A small JSON file under `/knowledge`, named `<slug>-<nanoid>.ref`:

```jsonc
// /knowledge/q3-report-a1B2c3.ref
{ "path": "/notion/reports/q3.pdf" }
```

`path` is a unified-tree absolute path — local (`/docs/a.md`) or mounted (`/notion/…`) — pointing at a file or a directory.
A reference whose target points into /knowledge is skipped, so the reference files never index themselves (and a directory target can't recurse into `/knowledge`).

### Why references (not a folder of files)

- Mounted external items are read-only and live in the provider; they can't physically be dropped into a `knowledge/` folder. A reference expresses "index this" as a path, so local and external are handled identically.
- Reading at the unified `WorkspaceFs` level (not the raw `Vfs`/`Resource` layer) means a single scanner covers both local disk and mounts.
- No per-item state to persist. Content-hash document ids mean identity comes from the bytes, and each resync reconciles the referenced set against `Store.list()` — so the `Store` is the sole source of index state, with no `(path → fingerprint)` table.

## Changes

**Core (`state/workspace/`)**
- `resync.rs` (new) — `Resyncer`: per-workspace `Store` cache + resync lock; `resync()` reconcile (scan `/knowledge` refs → resolve targets via `WorkspaceFs` → `ingest_many` → purge unreferenced → `compact`); `collect_referenced()`/`add_target()` scanner; `KnowledgeRef`.
- `fs.rs` — `WorkspaceFs` gains a `read_to_end` helper and an injected `Resyncer`; `/knowledge` mutations (write/flush, rename, copy, delete) now spawn a resync, replacing the former logging stubs.
- `mod.rs` — `WorkspacesState` owns the `Resyncer`, injects it into every `get_fs`, and exposes `resyncer()` / `all_ids()`.

**HTTP + scheduling**
- `router/knowledge.rs` (new) — browser-facing membership API:
  - `GET/POST /workspaces/{wid}/knowledge/refs` — list / create a reference (validates the target exists, local or mounted)
  - `DELETE /workspaces/{wid}/knowledge/refs/{name}`
  - `POST /workspaces/{wid}/knowledge/resync` — force a resync
- `main.rs` — optional periodic resync loop (`AGENT_K_INDEX_INTERVAL_SECS`, disabled by default) to pick up provider-side changes to external targets, which produce no local write event.

## Testing

- `resync::tests` — scanner gathers local files, a directory's indexable descendants, and a mounted object while skipping non-`.ref` files, self-references, missing targets, and unsupported extensions; absent `/knowledge` yields an empty set.
- Full suite green (`cargo test -p agent-k-backend`); no clippy warnings in changed files.

## Notes

- **Change detection is automatic:** unchanged bytes → same content-hash id → `ingest` is a no-op; changed bytes → new id, old id purged as stale. Resyncs are serialized per workspace.
- **Purge safety valve:** a non-empty member set that ingests *nothing* (e.g. metadata provider unavailable) skips the purge, so a healthy index isn't wiped.
- **Ingest needs an ailoy provider** (title/purpose via ailoy's process-global default), which `backend_v2` doesn't initialize yet — so ingest currently fails at runtime (the safety valve prevents index loss). Wiring it is a prerequisite for this to actually index.
- **`/knowledge` is references-only:** dropping a real file there won't index it; create a reference.
- **Planned follow-ups:** skip re-reading unchanged large external targets via etag/version; isolate each source in its own `Store`.

### DIFF
--- a/backend/src/router/knowledge.rs
+++ b/backend/src/router/knowledge.rs
@@ -23,7 +23,7 @@ use uuid::Uuid;
 use ::workspace::{FsError as WsFsError, OpenOptions, WorkspaceFs};
 
 use crate::auth::AuthUser;
-use crate::state::{AppState, KNOWLEDGE_DIR, KnowledgeRef, REF_SUFFIX};
+use crate::state::{AppState, KNOWLEDGE_ROOT, KnowledgeRef, REF_SUFFIX};
 
 use super::{error::ApiError, error::err, workspace::require_owned_workspace};
 
@@ -75,13 +75,13 @@ pub(super) async fn create_ref(
         Err(_) => return Err(err(StatusCode::INTERNAL_SERVER_ERROR, "internal error")),
     }
 
-    match fs.create_dir(KNOWLEDGE_DIR).await {
+    match fs.create_dir(KNOWLEDGE_ROOT).await {
         Ok(()) | Err(WsFsError::Exists) => {}
         Err(_) => return Err(err(StatusCode::INTERNAL_SERVER_ERROR, "internal error")),
     }
 
     let name = ref_name_for(&target);
-    let ref_path = format!("{KNOWLEDGE_DIR}/{name}");
+    let ref_path = format!("{KNOWLEDGE_ROOT}/{name}");
     let body = KnowledgeRef {
         path: target.clone(),
     }
@@ -113,7 +113,7 @@ pub(super) async fn list_refs(
         .map_err(|_| err(StatusCode::INTERNAL_SERVER_ERROR, "internal error"))?;
 
     let mut items = Vec::new();
-    match fs.read_dir(KNOWLEDGE_DIR).await {
+    match fs.read_dir(KNOWLEDGE_ROOT).await {
         Ok(mut stream) => {
             while let Some(entry) = stream.next().await {
                 let Ok(entry) = entry else { continue };
@@ -121,7 +121,7 @@ pub(super) async fn list_refs(
                 if !name.ends_with(REF_SUFFIX) {
                     continue;
                 }
-                let ref_path = format!("{KNOWLEDGE_DIR}/{name}");
+                let ref_path = format!("{KNOWLEDGE_ROOT}/{name}");
                 let Ok(bytes) = read_all(&fs, &ref_path).await else {
                     continue;
                 };
@@ -154,7 +154,7 @@ pub(super) async fn delete_ref(
         .get_fs(wid)
         .await
         .map_err(|_| err(StatusCode::INTERNAL_SERVER_ERROR, "internal error"))?;
-    match fs.remove_file(&format!("{KNOWLEDGE_DIR}/{name}")).await {
+    match fs.remove_file(&format!("{KNOWLEDGE_ROOT}/{name}")).await {
         Ok(()) => Ok(StatusCode::NO_CONTENT),
         Err(WsFsError::NotFound) => Err(err(StatusCode::NOT_FOUND, "reference not found")),
         Err(_) => Err(err(StatusCode::INTERNAL_SERVER_ERROR, "internal error")),
--- a/backend/src/router/knowledge.rs
+++ b/backend/src/router/knowledge.rs
@@ -1,0 +1,262 @@
+//! HTTP API for a workspace's knowledge index membership.
+//!
+//! Membership is a set of references under `/files/knowledge`: each `*.ref` file
+//! holds the unified-tree path of a target (a local file, a mounted object, or a
+//! directory). Writing / deleting a reference goes through
+//! [`WorkspaceFs`](::workspace::WorkspaceFs), whose `/files/knowledge` hook spawns
+//! a resync — so these routes never touch the resyncer directly except for the
+//! explicit `resync`.
+
+use std::sync::Arc;
+
+use axum::{
+    Extension, Json,
+    extract::{Path, State},
+    http::StatusCode,
+};
+use bytes::Bytes;
+use futures_util::StreamExt as _;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+use ::workspace::{FsError as WsFsError, OpenOptions, WorkspaceFs};
+
+use crate::auth::AuthUser;
+use crate::state::{AppState, KNOWLEDGE_DIR, KnowledgeRef, REF_SUFFIX};
+
+use super::{error::ApiError, error::err, workspace::require_owned_workspace};
+
+#[derive(Debug, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct CreateRefRequest {
+    /// Absolute unified-tree path of the target to index, e.g. `/files/docs/q3.pdf`
+    /// (local) or `/s3-prod/reports/q3.pdf` (mounted). May point at a directory.
+    pub target_path: String,
+}
+
+#[derive(Debug, Serialize, JsonSchema)]
+pub struct RefResponse {
+    pub name: String,
+    pub target_path: String,
+}
+
+#[derive(Debug, Serialize, JsonSchema)]
+pub struct RefListResponse {
+    pub items: Vec<RefResponse>,
+}
+
+/// `POST /workspaces/{wid}/knowledge/refs` — reference a target into knowledge.
+pub(super) async fn create_ref(
+    State(state): State<Arc<AppState>>,
+    Extension(auth): Extension<AuthUser>,
+    Path(wid): Path<Uuid>,
+    Json(payload): Json<CreateRefRequest>,
+) -> Result<(StatusCode, Json<RefResponse>), ApiError> {
+    require_owned_workspace(&state, &auth, wid).await?;
+
+    let target = clean_target(&payload.target_path).map_err(|m| err(StatusCode::BAD_REQUEST, m))?;
+    if is_under_knowledge(&target) {
+        return Err(err(
+            StatusCode::BAD_REQUEST,
+            "cannot reference the knowledge directory itself",
+        ));
+    }
+
+    let fs = state
+        .workspaces
+        .get_fs(wid)
+        .await
+        .map_err(|_| err(StatusCode::INTERNAL_SERVER_ERROR, "internal error"))?;
+
+    match fs.metadata(&target).await {
+        Ok(_) => {}
+        Err(WsFsError::NotFound) => return Err(err(StatusCode::NOT_FOUND, "target not found")),
+        Err(_) => return Err(err(StatusCode::INTERNAL_SERVER_ERROR, "internal error")),
+    }
+
+    match fs.create_dir(KNOWLEDGE_DIR).await {
+        Ok(()) | Err(WsFsError::Exists) => {}
+        Err(_) => return Err(err(StatusCode::INTERNAL_SERVER_ERROR, "internal error")),
+    }
+
+    let name = ref_name_for(&target);
+    let ref_path = format!("{KNOWLEDGE_DIR}/{name}");
+    let body = KnowledgeRef {
+        path: target.clone(),
+    }
+    .to_bytes();
+    write_ref(&fs, &ref_path, body)
+        .await
+        .map_err(|_| err(StatusCode::INTERNAL_SERVER_ERROR, "failed to write reference"))?;
+
+    Ok((
+        StatusCode::CREATED,
+        Json(RefResponse {
+            name,
+            target_path: target,
+        }),
+    ))
+}
+
+/// `GET /workspaces/{wid}/knowledge/refs` — list the workspace's references.
+pub(super) async fn list_refs(
+    State(state): State<Arc<AppState>>,
+    Extension(auth): Extension<AuthUser>,
+    Path(wid): Path<Uuid>,
+) -> Result<Json<RefListResponse>, ApiError> {
+    require_owned_workspace(&state, &auth, wid).await?;
+    let fs = state
+        .workspaces
+        .get_fs(wid)
+        .await
+        .map_err(|_| err(StatusCode::INTERNAL_SERVER_ERROR, "internal error"))?;
+
+    let mut items = Vec::new();
+    match fs.read_dir(KNOWLEDGE_DIR).await {
+        Ok(mut stream) => {
+            while let Some(entry) = stream.next().await {
+                let Ok(entry) = entry else { continue };
+                let name = String::from_utf8_lossy(&entry.name()).into_owned();
+                if !name.ends_with(REF_SUFFIX) {
+                    continue;
+                }
+                let ref_path = format!("{KNOWLEDGE_DIR}/{name}");
+                let Ok(bytes) = read_all(&fs, &ref_path).await else {
+                    continue;
+                };
+                if let Ok(r) = serde_json::from_slice::<KnowledgeRef>(&bytes) {
+                    items.push(RefResponse {
+                        name,
+                        target_path: r.path,
+                    });
+                }
+            }
+        }
+        Err(WsFsError::NotFound) => {}
+        Err(_) => return Err(err(StatusCode::INTERNAL_SERVER_ERROR, "internal error")),
+    }
+    Ok(Json(RefListResponse { items }))
+}
+
+/// `DELETE /workspaces/{wid}/knowledge/refs/{name}` — drop a reference.
+pub(super) async fn delete_ref(
+    State(state): State<Arc<AppState>>,
+    Extension(auth): Extension<AuthUser>,
+    Path((wid, name)): Path<(Uuid, String)>,
+) -> Result<StatusCode, ApiError> {
+    require_owned_workspace(&state, &auth, wid).await?;
+    if name.contains('/') || name.contains("..") || !name.ends_with(REF_SUFFIX) {
+        return Err(err(StatusCode::BAD_REQUEST, "invalid reference name"));
+    }
+    let fs = state
+        .workspaces
+        .get_fs(wid)
+        .await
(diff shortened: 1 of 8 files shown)


backend/src/state/workspace/resync.rs
```diff
@@ -0,0 +1,415 @@
...
+
+        // Nothing referenced and no store yet → nothing to reconcile.
+        if member_count == 0 && !self.store_root(wid).join("index").exists() {
+            return Ok(());
+        }
+
+        let store = self.store_for(wid)?;
+        let mut store = store.write().await;
+
+        let result = store.ingest_many(items, PdfEngine::default()).await?;
```
selloriwoo: **`resync` holds every referenced file's bytes in memory at once, though ingest consumes them one at a time.**

`collect_referenced` reads every referenced target fully (`add_target` →
`read_to_end` per file/descendant) into a single `Vec<(Vec<u8>, FileType)>`,
then hands the whole thing to `ingest_many`. There's no size or count cap.

But `ingest_many` processes items sequentially (write corpus → translate →
index, one at a time), so the raw bytes don't all need to be resident. A
directory reference makes this reachable through intended usage — e.g.
`/s3-bucket/reports` expanding to 100 × 20 MB PDFs pulls ~2 GB into RAM in one
resync, and a single workspace can OOM the whole process.

(Note: this is separate from the no-fingerprint design's "re-read on every
resync" tradeoff — that's fine. The avoidable part is holding *all* bytes
simultaneously instead of streaming.)

**Suggested fix — batch the collect→ingest instead of loading everything first.**

up to a tunable byte budget (note ingest already allocates a ~128 MB IndexWriter)

1. Walk the refs and keep appending files to a `batch` until it reaches a
   byte budget (e.g. 128 MB).
2. `ingest_many(batch)` → add the returned `succeeded` ids to a `desired` set →
   drop the batch (frees its bytes before reading the next slice).
3. Repeat until all refs are processed.
4. Only after the last batch, reconcile once:
   `stale = list() − desired` → `purge_many(stale)` → `compact()`.


Only the small `desired` id set crosses batches (never the bytes), and keeping
purge at the end means the reconcile result is unchanged — just bounded memory.

backend/src/state/workspace/resync.rs
```diff
@@ -0,0 +1,415 @@
...
+/// Indexes a workspace's referenced knowledge into its [`Store`]. Cheap to clone
+/// (a pool handle, the data root, and two shared maps), so the same handle is
+/// injected into [`WorkspaceFs`] (write-triggered resync), the knowledge router
+/// (manual resync), and the periodic background loop.
+#[derive(Clone)]
+pub struct Resyncer {
+    db: SqlitePool,
+    data_root: PathBuf,
+    /// One live [`Store`] per workspace, opened on first use.
+    stores: Arc<DashMap<Uuid, SharedStore>>,
```
selloriwoo: **Cached `Store` handles are never evicted on workspace deletion (memory + disk leak, broken teardown contract).**

`store_for` inserts a `SharedStore` (an open tantivy `Index` holding open fds/mmaps)
into `stores` on first resync and it lives for the whole process — `stores` is only
ever read or `entry()`-inserted, never removed (no eviction, no `Drop`, no TTL).
Crucially, neither deletion path cleans it up:

- `delete_user` (wired via `delete_user_admin`) → `WorkspacesState::remove_files`
- `delete_workspace` → `WorkspacesState::remove`

Both `remove_dir_all(workspaces/{wid})` on disk (which includes `knowledge-index/`)
but leave `resyncer.stores[wid]` (and `locks[wid]`) untouched. Since `WorkspacesState`
— and thus the `Resyncer` and its `Arc<DashMap>` — is created once in `AppState`, the
entry persists process-wide.

This violates the teardown choreography documented in `state/mod.rs`:
> "a raw row cascade only drops rows ... sweep the artifacts the FK can't reach.
> Every caller goes through these ... so no delete path can forget the teardown."

The cached `Store`/lock is exactly such a DB-external artifact and was added without
being hooked into that choreography.

**Impact:**
- Memory leak: the tantivy `Index` for a deleted workspace stays resident; repeated
  account deletions accumulate.
- Disk leak (Unix): because the cached `Store` keeps the index files mmapped/open,
  `remove_dir_all` only unlinks the dentries — the inodes/blocks aren't reclaimed
  until the process exits.
- (Minor) Stale handle: if the periodic sweep captured this `wid` from `all_ids()`
  just before deletion, the spawned resync runs against the now-deleted index and errors.

**Suggested fix:** add a `Resyncer::forget(wid)` that drops the `stores`/`locks` entries,
and call it from the deletion choreography. Note `delete_user` goes through
`remove_files` (not `remove`), so both paths must evict — better, evict in the
`state/mod.rs` `delete_*` layer so no future delete path can miss it.

backend/src/state/workspace/resync.rs
selloriwoo: Note: batching won't help a single oversized file — it can't be split
(`read_to_end`/hash/translate need it whole), so a 2 GB PDF ref is read fully
into one `Vec` and can crash the process (OOM) regardless of batch size, with no
cap today. Cheap guard: `add_target` already stats the target, so check
`meta.len` and skip/fail anything over a limit.

backend/src/router/knowledge.rs
```diff
@@ -0,0 +1,247 @@
...
+    Extension(auth): Extension<AuthUser>,
+    Path(wid): Path<Uuid>,
+) -> Result<StatusCode, ApiError> {
+    require_owned_workspace(&state, &auth, wid).await?;
+    state.workspaces.resyncer().spawn_resync(wid);
+    Ok(StatusCode::ACCEPTED)
+}
+
+/// Normalise a target into a unified-tree absolute path, or `None` if empty.
+fn normalize_target(raw: &str) -> Option<String> {
```
nuri-yoo: **`normalize_target` doesn't reject `..`, so a reference can escape the workspace and index files outside it.**

`normalize_target` only trims whitespace, strips a trailing `/`, and ensures a leading `/`. It never rejects `..` segments. The value flows straight into `fs.metadata(&target)` and later `fs.read_to_end(&target)`, and `WorkspaceFs::resolve` resolves it naively: `self.root.join(rel_path.trim_start_matches('/'))`. `WorkspaceFs` has no `..` normalization anywhere, so `..` climbs out of the workspace files root at the OS level.

The result is that an authenticated workspace owner can reference any file the backend process can read — another workspace's files, or host paths like `/etc/passwd` — and its bytes get pulled into their knowledge index.

This is a new surface: the WebDAV frontend is protected because `DavPath::as_rel_ospath` normalizes/rejects `..`, but `create_ref` takes the JSON `target_path` and hands it to `WorkspaceFs` without that normalization. It's the first path that does so.

Reproduced against a running backend (logged in as a normal workspace owner):

```
POST /workspaces/{wid}/knowledge/refs
{"target_path":"/../../../../../../../../../../../etc/passwd"}
→ HTTP 201, ref created
```

The `..` count just has to match the depth of the workspace files root; an over- or under-shoot lands on a nonexistent path and 404s, which is why a wrong count can look like it's rejected when it isn't.

The `DELETE` handler already guards its `name` with `contains("..")`, but `create_ref`'s target is unguarded. A targeted fix here is to reject any target whose segments include `..` (`target.split('/').any(|c| c == "..")`). The more robust option is to normalize and reject root-escaping paths in `WorkspaceFs::resolve` itself, which covers `metadata` / `open` / `read_to_end` in one place rather than per route.

grf53: A similar issue(https://github.com/brekkylab/agent-k/pull/201#discussion_r3518791632) to what you mentioned has been raised previously.
I am still not entirely convinced about checking validity during the `WorkspaceFs::resolve()` step.

Meanwhile, I agree that the ref target path format is structurally weak. Therefore, we have restricted it to absolute paths that do not contain `.` or `..` and introduced helper functions to verify this at both entry points.

backend/src/state/workspace/resync.rs
grf53: 1398d10: Enumerate paths first, then read + ingest in byte-budgeted batches instead of loading everything at once (and skip oversized files).

backend/src/state/workspace/resync.rs
grf53: 29ea8cc: Evict the cached Store handle via `forget` in `remove_files`, the choke point both delete paths pass through.

backend/src/router/knowledge.rs
nuri-yoo: Thanks, the new absolute `.`/`..`-free check at both entry points closes what I ran into. I confirmed the API call now 400s, and a `.ref` planted directly (the WebDAV path) is skipped by the scanner, so the out-of-root bytes no longer reach the index.

One small framing note: in #201 this was still a latent possibility ("not exploitable today" while WebDAV was the only caller). `create_ref` turned out to be the non-WebDAV caller that thread anticipated, so I was able to reproduce it end to end as a normal workspace owner, which is really just confirmation that the concern from #201 had become reachable, and your fix addresses it.

Keeping the guard at the reference boundary seems reasonable. The one thing I'd leave as a possible follow-up is that since `resolve` is the single choke point all fs ops pass through, a lexical containment check there (not a blanket `..` reject, which would break a legit `/a/b/../c`) could be a cheap backstop for any future caller. Not blocking.

nuri-yoo (approved):

selloriwoo (approved):

backend/src/router/knowledge.rs
grf53: Thanks for checking this up.
I have been open to suggestions including extension of meaning of `resolve()`.

selloriwoo: Since you're open to extending `resolve()` — the hesitation about "validity checking there" goes away if we make it **normalization** rather than a **reject**:

```rust
    // Fold the relative path in isolation: we only ever pop from `parts`, and
    // `self.root` is appended untouched, so `..` can never climb past the root.
    let mut parts: Vec<OsString> = Vec::new();
    for comp in Path::new(rel_path.trim_start_matches('/')).components() {
        match comp {
            Component::Normal(c) => parts.push(c.to_os_string()),
            Component::ParentDir => { parts.pop(); }
            _ => {} // CurDir / RootDir / Prefix: ignore
        }
    }
    let mut out = self.root.clone();
    out.extend(parts);
    out
}
```

This isn't a validity check that can fail — it's lexical path folding, so:

- **No `Result`, no call-site churn** — signature stays `-> PathBuf`.
- **Keeps POSIX semantics** (`/..` == `/`).
- **Doesn't break a legit `/a/b/../c`** (→ `/a/c`), which a blanket `..` reject would.
- **Attack paths fold back inside the root** — `/../../etc/passwd` → a nonexistent in-root path → `NotFound`.

The reference-boundary guard stays as-is (a `.`/`..` in a stored ref target is genuinely malformed, so rejecting it there is correct). This is purely a cheap backstop for any future non-WebDAV caller. Non-blocking.

backend/src/state/workspace/resync.rs
```diff
@@ -0,0 +1,556 @@
...
+            batch.push((bytes, ft));
+        }
+        if !batch.is_empty() {
+            ingest_batch(&mut store, batch, wid, &mut desired).await?;
+        }
+
+        // A non-empty set that ingested nothing is a systemic failure (e.g. LLM
+        // provider down), not a full removal — skip the purge so a good index
+        // isn't wiped.
+        if member_count > 0 && desired.is_empty() {
```
ljhh-0611: Partial failure can purge a good doc, not just leave it stale. A target that fails to read hits continue, so its id never enters desired, and purge then deletes it (corpus + origin files, not just the index entry). The desired.is_empty() valve only guards total failure.

 Trigger is the main use case — a mounted S3/Notion ref: a transient range-read error maps to VfsError::Backend → FsError::GeneralFailure (only a real 404 is NotFound), the target drops from desired, and its indexed doc is purged, then re-ingested next clean pass. With the periodic sweep on, a flaky mount makes docs flap.

backend/src/router/knowledge.rs
grf53: It would be best if the suggestion took the form of another complete PR.

backend/src/state/workspace/resync.rs
grf53: I didn't fully understand what you said.

What exactly do you mean by saying that parts need to be changed, and how?

ljhh-0611: Sorry, let me make it concrete with a 2-file example.

Say `/knowledge` has two refs, both already indexed: `a.ref → /docs/a.md` (local) and `b.ref → /s3/b.pdf` (mounted). A resync runs:

1. `collect_targets` returns both → `member_count = 2`.
2. In the read loop, `/docs/a.md` reads fine, but `/s3/b.pdf` hits a transient S3 error, so it takes this arm and is skipped:
   ```rust
   let bytes = match fs.read_to_end(&path).await {
       Ok(b) => b,
       Err(e) => { tracing::warn!(...); continue; }   // b.pdf never ingested this pass
   };
   ```
3. So only `a.md` is ingested → `desired = { id_a }`.
4. The valve doesn't fire, because `desired` is not empty (it has `id_a`):
   ```rust
   if member_count > 0 && desired.is_empty() { return Ok(()); }   // false → continues
   ```
5. Purge computes `stale = everything in the store − desired = { id_b }`, and `purge` deletes `b.pdf`'s index entry **and its corpus + origin files**:
   ```rust
   let stale = store.list(false, 0, u32::MAX)?...filter(|id| !desired.contains(id))...;
   store.purge_many(stale);   // deletes id_b
   ```

**Net effect:** `b.pdf` is still referenced (`b.ref` is untouched, the object still exists) — a momentary read blip alone deleted its indexed document. It only comes back on a later clean resync, which re-downloads + re-translates it, and search silently misses it until then. The valve was meant to prevent exactly this, but it only catches the *all-failed* case, not one failure among several.

ljhh-0611 (commented):
Could we have these function name flow,

1. router/knowledge have resync() function that called spawn_resync() in Resyncer
2. spawn_resync() call resync() method in Resyncer struct

We use two resync() and use spawn_resync() between them.

It's just light suggestion, not that much severe, but  how can we rename resync() to another name?

ljhh-0611 (commented):
`spawn_resync` fires a fresh background task on every `/knowledge` mutation. The write-flush hook, plus each `rename`/`copy`/`remove_file` with no coalescing:

```rust
pub fn spawn_resync(&self, wid: Uuid) {
    tokio::spawn(async move { idx.resync(wid).await ... });  // unconditional
}
```

So one logical action touching several refs — a bulk WebDAV copy into `/knowledge`, a `PATCH /dirents` batch move, or a script creating N refs, fires N tasks. They queue on the per-workspace lock and each runs a *full* `resync()`: re-scan every ref, re-read every target, `list(0, u32::MAX)`, `compact`. The first N-1 produce the same end state as the last (pure redundant work), and each holds the store **write lock** from `store.write()` through `compact()`, so search on that workspace is blocked for the whole storm. The periodic sweep, when enabled, makes it worse by fanning `spawn_resync` across all workspaces at once.

Not a correctness bug, but I think that it would be user experience issue in real. The lock serializes everything and the final index is right; it's redundant work + a long lock hold.

Could we coalesce per workspace — if a resync is already in flight, mark it dirty and run once more at the end, instead of spawning a task per mutation?

backend/src/state/workspace/resync.rs
ljhh-0611: Fixed in ded60c6a.

**How.** The purge now keys off the *still-referenced* target set rather than "what ingested this cycle". Each workspace keeps a small per-target memo (`ref path → last indexed doc id`); on a read failure a still-referenced target is protected via its remembered id, so a transient blip no longer purges its doc. A removed ref is simply absent from the scan, so its doc is still purged — and, importantly, one unreadable/never-indexed ref no longer blocks the whole workspace's purge (so deleting a ref reliably takes effect even if an unrelated target is erroring).

The keep-set logic is extracted into a pure `build_keepset()` covered by regression tests: transient-blip protection, removed-ref purge, broken-ref-doesn't-block-others, ingest-failure keeps prior doc, content-change swaps the id, and a multi-cycle carry-forward. Also deduped overlapping targets (a dir ref + explicit file ref) and added a log on purge failures. (The full `reconcile` can't be driven end-to-end in tests yet since ingest needs an ailoy provider the backend doesn't wire — so the pure logic is what's covered.)

**Not handled here (follow-up).** The memo is in-memory, so it's lost on restart: a target that was indexed before a restart but is unreadable on the first post-restart cycle has no remembered id, and its doc could be purged. We kept this PR focused rather than fix that here. The robust fix is to persist the `(workspace, ref_path) → doc_id` mapping (a backend table), which is a schema change — so I'd rather open a separate issue and handle it in a follow-up PR. Will file it and link here.

backend/src/router/knowledge.rs
grf53: Since a solution proposal for this issue is being made in #209, I will resolve it here.

jhlee525 (approved):

PR #207, opened 2026-07-13
feat(session): titles, inline rename/delete, activity ordering, message dates

selloriwoo:
## Summary
End-to-end session title & management for app_v2: auto-generated titles that stream in live, inline rename and delete from both the chat header and the sidebar, activity-based Recents ordering, and per-message dates on hover.

## Features
- **Auto-titles** — generated from the first user turn (cheap model, 15s timeout, first-message fallback) and published over SSE so the sidebar/header fill in live; sanitized to one clean line and capped at 60 chars.
- **Rename** — inline edit from the chat header (click the title) and from each sidebar row's ⋯ menu; Enter/blur saves, Esc cancels, empty rejected, optimistic update with revert on failure; disabled while the title is still generating.
- **Delete** — from the sidebar ⋯ menu behind a confirm dialog; the row fades + collapses out, then is removed once the server delete confirms; deleting the open session navigates home.
- **Recents ordering** — sorted by last activity (`updated_at`, bumped on each run; title changes intentionally don't bump it) and refreshed on send so it reorders live.
- **Title UX** — shimmer skeleton while generating, then a one-shot typewriter reveal on first arrival (instant for existing sessions / reduced motion); near-limit character counter.
- **Message dates** — hovering a message reveals its date under the bubble, localized and date-only (ko "7월 17일", en "Jul 17"; tooltip adds the year), re-rendering on language toggle.
- **Browser tab** — the tab title reflects the current session (`{title} · Cowork`), updating live as the title streams in; the home / new-chat surface shows "New Chat · Cowork", so deleting the open session drops the tab back to a new-chat state.

## Backend (backend_v2)
- `PATCH /sessions/{id}` to rename (trim, reject empty, clamp 60, republish a `TitleEvent`); `SessionsState::set_title`.
- `TitleEvent` on the session channel; auto-titler in `SessionsState::run`.
- `MessageEvent`/`MessageResponse` and `list_messages(_since)` now carry `created_at`; the run loop stamps live events.
- `list_by_workspace` orders by `updated_at DESC`.

## Tests
**Frontend** (vitest, 108 pass)
- `SessionTitle`     — null→shimmer, ready→text, first arrival→typewriter reveal
- `messages`         — SSE `title` frame decodes into a title stream event
- `useMessageStream` — a `title` event patches the `['sessions']` query cache
- `transcript`       — each message's `created_at` flows through to the entry

**Backend** (cargo, 49 pass)
- `patch rename`     — trims the input and returns the updated title
- `patch blank`      — a whitespace-only title is rejected with 400
- `patch clamp`      — an over-long title is clamped to 60 characters
- `patch auth`       — a request without a bearer token gets 401
- `set_title`        — renaming a missing session returns NotFound
- `title sanitize`   — strips markdown and keeps only the first line

## Notes
- Also includes a cherry-picked IME-composition fix for the Composer — already on `app-v2-init`, so a no-op there.
- Message hover date currently shows for both user and assistant messages.

### DIFF
--- a/backend_v2/src/state/session.rs
+++ b/backend_v2/src/state/session.rs
@@ -129,11 +129,11 @@ impl SessionsState {
         self.data_root.join("sessions").join(id.to_string())
     }
 
-    /// Every session in `workspace_id`, oldest first.
+    /// Every session in `workspace_id`, newest first (for the Recents list).
     pub async fn list_by_workspace(&self, workspace_id: Uuid) -> StateResult<Vec<Session>> {
         let rows = sqlx::query(
             "SELECT id, workspace_id, agent_id, title, spec, runenv, created_at, updated_at \
-             FROM sessions WHERE workspace_id = ? ORDER BY created_at ASC",
+             FROM sessions WHERE workspace_id = ? ORDER BY created_at DESC",
         )
         .bind(workspace_id.to_string())
         .fetch_all(&self.db)
--- a/backend_v2/src/state/session.rs
+++ b/backend_v2/src/state/session.rs
@@ -269,15 +269,30 @@ impl SessionsState {
         let events = self.events.clone();
         tokio::spawn(async move {
             let title = crate::services::session_title::generate_session_title(&first_text).await;
+            // Gate on `title IS NULL`: the inline `needs_title` precheck ran
+            // before the (slow) LLM call, so a manual rename (`set_title`) may
+            // have landed since. Writing conditionally means we only ever fill
+            // an untitled session and never clobber a user's rename.
+            //
             // Like `set_title`, does not touch updated_at — titling shouldn't
             // reorder Recents.
-            if let Err(e) = sqlx::query("UPDATE sessions SET title = ? WHERE id = ?")
-                .bind(&title)
-                .bind(id.to_string())
-                .execute(&db)
-                .await
+            let affected = match sqlx::query(
+                "UPDATE sessions SET title = ? WHERE id = ? AND title IS NULL",
+            )
+            .bind(&title)
+            .bind(id.to_string())
+            .execute(&db)
+            .await
             {
-                tracing::warn!(session = %id, "failed to persist session title: {e}");
+                Ok(r) => r.rows_affected(),
+                Err(e) => {
+                    tracing::warn!(session = %id, "failed to persist session title: {e}");
+                    return;
+                }
+            };
+            // A rename beat us to it — leave their title in place and don't
+            // publish, or we'd broadcast a title the DB no longer holds.
+            if affected == 0 {
                 return;
             }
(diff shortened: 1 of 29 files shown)


ljhh-0611 (changes requested):
i18n Issue: the header rename UI and SessionTitle are hardcoded to Korean

Could you useTranslation for aria-label, not just korean hardcoded string in SessionTitle and $sessionId?

backend_v2/src/state/session.rs
```diff
@@ -212,6 +220,75 @@ impl SessionsState {
...
+            }
+        }
+
+        let db = self.db.clone();
+        let events = self.events.clone();
+        tokio::spawn(async move {
+            let title = crate::services::session_title::generate_session_title(&first_text).await;
+            // Like `set_title`, does not touch updated_at — titling shouldn't
+            // reorder Recents.
+            if let Err(e) = sqlx::query("UPDATE sessions SET title = ? WHERE id = ?")
```
ljhh-0611: The auto-titler writes unconditionally, so a title the user set later (via PATCH) can be silently overwritten. Gate it on title IS NULL so it only fills an untitled session and never clobbers a rename.

backend_v2/src/state/session.rs
```diff
@@ -212,6 +220,75 @@ impl SessionsState {
...
+            Err(e) => {
+                tracing::warn!(session = %id, "title precheck failed: {e}");
+                return;
+            }
+        }
+
+        let db = self.db.clone();
+        let events = self.events.clone();
+        tokio::spawn(async move {
+            let title = crate::services::session_title::generate_session_title(&first_text).await;
```
ljhh-0611, line 271: A markdown-only first message (e.g. "###") sanitizes to "", and the fallback path is empty too — so we can persist title = '', which the frontend keeps as an empty string (not null) and renders as a blank title bar. Skip the write when the result is empty, leaving the title NULL so the fallback/shimmer still works.

e.g. Below 271 line, we can just add like this condition
```rust
    if title.trim().is_empty() {
        return;
    }
```

backend_v2/src/state/session.rs
selloriwoo: Thank you! I fixed in [76811d8](https://github.com/brekkylab/agent-k/pull/207/commits/76811d879a71b3592afff28df38484175ab302ef)

backend_v2/src/state/session.rs
selloriwoo: If it's blank, then “NULL” is correct. I've corrected it. [b546409](https://github.com/brekkylab/agent-k/pull/207/commits/b546409232258e315ab1603ab9edce30aeac76e4)

ljhh-0611 (approved):
👍

PR #208, opened 2026-07-13
Backend rename

grf53:
## Description
`backend` -> `backend-v1`
`backend_v2` -> `backend`

## Note
This branch renames directories (`backend`→`backend-v1`, `backend_v2`→`backend`), so if main or other branches modified files under the old paths you'll hit **rename/edit conflicts** — always **keep the file at the new path (`backend`/`backend-v1`)** and re-apply the other side's changes on top. The most conflict-prone spots are the root `Cargo.toml` (keep `./backend` in members + `./backend-v1` excluded, never revive `backend_v2`) and `Cargo.lock` (on conflict, take `--theirs` then regenerate with `cargo build`).


nuri-yoo (approved):
Straightforward Migration and LGTM

PR #209, opened 2026-07-15
fix(workspace): enforce path containment in WorkspaceFs::resolve

selloriwoo:
## Description

`WorkspaceFs::resolve` joined a workspace-relative path onto the root without normalization, so a `rel_path` containing `..` climbed out of the workspace at the OS level (e.g. `/../../etc/passwd`, or `/../<other_wid>/files/...` into another workspace). `resolve` is the single choke point every local fs op passes through (`open` / `metadata` / `read` / `rename` / `copy` / `remove` / `readdir`), so the guard belongs here rather than in each protocol layer.

WebDAV happened to be safe because `DavPath` normalizes `..` before it reaches `resolve`, but that only protects the WebDAV caller. This hardens the shared primitive so containment holds for **every** caller (the concern raised in #204 / #201 / #206).

## Approach — clamp, not reject

`resolve` now folds the path lexically and clamps at the root instead of rejecting `..`. Rationale:

- **Doesn't break legitimate `..`** — `/a/b/../c` still resolves to `/a/c` (a blanket reject would `Forbidden` it, and the FUSE forwarder can forward a raw `..` lookup the kernel doesn't pre-resolve).
- **Keeps POSIX semantics** (`/..` == `/`) — it's path normalization, not a validity check that can fail.

## Testing

- `cargo test -p agent-k-backend` — green. Added `resolve_tests` covering normal paths, legitimate `..`, and a set of escape attempts (over-shoot `..`, cross-workspace, `/..`, empty) that all stay within the root.

## Notes

- Containment is **lexical**; it does not resolve symlinks. Acceptable because no symlink-creation op is exposed through this filesystem (WebDAV / ForwardFs / guest forwarder).

### DIFF
--- a/backend/src/state/workspace/fs.rs
+++ b/backend/src/state/workspace/fs.rs
@@ -12,8 +12,9 @@
 //! *every* caller of the filesystem — not just the WebDAV one — observes the
 //! same effects.
 
+use std::ffi::OsString;
 use std::io::{self, SeekFrom};
-use std::path::{Path, PathBuf};
+use std::path::{Component, Path, PathBuf};
 use std::pin::Pin;
 
 use bytes::{Buf, Bytes, BytesMut};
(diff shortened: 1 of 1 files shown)


backend/src/state/workspace/fs.rs
```diff
@@ -215,8 +216,32 @@ impl WorkspaceFs {
...
+    /// Containment is enforced here so it holds for every caller, not just the
+    /// protocol layers that happen to pre-normalise (WebDAV's `DavPath`). We fold
+    /// the relative path in isolation — pushing `Normal` segments and popping on
+    /// `..` — then append it to an untouched `self.root`. Because `..` only ever
+    /// pops from `parts` (never from the root) and a pop on an empty stack is a
+    /// no-op, a `..` can never climb above the workspace root: an escaping path
+    /// like `/../../etc/passwd` folds back to an in-root path rather than reaching
+    /// the host filesystem. This is lexical (POSIX-style `/..` == `/`); it does
+    /// not resolve symlinks, which are not creatable through this API.
     fn resolve(&self, rel_path: &str) -> PathBuf {
```
nuri-yoo, line 252: Now that `resolve` normalizes, `is_knowledge` (line 81) and the on-disk path can disagree on unnormalized input: `/knowledge/../secret.txt` classifies as knowledge but resolves to `root/secret.txt` (spurious event), while `/a/../knowledge/doc.txt` resolves into `knowledge/` but classifies as non-knowledge (missed ingestion). Latent today — the only in-tree caller (WebDAV) pre-normalizes and the handlers are stubs — but it breaks for exactly the future non-normalizing callers this PR defends against. Suggest extracting the fold into a helper and deriving both `resolve` and `is_knowledge` from the folded components. Fine as a follow-up; cheapest to fix before real ingestion lands.

nuri-yoo (approved):
LGTM. Reproduced the verification locally at c9a170c: `cargo test -p agent-k-backend` — 15 passed, including the two new `resolve_tests`. The clamp logic is sound: `..` only pops the folded stack, never the root, and `RootDir`/`Prefix` can't replace the root either.

One merge-ordering note: #204 moves this file to `workspace/src/fs.rs`, and `git merge-tree` reports a modify/delete conflict rather than following the move — so this fix must be ported into that branch before #204 merges (details on the #204 review). Merging this PR to main first is fine and doesn't depend on #204.

backend/src/state/workspace/fs.rs
selloriwoo: Thanks for catching this. it would have been a bug once ingestion lands.

nuri-yoo: confirmed the fix ([955cdf5](https://github.com/brekkylab/agent-k/pull/209/commits/955cdf514ec9d79592babbc45d43e04d09d22cd0))

PR #211, opened 2026-07-20
feat(models): port model catalog + GET /models to backend v2

nuri-yoo:
## Summary

Backend rename (#208) left the model catalog, the recommendation chains, and `GET /models` in `backend-v1`, which is excluded from the workspace. So v2 (`backend/`) shipped with no picker catalog and pinned every session to a hardcoded `anthropic/claude-sonnet-4-5`. This ports the catalog to v2 and refreshes it to the current flagship generation (as of 2026-07-20), so the composer picker works again and new sessions resolve a live model instead of a stale fixed default.

Frontend needs no change: `ComposerModelPicker` / `ProjectModelChainsEditor` / `api/models.ts` render entirely from the `GET /models` response, with no model ids of their own.

## Catalog

Added the current flagships that were missing:

- OpenAI: `gpt-5.6-sol` (max), `gpt-5.6-terra` (standard), `gpt-5.6-luna` (light)
- Anthropic: `claude-sonnet-5` (standard), `claude-fable-5` (max)
- Moonshot: `kimi-k3` (standard)

Kept alongside the existing `claude-opus-4-8`, `claude-haiku-4-5`, and Gemini 3.x entries. The recommendation chains were promoted to the new generation and stay provider-diverse (a single missing key never empties a chain).

Gemini deliberately keeps `gemini-3.5-flash` as its top text model: `gemini-3.5-pro` has slipped its launch repeatedly and has no public API model id yet, so there is nothing to catalog.

## Backend (v2)

- New `model/` module (`mod.rs` + `llm.rs`) mirroring the v1 layout: `ModelTier`, `AgentType`, `CATALOG`, `provider_available`, `resolve_model`/`resolve_model_in`, and the `GET /models` DTOs.
- `AgentType` is scoped to the two surfaces v2 drives (Coworker, DeepResearch). The v1 `Speedwagon`/`Buddy` chains, per-project chain overrides (`ProjectChains`/`validate_chains`), and the Speedwagon delegation helper are dropped — v2 has no session concept for them yet.
- `router/models.rs`: `GET /models` handler, registered under `auth_required`. It depends only on which provider keys are configured on the server, not on the caller.
- `session::build_spec` resolves the model through the agent-type chain instead of a fixed const, so an absent or unavailable `model` falls back to the first configured provider rather than a dead pin.
- `session::AgentType` now deserializes kebab-case (`deep-research`), matching the surface names `GET /models` advertises and what the frontend sends; `deep_research` stays accepted as a legacy alias. The snake_case-only wire format predates this PR (#208 carried it over), but porting the catalog is what made it reachable: the picker would recommend `deep-research` and `POST /sessions` would 422 on it.
- `provider_available` reads the pinned ailoy's `"default"` lang-model provider registry (`get_lm_providers`) — the old `default_provider()` accessor no longer exists after the ailoy AgentBuilder refactor.

Because ailoy resolves any `provider/*` id via a glob, a model does not have to be catalogued to run; the catalog only drives what the picker shows and what each agent recommends.

## Tests

`cargo test -p agent-k-backend` — 20 pass, including the 5 catalog tests and 2 session wire-format tests:

- `every_chain_entry_is_catalogued` — chains reference only catalogued ids
- `chains_are_provider_diverse` — no chain repeats a provider
- `agent_type_parses_canonical_values` — coworker / deep-research parse; speedwagon no longer does
- `resolve_always_returns_a_catalogued_model` — resolution never yields an empty/garbage id
- `catalog_response_annotates_and_recommends` — one recommendation per surface, each resolving to a catalogued id
- `agent_type_accepts_catalog_wire_values` — every surface name `GET /models` advertises parses back into a session `AgentType`
- `agent_type_accepts_legacy_snake_case_alias` — `deep_research` still accepted

Live smoke (one real one-turn call per catalogued model, not committed — hits provider APIs and costs tokens): **11 of 11 answered** as of 2026-07-21. The earlier run had `moonshotai/kimi-k3` failing on HTTP 429 `insufficient balance`; after the Moonshot account top-up it answers normally, confirming the id and wire adapter were fine all along.

## Notes

- GPT-5.6 answers through ailoy's OpenAI Responses adapter unchanged; the smoke test confirmed Sol/Terra/Luna all respond.
- Overlap with #204 (External source connection), which also targets `main` after the backend rename: it touches the same three files (`main.rs`, `router/mod.rs`, `router/session.rs`) but only add-only (a new `mod` declaration and a new `.api_route` line each), disjoint from the lines here. Whichever merges first, the other resolves by keeping both additions — no semantic conflict. The other open backend PRs (#200, #198, #194) predate the rename and will be rebased onto the current layout before they land, so they are not compared here; #209 touches only `state/workspace/fs.rs` and does not overlap.


backend/src/model/llm.rs
```diff
@@ -0,0 +1,336 @@
...
+    Standard,
+    Max,
+}
+
+/// Product-level agent surface that drives a session — selects the
+/// recommendation chain shown in the picker.
+///
+/// Mirrors the session-creation `AgentType` in [`crate::router::session`]; keep
+/// the two variant sets in sync.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Serialize, JsonSchema)]
```
grf53: I guess `PartialOrd` and `Ord` are added conventionally, but no use case are found and I can't get its necessity for `AgentType`.
I wonder there was a reason to add these and if not, it would good to be removed.

grf53 (approved):
Looks good to me.
The comment I wrote doesn't tell a blocking issue, but please take a look at it.

selloriwoo (approved):
It was ported successfully as intended
LGTM

backend/src/model/llm.rs
nuri-yoo: Addressed in latest commit [12864ff](https://github.com/brekkylab/agent-k/pull/211/commits/12864ffd33fbe728c08b0b103b61213692f87963)

PR #212, opened 2026-07-21
feat(vfs): add Gmail as a workspace source mount

selloriwoo:
## Summary
Adds **Gmail** as a workspace source: a mailbox mounted as files alongside S3/Notion, readable with plain filesystem tools (`ls`/`cat`/`find`/`grep`), searchable server-side through a virtual `.search/` dir, and deletable with `rm` (→ Trash). Bodies are rendered HTML→text for agents, everything fetched is cached across sessions on disk, and OAuth is handled server-side — the browser never sees the client secret, and each mount is identified by its account email.

```
/gmail/<label>/<yyyy>/<mm>/<subject>__<id>.gmail.json   # the email (JSON)
/gmail/<label>/<yyyy>/<mm>/<subject>__<id>/<filename>   # attachments (if any)
/gmail/.search/<query>/                                 # server-side search
```
---

Seeded from the [ailoy #435](https://github.com/brekkylab/ailoy/pull/435) draft; what ships here is described below.

---
### Listings: complete, small, timestamped
One id-scan per label builds a cached `date→ids` index — every `ls`/`find` under the label reuses it, so listings are complete and cost no per-directory API calls. Messages bucket into `yyyy/mm` (real INBOX: 643 flat date dirs → 5 year dirs), and every entry's mtime is the received time, so `find -newermt` and `ls -lt` just work. Per-label index caps at 5k messages as a quota guard; `.search` reaches the whole mailbox regardless.

### Reading: text bodies, lazy attachments
`cat` returns parsed mail — from/to/subject/labels plus an HTML→text body (scripts, styles, layout-table noise stripped; UTF-8 preserved). Attachments are real files in a sibling dir, listed from metadata and downloaded only when actually read; sender-controlled filenames are sanitized to single path segments.

### Search: `.search/<query>/`
`ls '/gmail/.search/from:boss refund newer_than:3m'` is one 5-unit API call; the newest 500 matches list as regular message files. Each search result is a first-class path — cat, attachments, stat, and rm all work on it. Search is an opt-in provider trait (`Resource::search()`): S3/Notion have no `.search` path at all, so agents are never taught a convention that fails there. The prompt forbids whole-mount greps and teaches this instead.

### Caching: three memory tiers + a cross-session disk cache
Memory: bounded message cache (10-min TTL — bodies are immutable), a global id→date map (cross-label dedup: 3,360 → ~1,200 fetches on a real mailbox), and a whole-attachment byte cache (the API has no ranged fetch; one 25 MB `cat` would otherwise re-download ~200×).
Disk: new provider-agnostic `vfs/disk_cache.rs` — immutable blobs + freshness-stamped snapshots, atomic writes, budget eviction. Bodies (2 GB) and attachments (2 GB, separate so one big file can't evict hundreds of bodies) persist per account; a new session adopts another session's label scan instead of redoing it. The cache key is the **account email**, resolved at mount-create — it survives re-consent, unlike anything token-derived.

### Fetching: batched, reconciled, rate-aware
`messages.get` goes out in 50-message batches, 3 in flight — sized to Gmail's ~250 units/sec pace. Retries reconcile by id: only what didn't come back is re-requested, so nothing is silently dropped. 429/5xx back off per Google's algorithm (16 s cap); the dormant `send` is excluded from 5xx retry (a 500 can arrive after the send queued — a blind retry could duplicate mail). Sequential scans read ahead exactly one month, riding idle batch slots.

### Writes: `rm` → Trash; send dormant
`rm` trashes the message from either route (label tree or search hit) and invalidates every cache that held it — body, attachment blobs, indexes, and all cached listings, so it vanishes immediately, including from the frontend's WebDAV view. The consent scope is the switch: `gmail.readonly` → 403, `gmail.modify` → active, no code change. `send`/`reply` scaffolding exists but is unwired and unadvertised.

### OAuth: server-side exchange, email identity
The frontend posts only `{ code, redirect_uri }`; the backend exchanges it with env-held client credentials and resolves the account email in the same step. The email is stored as a required config field and shown in `ProviderInfo::Gmail { email }` — the UI can finally tell mounts apart, and the secret never leaves the server.

---

## Testing
- **Unit (62):** parsers, sanitizers, date bucketing, batch-response parsing, backoff, cache bounds/eviction (memory + disk), `.search` routing, unlink gating, `process_message` shape.
- **Live, repeatable** (`#[ignore]`d, env-guarded): `.search/UMC` → 4 hits + `cat` in 1.4 s; a new session relists a 46-message month in **16 ms vs 9.9 s cold** (snapshot + blobs). Earlier full-mailbox pass: 14 labels incl. empty ones, KR/JP bodies, real `.hwp` attachment.
- **Benchmarks:** index build 10k ≈ 2.6 min; `grep -r` 17.0 → 4.6 min with read-ahead (measured on an earlier unbounded variant — the shipped design never blocks an `ls`, and `.search` is the intended path anyway).
- **In progress:** frontend consent → `create_mount` end-to-end.

## Config
**Backend env:** `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`.
**Mount create:** `{ "type": "gmail", "code": "…", "redirect_uri": "…" }`; consent = `access_type=offline` + `prompt=consent`, scope **`gmail.readonly`** or **`gmail.modify`** (enables `rm`).
**⚠️ Never include `gmail.metadata`** — its presence 403s `q=` and `format=full` even alongside broader scopes (verified live). One scope, exactly.


grf53 (approved):
Looks sufficiently good to me.

It seems there are quite a few design decisions, but I believe they were either discussed together or are sufficiently reasonable.

'disk cache' was interesting to me. I think it is a useful feature, but it seems there aren't many applications for it. When I briefly looked through Notion and S3, I didn't see any suitable uses other than Gmail. (Perhaps just large S3 items that have been fully read?)
I think there might be suitable targets among the other sources to add.

workspace/src/vfs/accessor/gmail.rs
```diff
@@ -0,0 +1,701 @@
...
+use std::time::{Duration, Instant};
+
+use base64::Engine;
+use futures::stream::{self, StreamExt};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use tokio::sync::Mutex;
+
+const OAUTH_TOKEN_URL: &str = "https://oauth2.googleapis.com/token";
+const GMAIL_API_BASE: &str = "https://gmail.googleapis.com/gmail/v1";
```
khj809: It seems the current code cannot override these base urls. Would be better to provide an option like `base_url` in `GmailConfig`, and check if it works against `https://enterprise-mock.brekkylab.com/gmail/v1` and `https://enterprise-mock.brekkylab.com/oauth2/token`.

selloriwoo: Added base_url to GmailConfig (default = Google. set per deployment via GOOGLE_API_BASE_URL env) 
verified the full flow against `https://enterprise-mock.brekkylab.com` with the new gmail_live_full_stack_probe test.

workspace/src/vfs/resource/gmail.rs
```diff
@@ -0,0 +1,1489 @@
...
+            let days = days_from_civil(y, m, d);
+            assert_eq!(civil_from_days(days), (y, m as u32, d as u32));
+        }
+    }
+
+    #[test]
+    /// An HTML-only mail's `href`s survive as `links`. The motivating case is
+    /// an over-25 MB file: Gmail replaces the attachment with a Drive URL, and
+    /// the rendered text keeps only the anchor's words.
+    #[test]
```
khj809: There's double `#[test]` here.
```suggestion
    /// An HTML-only mail's `href`s survive as `links`. The motivating case is
    /// an over-25 MB file: Gmail replaces the attachment with a Drive URL, and
    /// the rendered text keeps only the anchor's words.
    #[test]
```

workspace/src/vfs/resource/gmail.rs
```diff
@@ -0,0 +1,1489 @@
...
+            "ID42"
+        );
+        assert_eq!(id_from_name(&attach_dir_name("Hi there", "ID42")), "ID42");
+    }
+
+    fn live_config() -> Option<GmailConfig> {
+        Some(GmailConfig {
+            client_id: std::env::var("GMAIL_CLIENT_ID").ok()?,
+            client_secret: std::env::var("GMAIL_CLIENT_SECRET").ok()?,
+            refresh_token: std::env::var("GMAIL_REFRESH_TOKEN").ok()?,
```
khj809: Although this function is only used for gmail test purpose, the env var with `GOOGLE_` prefix is more widely used conventional name (mirage also uses it). Suggest to change them as below:
```suggestion
            client_id: std::env::var("GOOGLE_CLIENT_ID").ok()?,
            client_secret: std::env::var("GOOGLE_CLIENT_SECRET").ok()?,
            refresh_token: std::env::var("GOOGLE_REFRESH_TOKEN").ok()?,
```

khj809 (approved):
LGTM just resolve the conflicts plz

grf53 (approved):

PR #213, opened 2026-07-23
feat(backend): live token streaming 

selloriwoo:
Port the streaming backend from #200 onto backend v2. Backend only — the frontend from #200 predates the rewrite and is dropped.

Supersedes #200.

## What

- `agent.run()` → `agent.run_stream()` + `MessageAssembler` (ported from #200 with its unit tests): live assistant text is published as it streams, each message still persists at its boundary as before.
- Session stream (`GET /sessions/{id}/messages/stream`, SSE) carries a tagged `SessionEvent`:
  - `message` — persisted message with `seq` (same dedup/catch-up as before)
  - `delta` — ephemeral text fragment + `cum_len` (UTF-16, for client dedup across the catch-up↔live boundary)
  - `run` — lifecycle: `started` / `done` / `stopped` / `error` (errors were previously server-log only)
- SSE instead of WebSocket — the channel is strictly server→client. Auth goes through the standard `Authorization` header (no more token query param), and `message` events carry their `seq` as the SSE id, so `Last-Event-ID` auto-reconnect resumes from the persisted cursor.
- Attaching mid-turn replays the in-flight turn's partial text as one cumulative delta.
- Stop drains already-produced output (`now_or_never`), persists the partial answer with an interrupt note, and stubs unanswered tool calls so the next turn's history stays provider-valid.

No ailoy bump needed — main's rev already includes the stream contract #200 was built against.

## Verified

- Unit: assembler boundary/flush/sub-agent gating (7 tests).
- E2E against a live model: full stream round trip (delta concat == persisted message, contiguous seqs), auth gate (401 without header), `Last-Event-ID` resume with mid-turn snapshot, stop-mid-stream persistence, follow-up turn after stop, error propagation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1152,7 +1152,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
 dependencies = [
  "axum-core",
- "base64",
  "bytes",
  "form_urlencoded",
  "futures-util",
@@ -1172,10 +1171,8 @@ dependencies = [
  "serde_json",
  "serde_path_to_error",
  "serde_urlencoded",
- "sha1",
  "sync_wrapper",
  "tokio",
- "tokio-tungstenite",
  "tower",
  "tower-layer",
  "tower-service",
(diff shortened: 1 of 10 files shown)


grf53 (approved):
Thanks.

Most of the changes appear to have been made by moving #200 to fit the current situation, and this seems to have been done well.

Also, the #200 section has been sufficiently reviewed, and I could not find any additional points to address.

nuri-yoo (approved):
LGTM. One stale comment to fix before merge (non-blocking): `backend/src/auth/middleware.rs:26-27` still says the token-in-query endpoints are "WebDAV and the message WebSocket that bypass the middleware." This PR moves the message stream to SSE behind `auth_required`, so it's no longer token-in-query and no longer bypasses the middleware.

PR #214, opened 2026-07-23
feat(backend): record message depth and source agent

selloriwoo:
Record each message's agent-loop provenance so clients can tell sub-agent output apart from the top-level conversation.

## What

- `messages.content` now stores `{depth, source_agent, message}`; legacy bare-`Message` rows still parse (read as depth 0, no source), so no migration.
- `depth` / `source_agent` are exposed on `GET /sessions/{id}/messages` and the SSE `message` event — depth 0 is the conversation, depth ≥ 1 is sub-agent internals for the frontend to render indented/collapsed.
- History replay now feeds only depth-0 rows to the model, matching ailoy's in-memory semantics — a sub-agent's work reaches the model through its capped depth-0 tool result, not its transcript.
- Seq allocation follows the last persisted row (`MAX(seq)+1`) since the filtered history length no longer tracks the row count.

| | Persisted in DB | Model context (replay) |
|---|---|---|
| v1 | depth 0 only | depth 0 only |
| v2 (main) | everything, depth dropped | everything (pollutes context) |
| this PR | everything, depth preserved | depth 0 only |

## Verified

- Unit: stored-format round trip + legacy fallback.
- E2E with a real sub-agent: depth/source stamped on persisted rows and events (depth: count`{0: 4, 1: 1}`) , follow-up turn accepted by the provider, seq continues past filtered rows; old-format DB reads back as depth 0.

### DIFF
--- a/backend/src/event.rs
+++ b/backend/src/event.rs
@@ -77,7 +77,17 @@ pub enum SessionEvent {
     /// A message persisted to the session's history at sequence `seq`. Mirrors
     /// the HTTP `GET /sessions/{id}/messages` item shape so catch-up over
     /// either transport is interchangeable on the client.
-    Message { seq: i64, message: Message },
+    Message {
+        seq: i64,
+        /// Nesting level relative to the top-level agent turn: `0` is the
+        /// conversation the user sees; `>= 1` is a sub-agent's internal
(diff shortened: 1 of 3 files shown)


grf53 (approved):
The change looks uncomplicated and clear for its goal.

PR #215, opened 2026-07-24
feat(fuse): Serve workspace VFS into the sandbox over a raw-FUSE tunnel

khj809:
Replaces the in-guest HTTP forwarder with a **raw-FUSE tunnel**: the guest relays its `/dev/fuse` traffic byte-for-byte over one persistent TCP connection, and the host runs the actual FUSE engine against the unified `WorkspaceFs`.

## Architecture
| piece | where | role |
|---|---|---|
| guest pump | `crates/ailoy-vfs-tunnel` (renamed from `ailoy-vfs-forwarder`) | mounts `/dev/fuse`, relays raw FUSE bytes. No protocol conversion, no `fuser`/`libfuse`. **Self-daemonizing.** |
| host engine | `workspace::TunnelServer` (`workspace/src/vfs/sandbox/tunnel.rs`) | `fuser::Session::from_fd` (no mount) over a `SOCK_DGRAM` socketpair, answering ops against `ForwardFs` |
| guest injection | `backend::sandbox_tunnel` | embeds the cross-built pump ELF, launches it; `session.rs` mounts via it |

Mirrors the old split (host `forward.rs` / guest `sandbox_fs.rs`) that this removes.

## Why
The old forwarder did FUSE↔HTTP conversion on both ends and opened a **new TCP connection per op**. On the same S3 file it was slower and unreliable (short/failed reads). The tunnel reads reliably; with page-cache **readahead** (the default) it pipelines reads — measured **~340 MB/s vs ~100 MB/s** for serial `direct_io`, and no worse on Notion (the workspace VFS resolves render-on-read sizes, so cached reads aren't clamped short).

## Setup simplification
The pump **self-daemonizes**: it creates the mountpoint, clears any stale mount, mounts, then forks to the background and signals readiness — so the host runs a **single `exec`** that returns exactly when mounted (no init script, no `/proc/mounts` polling, no `setsid` wrapper). Failures surface with a non-zero exit + diagnostic.

## Notes
- **Depends on the `fuser` fork** ([brekkylab/fuser](https://github.com/brekkylab/fuser/tree/feat/macfuse-abi), branch `feat/macfuse-abi`, pinned by rev in the root `[patch.crates-io]`): on macOS upstream `fuser` emits macFUSE wire structs the Linux guest kernel rejects with `EINVAL`; the fork gates that behind an off-by-default `macfuse-abi` feature. No-op on Linux hosts.
- `backend` is now **lib + bin** so integration tests can drive internals; in-VM checks live in `backend/tests/tunnel_e2e.rs` (all `#[ignore]` — they boot a microsandbox VM).
- Removed: in-guest HTTP forwarder binary, host `ForwardServer`, `backend/src/sandbox_fs.rs`, `workspace/tests/forward_http.rs`.

## Testing
- `cargo test -p workspace host_engine_serves_over_tcp` — in-process engine (INIT/GETATTR/LOOKUP), no VM.
- `backend/tests/tunnel_e2e.rs::{tunnel_debug, s3_read_throughput, notion_read_full_content}` — boot a VM + read S3/Notion through the tunnel (needs creds + microsandbox; verified locally).


jhlee525 (approved):
👍

PR #217, opened 2026-07-28
Restore Automation

grf53:
## Description

Port backend-v1's automation feature into the new backend.

An automation is a workspace-scoped job that "creates a session and runs **a single prompt**."
v1's multi-step `prompts` array is collapsed to one prompt — there is no per-step cursor, so a re-claimed run just re-runs its one prompt on a fresh session.

## Self-describing runs

A run snapshots the rendered prompt and agent surface it executes at creation time. It therefore depends on no automation at run time, and `automation_id` is nullable with `ON DELETE SET NULL`, so editing or deleting an automation preserves its runs as an audit record.

## Webhook auth (ported from v1, minus idempotency)

The webhook token is issued once on create and only its sha256 hash is stored (`webhook_token_hash`, globally unique). The raw token is surfaced only in the create response as `webhook_token`. Firing is a public `POST /webhooks/automations` with `Authorization: Bearer <token>`, where the hash identifies the trigger globally. Idempotency is intentionally deferred to the event layer.

## Changes

- **Migration `0003_automations.sql`** — `automations`, `automation_triggers` (+ `webhook_token_hash` column and partial unique index), self-describing `automation_runs`, append-only `automation_run_logs` (9 kinds, no per-step).
- **`state/automation.rs`** — automations CRUD, cron triggers, run lifecycle (claim/lease/finalize/cancel/retry/reap), RunLog audit log, webhook token-hash lookup.
- **`cron.rs`** — 5-field POSIX cron next-fire calculation.
- **`worker.rs`** — claim/execute workers, expired-lease reaper, cron ticker, per-run lease heartbeat with bounded retry.
- **`state/session.rs`** — `drive_prompt`: drives one turn to completion with a cancel token (unlike the fire-and-forget `run()`) so the worker can finalize.
- **`router/`** — automations CRUD + nested cron triggers; top-level `/automation-runs` (create from an automation or ad-hoc one-time, filtered list, get, cancel, logs); public `/webhooks/automations` receiver returning the token once.
- **`lib.rs`** — spawn the automation runtime at startup.

## Tests
`cargo build` / `clippy --all-targets` clean; `cargo test -p agent-k-backend` **63 passed / 0 failed**.


backend/src/state/session.rs
```diff
@@ -799,6 +799,143 @@ impl SessionsState {
...
+        };
+
+        let rows =
+            sqlx::query("SELECT content FROM messages WHERE session_id = ? ORDER BY seq ASC")
+                .bind(&session_key)
+                .fetch_all(&self.db)
+                .await?;
+        let history: Vec<Message> = rows
+            .iter()
+            .map(|r| serde_json::from_str::<Message>(&r.get::<String, _>("content")))
```
nuri-yoo: This decode can never succeed on a non-empty session. `persist_message` above writes each row in the `SessionMessage` wrapper shape (`{"depth":0,"source_agent":null,"message":{...}}`), and ailoy's `Message` requires `role`/`contents`, so reading the wrapper back as a bare `Message` fails with `missing field 'role'`. I verified with a test at this head — it passes, i.e. the exact bytes `persist_message` writes fail this decode:

```rust
#[test]
fn drive_prompt_history_decode_fails_on_wrapped_rows() {
    let stored = SessionMessage {
        depth: 0,
        source_agent: None,
        message: Message::new(Role::User).with_contents([Part::text("do it")]),
    };
    let content = serde_json::to_string(&stored).unwrap();
    let direct = serde_json::from_str::<Message>(&content);
    assert!(direct.is_err());
    assert!(direct.unwrap_err().to_string().contains("missing field"));
}
```

First execution passes because the session is empty, but three paths reach this with rows present, and each one burns a retry attempt on the serde error: (a) lease expiry, reaper requeue, re-claim; (b) restart, where the boot reap requeues every `running` row; (c) no crash at all: the run's session is visible in `GET /sessions` from creation, so if anyone posts a message to a deferred run's session before the worker claims it (which works fine, since `run()` reads the wrapper correctly), the claim then fails immediately. With `MAX_ATTEMPTS = 3`, two re-claims exhaust the run.

Fixing the decode alone is only half of it: on re-claim, `drive_prompt` also re-persists the user prompt, so the same prompt appears twice in the replayed history. Since the PR description already states the intended semantics ("a re-claimed run just re-runs its one prompt on a fresh session"), the simplest coherent fix seems to be clearing the session's messages at (re-)execution start. The alternative is continuation semantics: reuse `parse_stored_message` + the depth-0 filter + `last seq + 1` like `run()` does, and skip re-inserting the prompt when it's already the last user message. Either is fine by me; what's there now does neither.

The module note "a re-claimed run simply re-runs its one prompt on the (message-less) session" in `worker.rs` needs the same correction: the session is not message-less at re-claim time.

backend/src/state/automation.rs
```diff
@@ -0,0 +1,1483 @@
...
+
+/// Total attempts including the first; RETRY_BACKOFFS has MAX_ATTEMPTS-1 gaps.
+pub const MAX_ATTEMPTS: i64 = 3;
+
+/// Build a session [`AgentSpec`] from an automation's stored agent surface.
+/// Unknown `agent_type` falls back to the coworker preset.
+pub fn build_spec(agent_type: &str, model: Option<&str>) -> AgentSpec {
+    use agent_k::agents::{get_coworker_agent_spec, get_deep_research_agent_spec};
+    match agent_type {
+        "deep_research" => get_deep_research_agent_spec(
```
nuri-yoo: Two divergences from both v1 and the v2 session API.

`"deep_research"` is a third spelling. v1 used kebab-case everywhere, including its automation repository (`backend-v1/src/repository/automation.rs:21`: `coworker | speedwagon | deep-research | buddy`; `backend-v1/src/model/llm.rs:45`), and the v2 session API uses the typed `AgentType` with `"deep-research"`, whose comment explicitly asks to keep the variant sets in sync. As written, a client that reuses the session API's value gets the coworker preset silently (the `_` arm), and `create_automation` accepts any string without validation, so a typo also degrades silently. Reusing `model::AgentType` for the column and validating at create would close both.

The default model pins `anthropic/claude-sonnet-4-5`, which in this codebase appears only in tests; session creation resolves through `resolve_model` and the per-type catalog chain, falling back by provider availability. A deployment without an Anthropic key runs sessions fine but fails every automation run. Building the spec through the same `resolve_model` path (at run-creation time, consistent with the snapshot semantics) would keep the two surfaces aligned.

backend/src/state/automation.rs
```diff
@@ -0,0 +1,1483 @@
...
+            agent_type,
+            model,
+            previous_run_id: None,
+        };
+        self.insert_run_with_session(spec, at, Some(&payload))
+            .await?
+            .ok_or_else(|| StateError::InvalidData("failed to create run".into()))
+    }
+
+    /// Cron fire: create a run and advance the trigger's `next_fire_at` in one tx.
```
nuri-yoo: The comment says "create a run and advance the trigger's `next_fire_at` in one tx", but `insert_run_with_session` commits its own transaction and the trigger UPDATE runs separately after it. If the process dies between the two, or the UPDATE merely fails transiently (the error propagates and the trigger stays due), the next tick fires the same instant again and creates a duplicate run. Smallest fix: flip the order (advance `next_fire_at` first, then insert), which turns the failure mode from "duplicate run" into "one missed run"; for automation that seems the safer direction. Merging into one transaction works too but needs `insert_run_with_session` to accept a caller transaction. At minimum the comment should match the code.

backend/src/worker.rs
```diff
@@ -0,0 +1,312 @@
...
+        .append_log(run.id, RunLogKind::Started, None)
+        .await
+        .map_err(|e| e.to_string())?;
+
+    // The run is self-describing: it carries the rendered prompt it executes, so
+    // no automation lookup is needed (and ad-hoc runs have no automation).
+    let parts = vec![Part::text(run.prompt.clone())];
+    state
+        .sessions
+        .drive_prompt(run.session_id, parts, cancel.clone())
```
nuri-yoo, line 309: `drive_prompt` doesn't register in the sessions runs map, so the "is this session running?" gate that protects `run()` doesn't see worker-driven turns. A user posting to the run's session (it's listed in `GET /sessions`) drives it concurrently with the worker, both compute their own next seq, and the `(session_id, seq)` PK fails whichever inserts second; the user's turn errors out, or the run fails and retries. Registering the worker's drive in the runs map would make the user's concurrent `run()` get the existing 409 `AlreadyRunning`, and it also closes path (c) of the history-decode issue in `state/session.rs` for the queued-run window.

backend/migrations/0003_automations.sql
```diff
@@ -0,0 +1,84 @@
...
+-- A run is self-describing: it snapshots the rendered prompt + agent surface it
+-- executes, so it depends on no automation at run time and its audit trail
+-- survives later edits/deletion of the automation.
+CREATE TABLE automation_runs (
+    id              TEXT PRIMARY KEY,
+    -- NULL for ad-hoc one-time runs (they depend on no automation). ON DELETE
+    -- SET NULL so deleting an automation preserves its runs as an audit record.
+    automation_id   TEXT REFERENCES automations(id) ON DELETE SET NULL,
+    trigger_id      TEXT REFERENCES automation_triggers(id) ON DELETE SET NULL,
+    session_id      TEXT NOT NULL UNIQUE REFERENCES sessions(id) ON DELETE CASCADE,
```
nuri-yoo, line 61: Question on intent: `DELETE /sessions/{id}` exists, and this cascade means deleting a run's session deletes the run and its logs, so the audit-preservation property this PR establishes for automation deletion (SET NULL + snapshot) doesn't hold against session deletion, and run sessions are ordinary entries in the session list. Deleting mid-run is also unguarded: the deletion choreography in `state/mod.rs` notes that a row cascade never cancels an in-flight run, and the worker's drive isn't in the runs map either, so the worker just hits an FK error on its next message insert and the failure isn't retried (the run row is gone). If user-initiated session deletion is meant to take the run's audit trail with it, that's a defensible call, but then it's worth a comment here; otherwise refusing to delete a session with a non-terminal run would be the minimal guard. Could also be follow-up material rather than this PR.

backend/src/state/session.rs
```diff
@@ -799,6 +799,143 @@ impl SessionsState {
...
+    /// Persist one message at `seq` and publish it on the session channel.
+    async fn persist_message(
+        &self,
+        session_key: &str,
+        seq: i64,
+        message: &Message,
+        channel: &str,
+    ) -> StateResult<()> {
+        let stored = SessionMessage {
+            depth: 0,
```
nuri-yoo, line 861: Latent today; flagging because the fix is one line. `agent.run()` yields `MessageOutput` with `depth`/`source_agent`, and this persists everything as depth 0 / no source, unlike `run()` which stores `output.depth.unwrap_or(0)` and replays only depth-0 rows. Right now neither preset (`coworker`, `deep_research`) has sub-agent tools, so depth >= 1 never actually occurs and there is no observable difference. The moment a sub-agent tool lands in a preset, automation sessions will record sub-agent internals as top-level conversation and replay them into the model history on the next user turn. Passing the output's own `depth`/`source_agent` through now costs nothing.

backend/src/state/session.rs
```diff
@@ -799,6 +799,143 @@ impl SessionsState {
...
+            .bind(&session_key)
+            .fetch_optional(&self.db)
+            .await?
+            .ok_or(StateError::NotFound)?;
+        let spec: AgentSpec = serde_json::from_str(&row.get::<String, _>("spec"))?;
+        let has_runenv: bool = row.get("runenv");
+
+        let dir = self.data_root.join("sessions").join(&session_key);
+        let archive_path = dir.join("sandbox.tar.zst");
+        let runenv = if has_runenv {
```
nuri-yoo: Also latent today. This branch is unreachable: automation sessions are inserted with `runenv = 0`, nothing updates `sessions.runenv` after creation, and the worker is `drive_prompt`'s only caller. But it's ~40 lines that quietly diverge from `run()` when it ever does execute: plain `try_from_archive` without the network policy re-application (`try_from_archive_with_network(..., Public)`) and without the workspace VFS tunnel mount, so the sandbox would come up with no guest-to-host egress and no `/mnt/workspace`. I'd replace the branch with `if has_runenv { return Err(...) }` so the divergence can't silently activate; supporting runenv properly can come with whatever feature first needs it.

backend/src/state/session.rs
grf53: Now we wipe the history when we start an automation run.
Comments are revised along with that.

backend/src/state/automation.rs
grf53: applied.

backend/src/state/automation.rs
grf53: ### Changes
- `fire_cron_trigger()` now executes a single transaction including `next_fire_at` advance and run insertion (using `insert_run_in()`).
- `list_due_cron_triggers()` now checks the `automation` is disabled(like-v1). So the logic in the latter part is slightly simplified.

backend/src/worker.rs
grf53: I believe it is appropriate to prohibit additional message POSTs to automation sessions.

I have utilized the `origin` field from v1 for that, which determines from the session row itself whether a specific session was created by a user or automation.

Additionally, we have added an `origin` filter based on the same field to the session list API.

backend/migrations/0003_automations.sql
grf53: I don't think we need to keep runs on deletion of corresponding session, but it could be inconsistent with former decision that prohibits message POSTs to automation sessions.
So I think it is good to prohibit DELETE the automation sessions by user directly, but allow users to delete automation runs(and the corresponding sessions would be deleted along with those).

backend/src/state/session.rs
grf53: Since this part is not related to automation, I think it would be better to address it as a separate issue.

backend/src/state/session.rs
grf53: I considered to support runenv now, but its backend could be revised soon, so I just accepted your suggestion to return error explicitly. Thanks.

backend/migrations/0003_automations.sql
nuri-yoo: The run-as-unit direction looks right, and the 409 guard covers queued/running. One gap is left in the exact flow the guard's comment recommends (cancel first, then delete): `cancel_run` is a bare UPDATE (`status='cancelled', lease_until=NULL`) and nothing delivers the cancellation to the worker. `drive_prompt` never registers in the sessions runs map — the map is touched only at session.rs:352, :419, :423, :845, all in or for `run()` — so the `cancel()` inside `SessionsState::remove` finds nothing, and the worker's only signal is its next heartbeat `renew_lease` returning false, up to `HEARTBEAT_INTERVAL` = 30s later. Since `cancelled` is terminal, `delete_run`'s `matches!(status, Queued | Running)` passes immediately, so the session row is deleted under the still-driving worker: its next `persist_message` hits the FK violation, and the run row is already gone, so `finalize_run` returns false and nothing records the failure. Nothing calls the new endpoint yet, but the UI already has cancel wired up, so the first delete button placed next to it makes cancel-then-delete the natural gesture. Repro at this head (passes):

```rust
// probe_insert_message = the INSERT persist_message runs, minus the event publish
let claimed = state.claim_due_run(now, now + Duration::minutes(3)).await?.expect("claimable");
probe_insert_message(&pool, &run.session_id.to_string(), 0).await.expect("first write ok");
assert!(state.cancel_run(run.id, &json!({"source":"manual"})).await?);          // bare UPDATE
assert!(!state.renew_lease(run.id, Utc::now() + Duration::minutes(3)).await?);  // only signal, polled at 30s
assert!(!matches!(state.get_run(run.id).await?.unwrap().status, RunStatus::Queued | RunStatus::Running));
sqlx::query("DELETE FROM sessions WHERE id = ?").bind(run.session_id.to_string()).execute(&pool).await?;
let err = probe_insert_message(&pool, &run.session_id.to_string(), 1).await.expect_err("must fail");
assert!(format!("{err}").to_lowercase().contains("foreign key"));
assert!(state.get_run(run.id).await?.is_none());
```

A status check can't close it from the delete side: `cancel_run` also nulls `lease_until`, so the DB has no marker separating "cancelled, worker unaware" from "cancelled, worker stopped". Registering the worker's drive in the runs map is the clean fix — then `remove()`'s cancel reaches the in-flight agent and the window shrinks to the same millisecond-scale race `remove()` already documents and accepts for user sessions. Alternatively `cancel_run` could leave the lease for the worker to clear on acknowledgment, with `delete_run` requiring a released lease.

backend/src/state/automation.rs
```diff
@@ -0,0 +1,1614 @@
...
+
+    /// Cron triggers whose `next_fire_at` has elapsed (enabled only).
+    pub async fn list_due_cron_triggers(
+        &self,
+        now: DateTime<Utc>,
+    ) -> StateResult<Vec<AutomationTrigger>> {
+        let sql = format!(
+            "SELECT {TRIGGER_COLS} FROM automation_triggers \
+             WHERE kind = 'cron' AND enabled = 1 AND next_fire_at IS NOT NULL AND next_fire_at <= ? \
+               AND automation_id IN (SELECT id FROM automations WHERE enabled = 1) \
```
nuri-yoo, line 904: One semantics change worth confirming: previously a disabled automation's triggers kept advancing `next_fire_at` (the deleted comment "a disabled automation still shouldn't re-fire the same instant forever" was about this), but now the due scan skips them, so `next_fire_at` freezes while disabled. If the automation stays off past a scheduled slot, re-enabling surfaces the trigger still due at that old instant and it fires at once with a stale `scheduled_for` — verified with a test at this head: trigger due 6h ago, disable, re-enable, `list_due_cron_triggers` returns it at the 6h-old instant and `fire_cron_trigger` creates the run carrying it. Exactly one catch-up fire, since the next advance is computed from `now`. If fire-once-on-re-enable is intended, a comment would settle it; if not, NULL-on-disable plus recompute-on-enable restores the previous behavior and also makes the migration comment ("`next_fire_at` is … NULL when the trigger is disabled") true, which nothing currently implements. Same treatment would cover trigger-level disable whenever that lands — the frontend already calls `PATCH /automations/{id}/triggers/{trigger_id}` for an enabled toggle, which has no handler yet.

grf53: That was intentional, but I missed to apply guards for that.
Setting NULL for disable ones would be good safety guard, so I applied that too.
Applied recompute-on-enable & NULL-on-disable.

nuri-yoo (commented):
Checked ce9653e0 + 3916ce47 — the disable/re-enable handling matches what v1 did, and covering the creation path too (`payload.enabled && automation.enabled`) closes the case I hadn't raised. `cargo test -p agent-k-backend --lib` is 68 passed / 0 failed and clippy is clean at 3916ce47.

Timing note so the thread reads straight: these two commits were already written when I posted my previous comment, and landed right after it, so the `next_fire_at` half of that comment crossed with the fix rather than preceded it.

Two things on the new code below, plus the cancel/delete item from the last round is still open — I re-ran that repro at this head and it still passes; neither commit touches `cancel_run`, `drive_prompt`, or `delete_run`.

backend/src/state/automation.rs
```diff
@@ -0,0 +1,1773 @@
...
+    }
+
+    /// Cron triggers whose `next_fire_at` has elapsed (enabled only).
+    pub async fn list_due_cron_triggers(
+        &self,
+        now: DateTime<Utc>,
+    ) -> StateResult<Vec<AutomationTrigger>> {
+        let sql = format!(
+            "SELECT {TRIGGER_COLS} FROM automation_triggers \
+             WHERE kind = 'cron' AND enabled = 1 AND next_fire_at IS NOT NULL AND next_fire_at <= ? \
```
nuri-yoo, line 903: Dropping `automation_id IN (SELECT id FROM automations WHERE enabled = 1)` makes `next_fire_at IS NOT NULL` the sole encoding of "schedulable", and three places now have to maintain that invariant — `update_automation` (inside its tx), `set_trigger_enabled` (reads `automation.enabled`, then a separate UPDATE), and `router/automation.rs:379` (reads `automation.enabled` from `require_owned_automation`, then a separate INSERT). The last two read and write outside a transaction, so a `PATCH /automations/{id}` that disables while either is in flight leaves `enabled = 1` with `next_fire_at` set under a disabled automation.

That state used to be harmless because this subquery filtered it. Now it becomes permanent: the ticker picks the trigger up, `fire_cron_trigger` correctly refuses the run, but the refusal path rolls the whole tx back (automation.rs:1068-1070), which restores the due `next_fire_at` — so the next tick sees the same trigger again, 15s later, forever, and nothing self-heals until someone toggles the automation. Repro at this head (passes): disable the automation, insert an enabled cron trigger with a past `next_fire_at`, then loop `list_due_cron_triggers` + `fire_cron_trigger` three times — each round returns the trigger, creates no run, and leaves `next_fire_at` unchanged.

Cheapest fix is to keep the subquery as a backstop; `idx_triggers_due` still carries the scan and the invariant stays the fast path. Alternatively, have the disabled branch of `fire_cron_trigger` NULL `next_fire_at` instead of rolling it back, which repairs the row on first contact.

backend/src/state/automation.rs
```diff
@@ -0,0 +1,1773 @@
...
+            "SELECT {AUTOMATION_COLS} FROM automations WHERE workspace_id = ? ORDER BY created_at DESC"
+        );
+        let rows = sqlx::query(sqlx::AssertSqlSafe(sql))
+            .bind(workspace_id.to_string())
+            .fetch_all(&self.db)
+            .await?;
+        rows.iter().map(Automation::from_row).collect()
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
nuri-yoo: This `#[allow]` was `update_automation`'s; inserting the helper in front of it left it on a three-argument function (and above its doc comment). `update_automation` got its own copy, so this one can just go.

backend/migrations/0003_automations.sql
nuri-yoo: Still reproduces at 3916ce47 — neither of the two new commits touches `cancel_run`, `drive_prompt`, or `delete_run`, so cancel still reaches the worker only through the next `renew_lease`, and `delete_run`'s guard still passes the moment the status flips to `cancelled`.

backend/src/state/automation.rs
grf53: Did both.
`list_due_cron_triggers` gets the `automation_id IN (enabled)` backstop again so stale rows never reach the fire path (`idx_triggers_due` still carries the scan), and `fire_cron_trigger`'s disabled branch now NULLs `next_fire_at` instead of rolling it back, so any row that does reach it self-heals on first contact (re-enable recomputes it).

backend/migrations/0003_automations.sql
grf53: Went with registering the worker's drive. Thanks.

nuri-yoo (approved):
Verified all three at 41c8b14:

- Compile: `f50cee91` merges main cleanly, and `cargo check --all-targets` is error-free — the sqlx 0.9 `SqlSafeStr` sites are wrapped with `AssertSqlSafe` (constants/literals only, never user input; every dynamic value stays bound).
- Cancel-then-delete: `drive_prompt` now registers in the sessions runs map for its duration and `cancel_run` signals it directly, so the FK-crash window is closed.
- Cron re-enable: `list_due_cron_triggers` has the `enabled` backstop back, and `fire_cron_trigger`'s disabled branch NULLs `next_fire_at` and self-heals instead of leaving a stale row; covered by a new test.

Full suite: 71 passed / 0 failed.

LGTM, approving.

PR #219, opened 2026-07-28
chore(deps): modernize stable dependencies across stacks

nuri-yoo:
## Description

This is the one-off dependency modernization tracked by #216. It updates the Rust workspace, Ailoy, Docling's Python environment, and the web application together so their migrations and cross-stack compatibility are reviewed in one place.

The PR keeps every existing dependency declaration, including dependencies without a current Rust call site. Stable releases are preferred throughout. The only prerelease exception is wreq: after publishing 5.3.0, upstream has continued publishing the active client line as successive v6.0 release candidates rather than cutting another stable 5.x or 6.0 release. The selected RCs are pinned exactly and remove the reachable `lru 0.13` advisory.

## Ailoy bump and prerequisite

The first dependency change remains the Ailoy bump:

| Dependency | Old | New | Included upstream work | Verification |
| --- | --- | --- | --- | --- |
| Ailoy | `e3f9091` | `0767a08` | brekkylab/ailoy#441: Schemars 0.9 migration, microsandbox 0.6.7, remaining stable refresh, Clippy cleanup, and exact wreq v6 RC migration | Ailoy check/clippy/test; RustSec scan; old archive → new runtime load/boot/exec; production archive/restore E2E |

brekkylab/ailoy#441 merged on 2026-07-29, squashed onto Ailoy `develop` as `0767a08`, and the pin now names that commit. It previously pointed at `53039e79`, the pre-squash branch head, which is reachable only from the merged feature branch and is not an ancestor of `develop`. The swap is content-identical (`git diff 53039e79 origin/develop` is empty), so only the provenance changes: the pin no longer depends on a branch head that could be deleted. This removes the reason the description previously gave for holding the PR back.

### [`a48fde6`](https://github.com/brekkylab/agent-k/commit/a48fde6b1d5ab26a587c8604e9dcd37f69e863e9) — chore(deps): modernize Rust dependency stack

Updates the direct Rust dependency set and performs the source migrations required by the new APIs:

- replaces dynamic SQLx query construction with static SQL selected through a typed session-owner enum;
- migrates Tantivy indexing and search to the 0.26 document/value and `TopDocs` APIs;
- rebuilds an incompatible Tantivy index while preserving its source corpus;
- selects jsonwebtoken 11's `aws_lc_rs` provider explicitly;
- updates reqwest's rustls feature configuration;
- adds optimized development-profile hashing for microsandbox archive/restore;
- pins the accompanying Ailoy migration.

#### Rust dependency audit

| Dependency | Old | New | Migration notes | Verification |
| --- | --- | --- | --- | --- |
| Ailoy | `e3f9091` | `0767a08` | Pins brekkylab/ailoy#441 as merged onto `develop`, including microsandbox 0.6.7, Schemars 0.9, stable refreshes, and the wreq RC migration. | Workspace check/clippy/test; cross-version and production archive/restore E2E |
| aide | `0.14` | `0.15.1` | Migrates OpenAPI integration together with Schemars 0.9. | Backend compile and 52 backend tests |
| schemars | `0.8` | `0.9` | Uses Aide 0.15's compatible line and changes the feature to `chrono04`. | Schema-bearing routes compile; workspace tests |
| sqlx | `0.8.6` | `0.9` | Splits runtime/TLS features and replaces dynamic `SELECT` strings with static typed branches. | Backend state tests |
| tantivy | `0.22` | `0.26` | Adopts `TopDocs::order_by_score`, the value trait API, and new string extraction. Rebuilds old indexes from the preserved corpus. | Store tests and real 0.22 → 0.26 rebuild probe |
| reqwest | `0.12` | `0.13` | Changes `rustls-tls` to reqwest 0.13's `rustls` feature. | Workspace check/clippy/test |
| jsonwebtoken | `9` | `11` | Enables `aws_lc_rs` explicitly to prevent runtime provider initialization failure. | Workspace tests |
| axum-extra | `0.10` | `0.12` | Declaration retained; no active call-site migration required. | Backend compile and tests |
| croner | `2` | `3` | Declaration retained; no active call-site migration required. | Workspace compile and tests |
| nanoid | `0.4` | `0.5` | Declaration retained; no active call-site migration required. | Workspace compile and tests |
| sha2 | `0.10` | `0.11` | Declaration retained; development hashing is optimized for the microsandbox archive chain. | Archive/restore E2E |
| tower-http | `0.6` | `0.7` | CORS feature retained; no call-site change required. | Backend compile and tests |
| html-to-markdown-rs | `3.4.0-rc.2` | `3.9.2` | Replaces the old RC with the current stable line. | HTML translation and workspace tests |
| strum | `0.27` | `0.28` | Keeps derive support on the current stable line. | Agent-k compile and tests |
| kreuzberg | `4.9.8` | `4.10.2` | Keeps bundled PDFium integration and the exact pin. | Parser and workspace tests |

### [`24b9bd3`](https://github.com/brekkylab/agent-k/commit/24b9bd3d28037ef0afdc5600a2802c6ee48f61aa) — chore(deps): refresh Docling Python stack

Refreshes the complete uv lock graph instead of changing only the top-level package. The resulting environment contains 114 locked stable packages and clears every vulnerability found in the pre-update export.

#### Python dependency audit

| Dependency | Old | New | Migration notes | Verification |
| --- | --- | --- | --- | --- |
| docling | `2.99.0` | `2.115.0` | Top-level document conversion runtime upgrade. | `uv sync`; real `payslips.pdf` conversion |
| docling-core | `2.79.0` | `2.88.0` | Updated in Docling's compatible resolved graph. | Real PDF conversion |
| docling-ibm-models | `3.13.2` | `3.13.3` | Compatible model refresh. | Environment sync and conversion |
| torch | `2.11.0` | `2.13.0` | Retains CPU wheels for supported platforms. | Environment sync and conversion |
| torchvision | `0.26.0` | `0.28.0` | Updated in lockstep with torch. | Environment sync and conversion |
| transformers | `5.6.2` | `5.8.1` on macOS / `5.14.1` elsewhere | Resolves stable platform-specific versions required by the graph. | Marker resolution and conversion |
| opencv-python | `4.13.0.92` | `5.0.0.93` | Major transitive image-processing update through Docling. | Real PDF conversion |
| complete uv graph | 112 entries | 114 entries | Full stable lock refresh; no direct dependency removed. | Exported lock audit: zero known vulnerabilities |

### [`2a1f136`](https://github.com/brekkylab/agent-k/commit/2a1f13639af7444fefd4ceb2b61b515c83461acb) — chore(deps): migrate web stack to stable releases

Updates the application and test/build toolchain together. PDF rendering moves from the react-pdf component API to PDF.js 6's direct `getDocument` / `PDFPageView` API without removing the `react-pdf` declaration.

The migration adds explicit worker setup, text and annotation layers, cancellation/cleanup behavior, focused PDF preview coverage, and regenerated TanStack route output. Failure handling landed later, in `a3db9e5`.

#### Web dependency audit

| Dependency | Old | New | Migration notes | Verification |
| --- | --- | --- | --- | --- |
| `@tanstack/react-query` | `5.62.7` | `5.101.4` | Stable query runtime refresh. | TypeScript lint, 177 tests, build |
| `@tanstack/react-router` | `1.95.0` | `1.170.18` | Updated with regenerated route tree. | Lint and build |
| `@tanstack/router-plugin` | `1.95.0` | `1.168.23` | Kept compatible with the router runtime. | Build |
| `dayjs` | `1.11.13` | `1.11.21` | API-compatible refresh. | Tests and build |
| `i18next` | `26.3.0` | `26.3.6` | Current metadata supports TypeScript 7 directly; no peer override remains. | `pnpm peers check`, lint, tests |
| `papaparse` | `5.5.3` | `5.5.4` | Patch update; declaration retained. | Tests |
| `pdfjs-dist` | `5.4.296` | `6.1.200` | Migrates preview rendering to direct PDF.js 6 APIs, layers, worker, and cleanup lifecycle. | Focused PDF tests, full suite, build |
| `react` / `react-dom` | `19.2.6` | `19.2.8` | Runtime and types updated together. | Tests and build |
| `react-i18next` | `17.0.8` | `17.0.11` | Current metadata supports TypeScript 7 directly; no peer override remains. | Peer check, lint, tests |
| `zustand` | `5.0.2` | `5.0.14` | Stable state-store refresh. | Tests |
| `@playwright/test` | `1.60.0` | `1.62.0` | Stable browser-test tooling refresh. | Resolution and build |
| `@testing-library/react` | `16.1.0` | `16.3.2` | Updates the renderer used by PDF preview tests. | 177 tests |
| `@types/node` | `22.10.5` | `24.13.3` | Matches the selected Node 24 runtime line. | TypeScript lint and build |
| `@types/react` | `19.2.14` | `19.2.17` | Patch type refresh. | TypeScript lint |
| `@vitejs/plugin-react` | `6.0.1` | `6.0.4` | Stable plugin refresh. | Build |
| `jsdom` | `25.0.1` | `30.0.0` | Final bump is enabled by pinning Node 24.18. | 177 tests and build under Node 24.18 |
| `typescript` | `5.7.2` | `7.0.2` | Stable TypeScript 7 migration; current i18next peers accept it directly. | Peer check and `tsc -b --noEmit` |
| `vite` | `8.0.12` | `8.1.5` | Stable build-tool refresh. | Production build |
| `vitest` | `2.1.8` | `4.1.10` | Major test-runner migration. | 21 files / 177 tests |
| Node toolchain | unpinned; local 24.13 | `24.18.0` | Adds `.node-version` and declares `>=24.15.0 <25` in the app manifest. | Frozen install, tests, lint, build |

`react-pdf` remains declared at `10.4.1`.

### [`6e5705b`](https://github.com/brekkylab/agent-k/commit/6e5705befbacc90d41448e1ac768d9976dae837b) — fix(build): use bundled rust-lld for VFS tunnel

The backend build script cross-compiles the VFS tunnel pump for a Linux-musl guest. rustup provides `rust-lld` inside the active toolchain rather than installing a system-wide `lld`.

This commit:

- reads the active sysroot from `rustc --print sysroot`;
- locates the host-specific bundled `rust-lld`;
- passes its absolute path through `CARGO_ENCODED_RUSTFLAGS`;
- falls back to `lld` if the bundled binary cannot be found;
- isolates the nested Cargo build from outer flags and wrappers.

### [`fcd9f47`](https://github.com/brekkylab/agent-k/commit/fcd9f47a265810108b7bea604f6116b3e30f4cb8) — chore: resolve current stable clippy warnings

Resolves the pre-existing workspace warning set exposed by the final `-D warnings` gate:

- simplifies nested conditionals and error handling;
- removes needless borrows and redundant closures;
- updates path arguments and parser code to current idioms;
- silences `unexpected_cfgs` for the existing `internal` cfg (first as a Cargo feature, corrected to `check-cfg` in `b46dac0`);
- retains dormant Speedwagon cases with targeted `#[expect]` attributes.

No test case or dependency is removed.

### [`569d83e`](https://github.com/brekkylab/agent-k/commit/569d83ef4ebb5442b5f2ac470bc96d43d78e9cd1) — chore(deps): adopt wreq v6 RC via Ailoy

- repins Ailoy from `c2163c4` to `6aca191`;
- brings in exact `wreq =6.0.0-rc.29` and `wreq-util =3.0.0-rc.14` pins;
- preserves charset, compression, streaming, and Firefox emulation behavior through explicit features;
- removes wreq's active `lru 0.13` path and resolves `lru 0.18.1`;
- reduces the RustSec result from 1 vulnerability plus 4 warnings to 1 vulnerability plus 3 warnings.

### [`df20090`](https://github.com/brekkylab/agent-k/commit/df20090c6579ba6fc7243a04a39d92a3e2a1fdf2) — chore(deps): update jsdom for Node 24

- adds a repository `.node-version` pin for Node 24.18.0;
- declares the supported Node range as `>=24.15.0 <25`;
- updates jsdom from 29.1.1 to 30.0.0;
- removes the obsolete i18next/react-i18next TypeScript 7 peer overrides;
- refreshes the lockfile and retains `@types/node` on the matching 24.x line.

### [`1acddfe`](https://github.com/brekkylab/agent-k/commit/1acddfe352fad8d306360b688ed1bc40557e4b83) — chore(deps): repin ailoy to brekkylab/ailoy#441 review head

- moves the pin from `6aca191` to `53039e7`, picking up two review fixes on that PR: a declared `rust-version` of 1.95.0, and re-enabling wreq's `system-proxy` feature;
- the proxy fix matters here. wreq 6 moved macOS and Windows OS-level proxy detection behind a `system-proxy` feature absent from its default set, so the v6 migration had silently dropped it. Neither Ailoy nor this workspace ever configures a proxy explicitly, so `web_fetch` and `web_search` would have failed with plain connection errors for anyone whose proxy is set only through the OS rather than `HTTP_PROXY`;
- the `lru` line is a label change, not a version move: `msb_krun_devices` and `dav-server` both require the 0.16 range and were already resolved to 0.16.4, which is past the `>= 0.16.3` patch for `RUSTSEC-2026-0002`.

### [`b46dac0`](https://github.com/brekkylab/agent-k/commit/b46dac0b63b4a33d9c3488e2311b6cb20c84adbf) — fix(build): stop declaring internal as a Cargo feature

`a48fde6` declared the pre-existing `cfg(feature = "internal")` as a real Cargo feature to silence `unexpected_cfgs`. That made `--all-features` enable the FinanceBench harness in `knowledge_base/preset.rs` and two gated tests, all of which import `knowledge-base-examples`: a crate that lives outside this repository and is neither vendored, submoduled, nor published. So `cargo check --all-features` could not build agent-k at all.

Declaring the dependency does not fix it either. The crate is only reachable as a path dependency on a sibling checkout, and Cargo resolves path dependencies at manifest load regardless of which features are enabled, so adding it breaks plain `cargo check` for any clone that lacks that sibling directory. It also pulls `aws-config` and `aws-sdk-s3` into the graph, taking the lockfile from 901 to 962 packages and adding duplicate `reqwest`, `dirs`, and `http` versions.

This commit drops the feature declaration so the harness returns to the dormant state it has on `main`, and tells rustc the cfg name is expected via `check-cfg`. The harness stays opt-in through `RUSTFLAGS` where the crate is available. Making it properly buildable needs a git dependency or a submodule, which is a packaging decision for its own change rather than this dependency refresh.

### [`a563660`](https://github.com/brekkylab/agent-k/commit/a563660) — chore(deps): repin Ailoy to the merged Schemars migration

- moves the Ailoy pin from `53039e79`, the pre-squash head of the merged `deps/migrate-schemars-0.9` branch, to `0767a08` on Ailoy `develop`;
- content-identical swap (`git diff 53039e79 origin/develop` is empty), so the resolved graph does not move;
- only `Cargo.toml` and one `Cargo.lock` source line change.

### [`96ca0f6`](https://github.com/brekkylab/agent-k/commit/96ca0f6) — build: pin the Rust toolchain and declare the workspace MSRV

Raised in review: the Ailoy repin raises the MSRV and nothing in the repo records it. `e3f9091` declared no `rust-version`; `0767a08` declares 1.95.0 and Ailoy inherits it. sqlx 0.9 independently requires 1.94.0. On an older toolchain the workspace fails at resolve time before anything compiles, and `cargo update` cannot route around the Ailoy half because that pin names a rev with no alternative version to select.

- adds `rust-toolchain.toml` with `channel = "1.97.1"` — the stable this branch is verified on — plus `components` and `targets`;
- `targets` is not optional. rustup installs targets per toolchain, so a freshly pinned toolchain starts without the musl std that `backend/build.rs` needs to cross-compile the in-guest VFS tunnel pump. That path degrades silently: the sub-build fails with `E0463`, `build.rs` writes a zero-byte stub, and the only signal is a `cargo:warning`, leaving in-VM mounts unavailable at runtime behind a green build. Both arches are listed because `build.rs` derives the guest target from the build's own arch;
- adds `rust-version = "1.95.0"` on `[workspace.package]`, inherited by the four member crates, so an unsupported toolchain names agent-k rather than only a transitive git dependency;
- `Cargo.lock` does not move.

The exact channel pin needs a periodic bump, since `1.97.1` goes stale and every contributor fetches that specific toolchain.

### [`b562c2d`](https://github.com/brekkylab/agent-k/commit/b562c2d) — build: add .nvmrc so nvm honors the Node pin

`.node-version` is read by fnm, nodenv, and asdf, but nvm looks for `.nvmrc` and finds nothing, so the Node pin `df20090` added did nothing for nvm users. That matters because jsdom 30 requires `^22.22.2 || ^24.15.0 || >=26.0.0`: on an out-of-range Node the tests fail with an error that points at jsdom rather than at the Node version.

`engines` stays advisory. pnpm 11 prints the same `Unsupported engine` warning with or without `engine-strict=true` and installs either way, so setting it would imply a gate it does not provide.

### [`a3db9e5`](https://github.com/brekkylab/agent-k/commit/a3db9e5) — fix(app): surface PDF preview failures instead of a blank stage

The PDF.js 6 migration in `2a1f136` caught both failure paths with `console.error` only. A document that failed to load rendered an empty stage, which a user cannot tell apart from a blank or still-loading PDF, and a single page that failed to draw left an unlabelled gap.

- load failures replace the stage with the existing `preview.error_title` / `preview.error_body` copy, already present in both locales for `FallbackCard`, and hide the zoom controls, which do nothing with an empty stage;
- a page that fails to draw labels its own slot and leaves the rest of the document alone;
- PDF.js owns the container node passed to `PDFPageView` and the effect cleanup calls `replaceChildren()` on it, so React must not render children there. The per-page label sits under a wrapper React owns, one level up, and the box-shadow rule moves from a child to a descendant selector to match;
- renames the `document` state and the `PdfPage` prop to `pdfDoc`. Naming them `document` shadowed the DOM global for the whole component body, so any DOM call added to either function later would have silently resolved to a PDF proxy;
- adds two tests, taking the suite from 177 to 179.

### [`4fb7123`](https://github.com/brekkylab/agent-k/commit/4fb7123) — chore(app): pin packageManager to the pnpm this branch was verified with

`.node-version` and `.nvmrc` pin Node, but pnpm stayed floating, so the lockfile this branch regenerated could be re-resolved by whichever pnpm a contributor happened to have. This adds `"packageManager": "pnpm@11.3.0"` — the version the frozen install, tests, lint, and build were run with — next to the existing `engines.node` range.

## Deliberately constrained or retained

| Dependency | Selected | Newer candidate | Reason |
| --- | --- | --- | --- |
| schemars | `0.9` | `1.2` | Aide 0.15 uses Schemars 0.9. Wait for stable Aide 0.16 rather than introducing the alpha line or splitting the schema graph. |
| wreq | exact v6 RC | unreleased 6.0 stable | Explicit narrow exception: since 5.3.0, upstream releases have continued on the v6.0 RC line; stable 5.3 retains the reachable `lru` advisory fixed on that maintained line. |
| `@types/node` | `24.13.3` | `26.x` | Kept aligned with the selected Node 24 runtime. |
| react-pdf | `10.4.1` | unchanged | Retained under the no-removal policy while the application calls PDF.js 6 directly. |
| fuser | patched `0.18` | unchanged | The macFUSE ABI patch used by the raw-FUSE tunnel is still required. |

No alpha, beta, nightly, or canary package is introduced. The wreq family is the only release-candidate exception and is pinned exactly.

## Upstream follow-up

Both upstream migrations for the two remaining warnings have now merged. Neither has reached this lockfile yet, and the two are no longer at the same distance from it.

| Finding | Upstream PR | Migration | Upstream status | What still stands between it and this lockfile |
| --- | --- | --- | --- | --- |
| `RUSTSEC-2025-0141` — `bincode 2.0.1` is unmaintained | [superradcompany/imago#3](https://github.com/superradcompany/imago/pull/3) | Selectively backports the production changes from canonical [`hreitz/imago!57`](https://gitlab.com/hreitz/imago/-/merge_requests/57), replacing bincode's fixed-int big-endian derives with the upstream `on_disk_struct` codec. No new QCOW2 version or image conversion is required. | Merged 2026-08-02 as `21199f6b` and released in `msb-imago 0.1.3`; `0.1.4` is current and carries one further QCOW2 fix. Neither version declares bincode. | Nothing outside this repository. `msb_krun_devices 0.1.25` requires `imago ^0.1.0`, so the move is a lockfile-only `cargo update -p msb-imago` here: `0.1.1 -> 0.1.4`, dropping `bincode`, `bincode_derive`, `unty`, and `virtue`. This branch has not taken it yet, so the warning still stands in the after-scan below. |
| `RUSTSEC-2025-0134` — `rustls-pemfile 2.2.0` is unmaintained | [superradcompany/microsandbox#1228](https://github.com/superradcompany/microsandbox/pull/1228) | Replaces the archived wrapper in microsandbox image and network certificate loading with `rustls-pki-types 1.14.1` and `CertificateDer::pem_slice_iter`, preserving the existing error policies. | Merged 2026-07-31 as `e895d823`; microsandbox `main` no longer resolves `rustls-pemfile` at all. | A microsandbox release. `rustls-pemfile` enters through `microsandbox-image` and `microsandbox-network`, which declare it in their own manifests, so no lockfile update reaches it. The newest published line is `0.6.8`, tagged 2026-07-29 and therefore without `e895d823`. Once a release carries it, Ailoy must bump its `0.6.7` pin and agent-k must repin Ailoy. |

The imago fix has not landed in microsandbox either — its `main` lockfile still resolves `msb-imago 0.1.1` — but that does not gate agent-k, which resolves the crate through its own lockfile.

Both moves change the after-scan numbers below, so neither is folded into this PR silently. The `msb-imago` bump is available now and costs a lockfile change plus a re-run of the Rust audit; the microsandbox one waits on a release either way.

## Security audit

The before and after scans use the same advisory data and audit commands against `origin/main` and this PR's lockfiles. Lockfile findings do not by themselves imply runtime reachability.

### Before and after

| Ecosystem | Before this PR (`origin/main`) | After this PR | Change |
| --- | --- | --- | --- |
| Rust / RustSec | 932 locked packages; 10 vulnerability findings; 12 warning entries (7 unmaintained, 4 unsound, 1 yanked) | 901 locked packages; 1 vulnerability finding; 3 warning entries, all unmaintained | Removes 9 vulnerability findings and 9 warning entries; no unsound or yanked package remains |
| Python / `pip-audit` | 31 known vulnerabilities across 7 packages | 0 known vulnerabilities | Resolves all 31 findings |
| Web / `pnpm audit` | 11 findings: 1 critical, 5 high, 4 moderate, 1 low | 0 known vulnerabilities | Resolves all lockfile findings |

#### Findings present before this PR

| Ecosystem | Affected packages before the update |
| --- | --- |
| Rust vulnerabilities | `astral-tokio-tar 0.6.1`, `crossbeam-epoch 0.9.18`, `hickory-proto 0.26.0-alpha.1` (2 advisories), `lopdf 0.40.0`, `quick-xml 0.40.1` (2 advisories), `quinn-proto 0.11.14`, `rmcp 0.11.0`, `rsa 0.9.10` |
| Rust warnings | Unmaintained: `bincode`, `fxhash`, `instant`, `paste`, `proc-macro-error2`, `rustls-pemfile`, `ttf-parser`; unsound: `anyhow`, `lru 0.12`, `lru 0.13`, `memmap2`; yanked: `spin 0.9.8` |
| Python vulnerabilities | `idna 3.13` (2), `pillow 12.2.0` (20), `pydantic-settings 2.14.0`, `setuptools 81.0.0` (2), `soupsieve 2.8.3` (2), `torch 2.11.0`, `urllib3 2.6.3` (3) |
| Web findings | `@babel/core`, `esbuild`, `form-data`, `postcss`, `vite`, `vitest`, and `ws` |

### Remaining Rust findings

| Finding | Reachability | Disposition |
| --- | --- | --- |
| `bincode 2.0.1` — unmaintained | Reaches the graph only through `msb-imago 0.1.1` | Removed upstream by [superradcompany/imago#3](https://github.com/superradcompany/imago/pull/3), merged and released in `msb-imago 0.1.3` / `0.1.4`. Clearable here by a lockfile-only bump, which this branch has not taken |
| `rustls-pemfile 2.2` — unmaintained | Active through microsandbox image/network TLS handling | Removed upstream by [superradcompany/microsandbox#1228](https://github.com/superradcompany/microsandbox/pull/1228), merged 2026-07-31 but not in any published release through `0.6.8`; remains until a release carrying it is adopted by Ailoy and then agent-k |
| `paste 1.0.15` — unmaintained | Lock-only optional package; unreachable in the active all-features workspace graph | Upstream repository is archived |
| `rsa 0.9.10` — timing-side-channel advisory | Lock-only optional package; unreachable in the active all-features workspace graph | No patched stable release; upstream fix remains in progress |

## Verification

- Ailoy:
  - `cargo check --all-features --all-targets`
  - `cargo clippy --all-features --all-targets -- -D warnings`
  - `cargo test --all-features` — 327 passed, 0 failed, 17 ignored
  - doctests — 11 passed, 0 failed, 7 ignored
  - RustSec — 1 vulnerability, 3 unmaintained warnings; no `lru` advisory
- agent-k Rust workspace, on the pinned 1.97.1 toolchain:
  - `cargo check --workspace --all-targets`
  - `cargo check --workspace --all-features --all-targets`
  - `cargo clippy --workspace --all-targets -- -D warnings`
  - `cargo test --workspace --all-targets`
  - agent-k: 80 passed, 1 ignored
  - backend: 52 passed
  - workspace: 31 passed, 2 ignored
  - workspace integration: 4 passed
- migration/E2E:
  - real Tantivy 0.22 → 0.26 rebuild with corpus preservation
  - microsandbox 0.6.1 archive → 0.6.7 load, boot, and exec
  - production archive/restore E2E on microsandbox 0.6.7
- Web under Node 24.18.0:
  - frozen `pnpm install`
  - `pnpm peers check`
  - `pnpm lint`
  - `pnpm test:unit` — 21 files / 179 tests
  - `pnpm build`
  - `pnpm audit` — no known vulnerabilities
- Python:
  - `uv sync`
  - real Docling conversion of `payslips.pdf`
  - exported uv lock audit — no known vulnerabilities

## Upgrade note

Tantivy 0.26 cannot open a 0.22 index. `indexer::open_or_create` detects that, removes the unreadable directory, and creates a fresh index; the `corpus/` and `origin/` trees are untouched, so nothing has to be re-translated. But `Store::new` does not re-index on its own, and the rebuild only happens when `KnowledgeIndexes::reconcile` next runs — on a write under `/files/knowledge`, on the explicit resync route, or on the `AGENT_K_INDEX_INTERVAL_SECS` periodic task, which is off unless set.

So on an existing deployment, knowledge search returns zero results between the upgrade and the first resync. Trigger one resync per workspace after deploying.

Closes #216


selloriwoo (approved):

app/package.json
```diff
@@ -12,36 +16,36 @@
...
-    "zustand": "^5.0.2"
+    "zustand": "^5.0.14"
   },
   "devDependencies": {
-    "@playwright/test": "^1.60.0",
-    "@tanstack/router-plugin": "^1.95.0",
-    "@testing-library/react": "^16.1.0",
-    "@types/node": "^22.10.5",
+    "@playwright/test": "^1.62.0",
+    "@tanstack/router-plugin": "^1.168.23",
```
khj809, line 39: Using the pinned `@tanstack/router-plugin@1.168.23`, it always generate `routeTree.gen.ts` different from current branch's one, which makes the working tree always dirty.
I confirmed that the hash of generated `routeTree.gen.ts` is deterministically same as main branch's one, so I suspect that the current `routeTree.gen.ts` must have been generated with different version of `@tanstack/router-plugin`.
Could you check by clean-installing node modules and running router plugin again?

nuri-yoo: Re-verified in a fully clean setup (fresh worktree at 4fb7123, fresh pnpm store, `--frozen-lockfile`, plugin confirmed on disk as 1.168.23): regeneration reproduces the committed file byte-for-byte and the tree stays clean. With 1.167.35, which main's lockfile resolves `^1.95.0` to, regeneration reproduces main's file exactly, so the generator's ordering changed between those two releases. Could your node_modules still be the one installed on main? `verify-deps-before-run=false` in `app/.npmrc` lets vite run on a stale install after a branch switch, so `pnpm install` on this branch plus a `pnpm ls @tanstack/router-plugin` check would confirm.

khj809: Below is my output of your command, and it seems the plugin was installed correctly.
```
Legend: production dependency, optional only, dev only

cowork-app@0.1.0 /Users/haejoonkim/Desktop/workspace/BrekkyLab/agent-k/app (PRIVATE)
│
│   devDependencies:
└── @tanstack/router-plugin@1.168.23

1 package
```
But still getting the different `routeTree.gen.ts`.. I'll try to figure out why.

khj809: I found the reason. It was due to my stale vite dev server process, which was running from few weeks ago.
The already running process didn't catch the reinstalled node_modules, and keeps generating `routeTree.gen.ts` with older generator.
Sorry for the confusion!

nuri-yoo: Thanks for tracking it down. BTW I keep my Mac running for weeks without a restart too, so the similar thing happens to me sometimes.

khj809 (approved):

PR #221, opened 2026-07-30
fix(sandbox): loopback-bind the VFS tunnel, time-box docling, scope guest host reach

nuri-yoo:
Host-side half of the sandbox-boundary hardening from an internal security review. The ailoy half is brekkylab/ailoy#443, merged 2026-08-07.

**Now merged with `main`; base is `main`.** This started stacked on #219 (dependency modernization), which landed 2026-08-03, and #212 (Gmail workspace source mount) landed while the branch was open. `33a4660` merges both in, so the diff against `main` is this PR's own changes again.

Where the two halves overlapped, the resolution goes to `main`: its squashed #219 revision of the PDF preview and the pnpm/packageManager pin supersede this branch's earlier copies of the same work, which leaves `app/` here identical to `main`. The two real merges are in the VFS tunnel. `backend/src/state/session.rs` keeps this branch's spawn-then-grant-then-attach order — the tunnel port has to be known before the sandbox's network policy is fixed — and picks up main's `workspace_prompt` append and `GUEST_MOUNT_ROOT`. `backend/tests/tunnel_e2e.rs` keeps both sides' new tests, with main's `agent_finds_the_mount_unprompted` moved off the removed `mount_vfs_tunnel_in_guest` and onto the two-step API.

## What the review found still open on `main`

The findings were written against `feat/v2-model-catalog`. `main` is past #215, which replaced the HTTP forward server with the raw-FUSE tunnel, so two of the five items no longer exist and one is narrower than it was:

| Finding | State on `main` | Here |
| --- | --- | --- |
| Forward server on the wildcard address, non-constant-time token | Alive in the tunnel server | Fixed |
| Parser has no timeout, leaves orphans | Alive, on both PDF engines | Fixed, docling in `5fb431c` and Kreuzberg — the one the server actually uses — in `ddb5526` |
| Guest-writable bind mount of the user's input | Alive in the two CLI binaries; the backend binds no host directory | Fixed, measured in a guest |
| `.cmd/` route acting with host-held credentials | Gone. The route was removed with the HTTP forward server. `Resource::command` is still implemented but nothing reaches it; `grep -c cmd` over the VFS write path is 0 | Not touched |
| Guest reaches every host port | Fixed in Ailoy; needs a call-site change here | Done, see below |

`.cmd/` is worth one note even though it is unreachable: `workspace/src/vfs/resource/notion.rs` still instructs the model to `echo … > .cmd/page-create`, so the agent is being told to use a route that no longer answers. Fixing that instruction, or restoring the route behind a per-op allowlist, is a separate change.

## Tunnel server: loopback bind, constant-time token

`TunnelServer` bound the wildcard address. Everything it serves, the whole unified workspace tree with provider mounts included, sat behind one token on a port every peer on the same LAN could reach. The comparison was a plain `!=` on strings, and a rejected connection was logged at `debug` next to ordinary disconnects.

It now binds loopback. **The review's premise here was wrong** and worth correcting: it said a loopback-only bind was not an option because the guest arrives via the gateway address. It is an option, because microsandbox runs a user-space network stack: the guest dials `host.microsandbox.internal:<port>` and the host side of that connection is opened by the microsandbox process itself. Measured with a loopback-only listener and a guest `nc` against the gateway address:

```
guest:  172.16.4.61  host.microsandbox.internal    →  GUEST_CONNECT_OK
host:   accepted, peer = 127.0.0.1:51923
```

So the guest still reaches it and the LAN no longer can. That is a better outcome than binding the gateway interface, which would not have worked anyway from a server started before the sandbox exists.

The token is compared with `subtle::ConstantTimeEq`, and a rejection logs at `warn` with the peer address — the only client that should reach this port is the guest pump, which is handed the token directly, so anything else dialing it is worth seeing.

The handshake is also time-boxed. Every connection costs an OS thread and the token read had no deadline, so a client that connected and then said nothing held its thread for the life of the process; loopback narrowed who can do that to local processes and the guest, but did not bound it. The deadline is cleared before the relay, because a mounted tunnel idles between requests and the clone taken for the read direction shares this socket's options.

Tests: the listener's address must be loopback, a wrong token of the same length gets no FUSE reply, the comparison accepts only exact equality, and a client that connects without presenting a token is closed rather than held — observed as EOF from the client side, which without the deadline is instead a client-side timeout.

## docling: time-boxed, and the kill reaches the parser's children

`wait_with_output` had no deadline, so a document that made the parser spin held the calling worker for as long as the host process lived. The parser runs on the host with the worker's own privileges, and its input comes from outside (uploaded, or fetched from the web).

The run is now wrapped in a timeout (300 s default). The write to stdin is inside it too, not just the wait: a parser that never drains stdin fills the pipe buffer and blocks the write instead, which is the same hang through a different door.

On expiry the whole process group is killed, not just the child. The command gets its own group and the timeout path SIGKILLs it; `kill_on_drop` stays as a backstop. This matters because docling shells out (OCR helpers, for one), so killing only the process we spawned leaves those running.

The future owning the child is pinned to a named local so it is visibly still alive in the timeout arm. Passing it to `timeout` directly also works, but only because a temporary in a match scrutinee outlives the arms — an invisible rule that binding the result with `let` first would quietly remove. (What actually keeps `killpg` from reaching an unrelated group is narrower than this description originally claimed; see `8ba588f`.)

`run_bundle` takes the executable path instead of resolving the bundle, so the timeout path is exercisable against a stand-in parser. The regression test uses one that starts a child and hangs: the call must fail on its own deadline, and the recorded grandchild must be gone afterwards. Confirmed to discriminate by re-running it with the process group removed, where the grandchild survives:

```
parser's child 30657 survived the timeout
```

Memory and CPU ceilings are deliberately not added. `RLIMIT_AS` interacts badly with a PyInstaller bundle's mmap-heavy startup, and the honest answer for resource isolation is running docling in its own sandbox, a larger change than this.

### The engine the server actually uses

All of the above is the docling path, and nothing on the server reaches it — raised in review, and correct. `PdfEngine` derives `Default` with `Kreuzberg` as the default variant (`translator/mod.rs:13-17`), and every caller in the workspace passes `PdfEngine::default()`: knowledge ingest at `backend/src/state/knowledge/mod.rs:300`, plus `knowledge_base/mod.rs:514`, `:548`, and `preset.rs:36`. Nothing constructs `PdfEngine::Docling` at all — the only `from_str_opt` callers live in `backend-v1`, which the root manifest excludes from the workspace — so the Docling arm of `translate_pdf` is unreachable from the running backend.

That path was calling `ExtractionConfig::default()`, whose `extraction_timeout_secs` Kreuzberg documents as "`None` means no timeout (unbounded extraction time)". `ddb5526` sets the same budget the docling path uses, so one number covers both. Verified against the `payslips.pdf` fixture that the field is honoured in this build: the default config extracts fine, and a zero-second budget returns `Extraction timed out after 1ms (limit: 0ms)`.

The budget releases the caller, not the CPU. It is a `tokio::time::timeout` around the extraction future, and Kreuzberg 4.10.2 already runs pdfium on the blocking pool — both the batch and non-batch arms of `extractors/pdf/mod.rs` call `spawn_blocking` — so a spinning parse holds one blocking thread until it finishes rather than one of the runtime's own workers. Bounding that needs the parser out of process, the same larger change noted above.

## Sandbox postures: the guest gets the tunnel port and nothing else

Ailoy's host egress rule used to match every port, so any host service on a reachable address was in reach from inside the guest. With that narrowed to named ports (ailoy#443), the sandbox has to say which host port it needs, and the only one it needs is the tunnel server's.

That port is ephemeral and the network policy is fixed when the sandbox is created, so the order had to change: start the tunnel server, read its port, restore the sandbox granting that port, then attach the guest pump. `mount_vfs_tunnel_in_guest` splits into `spawn_vfs_tunnel` and `attach_vfs_tunnel_in_guest`, and the session run path calls them in that order. Two loose ends from that split are tidied here: `_vfs_forward` no longer owns the server, so the binding is gone and the lifetime note moves to `vfs_tunnel`, which is what actually holds the mount up (it is declared first and so drops last); and the archive-existence check moves ahead of the spawn, so a session missing its archive no longer starts a server just to drop it.

Postures are stated at every sandbox now rather than inherited, `backend/tests/sandbox_spike.rs` included — it needs no network at all and says so:

| Call site | Posture | Why |
| --- | --- | --- |
| `state/session.rs` run | `Public` + the tunnel port | The guest installs packages and serves its FUSE mount from the host |
| `router/session.rs` create | `HostOnly` | Nothing runs here. `insert` stops and archives it immediately, and each run restores it with the posture that run needs |
| `coworker` runenv | `Public`, no host port | Installs Python packages; reaches no host service |
| `deep_research` runenv | `Public`, no host port | Same |

The last two were relying on the old default, which was public egress. Ailoy's default is the narrow posture now, so leaving them implicit would have quietly cut their network. Worth knowing for anything else in the tree that builds a sandbox without saying `.network(...)`.

The coworker's `input_dir` becomes read-only. Output belongs in `artifacts_dir`, and a guest that can overwrite the user's originals can also hand a host-side parser (docling, the xlsx and pptx skills) a file of its own choosing in place of what the user uploaded. Measured in a booted guest by `4a99c42`, and scoped to the two CLI binaries — `backend/src` contains no `VolumeMount` at all, so the served product never binds a host directory into a guest.

### Ailoy pin

Now `2b795e06`, the head of Ailoy `develop`. That commit is brekkylab/ailoy#443 squashed, merged 2026-08-07, and it sits on top of #442.

`with_host_ports` does not exist before that pin, so this PR cannot build against an older Ailoy.

Earlier versions of this description named a blocker here: the pin was on #443's branch head, which was not on Ailoy `develop` yet. #443 is merged and `33a4660` repins to the merged commit, so that blocker is closed.

How the pin got there is still worth recording, because three of the commits in it fix the posture machinery this PR is built on. It started at `d0cbf7b`, five commits behind #443's head. `73bb854` moved it to that head, picking up `809ff3f` (defaults `NetworkPosture::network` when deserializing — without it a stored posture that omits the field fails to parse outright, and both `{}` and `{"host_ports":[8080]}` are such postures), `4aa02d5` (keeps the parent's network policy across a fork), and `4859a32` (refuses the RFC 8215 local-use NAT64 range in the egress gate). All three are in the squash.

An earlier version of this description claimed the repin also completed the one #219 had pending. It does not: #219 did that itself in `a563660` and points at `0767a08`, and `0767a08..2b795e06` is the whole of #443, not a comment line.

## Review follow-ups

Five commits on top of the original set — one from review here, four from reviewing this branch against a build — plus `33a4660`, the merge with `main` described at the top:

### [`b70d0a7`](https://github.com/brekkylab/agent-k/commit/b70d0a7) — fix(vfs): bound the whole token handshake, not each read

`cfa9428` set the read timeout once before reading the token line. That is `SO_RCVTIMEO`, which applies per `recv` call, and `read_token_line` reads a byte at a time — so every byte restarted the timer. A client dribbling just under the timeout held its thread for the line limit times the timeout, roughly eleven minutes, rather than the five seconds intended; the accept loop spawns a thread per connection with no cap, so that is a thread the guest can hold at will. Measured against the real server at one byte every third of the budget: `held 44.13s against a 5s budget`.

The remaining budget is now derived from a single deadline before each byte. The new test fails on the old behaviour with that same measurement and passes at 7s once the deadline is total — the existing silent-client test could not tell the two apart.

### [`8ba588f`](https://github.com/brekkylab/agent-k/commit/8ba588f) — docs(docling): correct why the process-group kill is safe

Two comments described a mechanism the code does not use. `convert_pdf_to_md_within` said the parser is killed by letting the timeout drop the future; that was true before `5fb431c`, which keeps the future alive and SIGKILLs the group explicitly. And `kill_process_group`'s SAFETY note claimed the pid cannot have been reaped because the caller still owns the child handle — owning the handle does not prevent reaping, because `wait_with_output` is a join over `wait()` and the two pipe drains (tokio 1.48, `process/mod.rs:1463`), so `wait()` reaps as soon as the parser exits while the drains stay pending on a grandchild holding the pipes. That is exactly the OCR-helper case the kill exists for, so the stated invariant was false where it mattered. The call is still safe for a different reason — a pid stays reserved while it is in use as a process group id with live members — and the comment now says that.

### [`73bb854`](https://github.com/brekkylab/agent-k/commit/73bb854) — chore(deps): move the ailoy pin to #443's head

See the Ailoy pin section above.

### [`ddb5526`](https://github.com/brekkylab/agent-k/commit/ddb5526) — fix(knowledge): time-box the Kreuzberg PDF path too

From review: the docling timeout does not cover the engine the server runs. See "The engine the server actually uses" above.

### [`4a99c42`](https://github.com/brekkylab/agent-k/commit/4a99c42) — test(coworker): ask the guest whether attachments are read-only

Closes the one gap this description listed as unverified. See "The read-only attachments mount, measured" above.

### The read-only attachments mount, measured

Listed as unverified in an earlier version of this description, on the assumption that the xlsx and pptx skills lived in the container image and could not be read from here. They do not — they are in this repo under `agent-k/src/agents/skill/`, injected as `FileEntry`s — so the question was answerable after all.

`4a99c42` asks the guest directly, since from the host a silently-writable mount looks the same as a correct one:

```
read      exit=0 out="original"
overwrite exit=2 err="sh: 1: cannot create /root/attached/upload.txt: Read-only file system"
create    exit=1 err="touch: cannot touch '/root/attached/planted.sh': Read-only file system"
artifact  exit=0
```

`artifact` still succeeding is the control — the refusal is this mount's flag, not a read-only rootfs. Confirmed to discriminate by re-running with `readonly: false`, where both writes return 0 and the assertion fires.

Nothing in the intended workflow wanted that write. The system prompt describes attachments as input and requires output under `{{ARTIFACTS}}`; every path in the xlsx skill's documentation is `artifacts/...`; and the pptx edit case in the harness (`cases/coworker.rs`, case 15) already asks for `artifacts/slides_edited.pptx` with the original left untouched. The one call worth knowing about is `autofit_columns(path)`, which re-saves in place — harmless as documented, since its examples all name artifact paths, but it would now fail if handed an attachment.

Scope is narrower than it first looks: `get_coworker_agent_runenv` is called only from `agent-k/src/bin/run.rs` and `src/bin/test_case.rs`, and `backend/src` contains no `VolumeMount` at all. The served product never binds a host directory into a guest, so this change lands on the two CLI paths.

## Verification

Re-run on the merged tree, at the `2b795e06` pin:

```
cargo clippy --workspace --all-targets --features docling-sys/skip-bundle   # clean
cargo test --workspace --features docling-sys/skip-bundle                   # 210 passed, 0 failed, 12 ignored
```

210 rather than the 176 an earlier run of this reported: #212 brings its own Gmail suite, and three of the difference are this branch's — `b70d0a7`'s dribbling client and `ddb5526`'s two for the Kreuzberg budget. Each of the 12 ignored needs something a plain `cargo test` does not have here: a booted VM, or credentials (`AWS_S3_*`, `NOTION_API_KEY`, `GMAIL_*`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY`). The two that cover this PR's reordering are run explicitly below.

Run three times over to confirm the new docling timeout test is not load-sensitive — an earlier 500 ms deadline raced the stub's own startup, so it is 2 s now.

### The in-VM mount, measured

The reordering is verified against a real VM, not only compiled. Two new tests in `backend/tests/tunnel_e2e.rs` need no provider credentials. A local-files `WorkspaceFs` is enough to exercise spawn, grant, attach, and a read back through FUSE. Re-run on the merged tree, since `session.rs` and this file are where the merge had real decisions to make:

```
cargo test --workspace --features docling-sys/skip-bundle --test tunnel_e2e -- --ignored --exact \
  local_workspace_mounts_with_only_the_tunnel_port_granted \
  mount_fails_when_the_tunnel_port_is_not_granted --nocapture --test-threads=1

local_workspace_mounts_with_only_the_tunnel_port_granted ... ok
mount_fails_when_the_tunnel_port_is_not_granted ... ok
    guest: resolved host.microsandbox.internal:59553 -> 172.16.0.69:59553, connecting
           connect attempt 0..9 failed: Connection refused (os error 111)
    host:  mount refused as expected: in-guest tunnel mount failed (exit 1)

test result: ok. 2 passed; 0 failed; 0 ignored; 4 filtered out; finished in 3.86s
```

The first test prints nothing on success; what it asserts is that the guest's `cat /mnt/ws/files/hello.txt` comes back as `tunnel-ok`, so passing is the whole signal.

The pair is what makes it meaningful. With the port granted the guest reads a host file through the tunnel; with the identical setup minus the grant, name resolution still works (`Rule::allow_dns()` doing its job) and only the connect is refused, immediately rather than by timeout. So the grant is pinned to one port rather than to "some port", and `SandboxNetwork::Public` on its own genuinely does not reach the host any more.

That second result is also the migration hazard, demonstrated: on this branch's Ailoy pin, the previous code (which built the sandbox before the server existed and never named a port) produces exactly that failure.

The pump ELF does build here after all. An earlier claim that it could not was wrong: the 0-byte stub in `target/` was a cached failure from 07-28, and `build.rs` already resolves the toolchain's bundled `rust-lld`. A rebuild stages a 447 KB static aarch64 Linux ELF and `tunnel_available()` is true.

Not run on the merged tree, and worth naming rather than leaving to the ignore list:

- `s3_read_throughput`, `notion_read_full_content`, `tunnel_debug`, `gmail_live_mirror` — need `AWS_S3_*` / `NOTION_API_KEY` / `GMAIL_*`, which are not present here. They exercise provider accessors rather than anything this PR changes, and the tunnel path they share is what the two tests above cover.
- `agent_finds_the_mount_unprompted` — main's test, moved here onto the two-step API. It boots a VM *and* spends real LLM tokens, so it is the one merge decision in `tunnel_e2e.rs` that is compile-verified rather than run. The API it now calls is the same pair the two tests above exercise live.
- `attachments_are_read_only_in_the_guest_and_artifacts_are_not` — run in a guest at the earlier pin (output quoted above), not re-run since the merge. `main` since `be01c6b` contains no change to `VolumeMount` at all, so the mount it asserts on is untouched.

## cargo audit

Re-run on the merged tree, unchanged from the pre-merge run. An earlier run of mine reported seven vulnerabilities, but that was against a tree from before #219 landed; its dependency work had already cleared five of them. Now, over 902 crate dependencies:

```
1 vulnerability, 3 unmaintained warnings
```

The one vulnerability is `rsa` 0.9.10, RUSTSEC-2023-0071 (Marvin, key recovery through timing sidechannels). It has no patched release at all, so it cannot be upgraded. It is also not in the build graph: the only package depending on it is `sqlx-mysql`, and `cargo tree -i rsa --workspace --all-features` prints nothing, so it sits in `Cargo.lock` without ever being compiled. Removing it would mean sqlx dropping its optional MySQL driver from the lock graph, which is not something this repo controls.

Cleared by #219, recorded here because I previously listed them as outstanding:

| Crate | Advisory | Now at |
| --- | --- | --- |
| `astral-tokio-tar` | RUSTSEC-2026-0145 | 0.6.4 (needed `>=0.6.2`) |
| `crossbeam-epoch` | RUSTSEC-2026-0204 | 0.9.20 |
| `quick-xml` | RUSTSEC-2026-0195, RUSTSEC-2026-0194 | 0.41.0 |
| `quinn-proto` | RUSTSEC-2026-0185 | 0.11.16 |
| `lopdf` | RUSTSEC-2026-0187 | 0.44.0, via `kreuzberg` 4.10.2 |

`lopdf` is worth calling out: its exact `kreuzberg` pin was the recorded blocker for that advisory, and #219's bump to 4.10.2 is what moved it. That path parses attacker-influenced PDFs on the host in the same way docling does, so it mattered more than its position in the list suggested.

The three unmaintained warnings are `bincode` 2.0.1, `paste` 1.0.15, and `rustls-pemfile` 2.2.0. None has a vulnerability attached.

No advisory against `jsonwebtoken`: #219 moved it to 11, and the note that sent me looking (a jsonwebtoken 9 auth bypass) is not in the current database.

## Follow-ups not in this PR

- The tunnel token is passed to the guest through the `exec_shell` command string, so it lands in the guest's process table and in any host-side logging of that command. The guest is root and is the legitimate holder, so this is about the log channel, not the guest.
- Ailoy's `python_repl` bootstraps by downloading the `uv` binary from a `releases/latest` URL and running it, with no checksum and a mutable tag. It runs inside the guest, so the blast radius is the guest, but it is an unverified remote binary all the same. `SandboxNetwork::with_domain_suffixes` (also in ailoy#443) is the tool for narrowing it; measured live, an allowlist of `github.com` reaches the release asset while `pypi.org` stays blocked.
- Ailoy's `a2a` module fetches an agent card from an arbitrary URL host-side with no timeout, and turns the fetched name and description into a tool definition. `mod a2a;` is commented out, so it is not compiled today. It needs the same egress gate as `web_fetch` before it comes back.


docling-sys/src/lib.rs
```diff
@@ -93,10 +94,27 @@ impl Default for PdfOptions {
     }
 }
 
-/// Convert PDF bytes to Markdown.
-pub async fn convert_pdf_to_md(
+/// How long a single conversion may run before the parser is killed. The input
+/// is a document from outside — uploaded, or fetched from the web — and the
+/// parser runs on the host with the caller's own privileges, so a document that
+/// makes it spin must not be able to hold a worker forever.
+pub const DEFAULT_TIMEOUT: Duration = Duration::from_secs(300);
```
selloriwoo, line 101: The docling half works — `cargo test -p docling-sys --features skip-bundle` passes both tests here, group kill included. But the server never takes that branch, so the table's "Parser has no timeout, leaves orphans → Fixed" isn't true yet: `PdfEngine` defaults to `Kreuzberg` (`agent-k/src/knowledge_base/translator/mod.rs:15-19`), knowledge ingest uses that default (`backend/src/state/knowledge/mod.rs:300`), and it runs pdfium in-process with `ExtractionConfig::default()` — whose timeout field kreuzberg documents as "`None` means no timeout (unbounded extraction time)".

Five lines in `translator/pdf.rs` would make the claim true:

```rust
PdfEngine::Kreuzberg => {
    let bytes = tokio::fs::read(pdf_path).await
        .with_context(|| format!("failed to read {}", pdf_path.display()))?;
    let config = ExtractionConfig {
        extraction_timeout_secs: Some(docling_sys::DEFAULT_TIMEOUT.as_secs()),
        ..Default::default()
    };
    let handle = tokio::runtime::Handle::current();
    tokio::task::spawn_blocking(move || handle.block_on(extract_bytes(&bytes, PDF_MIME_TYPE, &config)))
        .await
        .context("kreuzberg extraction task panicked")?
        .with_context(|| format!("kreuzberg extraction failed for {}", pdf_path.display()))?
        .content
}
```

`spawn_blocking` is the part that matters: nothing outside `batch_mode.rs` reads `is_batch_mode()` in 4.10.2, so extraction runs inline on a runtime worker, and one spinning PDF holds one of core-count workers for the life of the process. On the blocking pool it leaks one thread of 512 and the service keeps answering. `extraction_timeout_secs` then returns an error wherever the future yields (OCR, IO), and the `tokio-runtime` feature it needs is already on via kreuzberg's defaults.

nuri-yoo: Fixed in `ddb5526`.

`extraction_timeout_secs` is wired as you described, and `tokio-runtime` is in kreuzberg's own default features. Checked against the `payslips.pdf` fixture rather than by reading: the default config extracts fine, a zero-second budget returns `Extraction timed out after 1ms (limit: 0ms)`.

The committed fix is the config change alone, without the `spawn_blocking` wrapper. `is_batch_mode()` is read outside `batch_mode.rs` in 4.10.2, and both sides of that branch in `extractors/pdf/mod.rs` (`:345` batch, `:417` non-batch) already call `spawn_blocking` under the `tokio-runtime` gate.

The finding table said `Fixed` on the docling work alone; the description now names both engines and which one the server runs.

selloriwoo (approved):

khj809 (approved):
Currently I'm working on using virtiofs-based sandbox filesystem powered by cortex, which will deprecate the sandbox tunnel logics.
So it's totally okay to merge your work so far, but I'd like to suggest not to put your efforts so much on this.
I'm sorry to say this too late..

PR #222, opened 2026-07-30
feat(vfs): add Google Drive as a workspace source mount

selloriwoo:
Mounts a Google Drive account as a live tree in the workspace VFS. The layout mirrors Drive's own sidebar, and nothing is kept on disk — unlike the Gmail mount, whose `messages.list` returns bare ids and so needs a mirror before anything can search it, Drive's `files.list` answers one folder in one call with every field a listing needs.

```
/My Drive/…              the account's folder tree
/Shared with me/         everything shared to this account
/<shared drive>/         each shared drive the account can see
```

## What each row becomes

| Drive row | Entry | Read |
|---|---|---|
| folder | directory | one `files.list`, cached 5 minutes |
| a file with bytes | its own name, exact size | `alt=media` with an HTTP `Range` |
| Docs / Sheets / Slides | `.gdoc.json` / `.gsheet.json` / `.gslide.json` | that document's own API |
| Forms / Maps / Drawings | not listed | nothing can render them |

Docs-editors files hold no bytes of their own, and every rendering Drive *can* export drops what an edit needs — csv gives one tab of a workbook, xlsx refuses above 10MB, txt loses paragraph indices. Each document's own API carries all of it, so that JSON is what the entry serves.

Tabs are named in quoted A1 notation, because a bare name is a cell reference whenever it can be one: measured, `ranges=A1` answered the *first* sheet's A1 cell rather than a sheet called `A1`, which would have attached one sheet's cells to another. A spreadsheet's cells come from `values:batchGet` rather than `includeGridData=true`. That flag bills per **allocated** cell at 578-920 bytes each, and a tab allocates 1000x26 whether or not one cell is filled: a real workbook of 210,125 allocated cells estimated 189MB, while its values were 194KB across nine tabs. A tab past the budget says so in `valuesOmitted`; a values endpoint that is missing or forbidden leaves `valuesUnavailable` and the workbook's shape still serves.

A document's length is unknown until it is produced, so a listing reports a placeholder and marks it an estimate; a handle resolves it on `open`, and the cache answers exactly once the bytes are in hand. Callers that discard lengths skip resolution entirely — the FUSE tunnel parses `name` and `is_dir` out of a listing, so `find`, `ls` and `Glob` never pay to render a document.

Two Drive files can share a name, so a collision is numbered — before the extension (`report (2).gsheet.json`), because a name numbered after it stays unique but drops out of every glob that would find it.

A PDF, pptx or docx keeps its text compressed and font-encoded, so reading the bytes cannot find a phrase inside one — measured on a 33MB PDF that plainly contains a phrase: zero matches, even after inflating every stream in it. Google extracted that text at upload, and asking the index it built (`q=fullText contains`) does find it. That lands here as the `search` domain command — phrase in, one JSON line per hit out, each line carrying its own explanation when the hit is a type this mount cannot list — which **nothing can invoke yet**: no transport routes a `.cmd/<name>` write to `Resource::command`. The command itself is exercised directly by a live test; the path the prompt advertises is not.

## Measured against a real account

838 files, walking 60 folders.

| | |
|---|---|
| walking 60 folders (`find` / `Glob`) | 25.7s, zero renders |
| one folder's listing (29 entries) | 1.2s |
| `cat` a spreadsheet | 14MB in Drive, 194KB served, 1.58s (0.00s again) |
| `cat` a document | 26KB in Drive, 3.4MB served, 1.8s |
| `cat` a presentation | 8.8MB in Drive, 2.3MB served, 1.3s |
| `grep` a Korean phrase across the tree | 3 hits, 165s, 72.5MB moved (42 text files searched, 795 binaries skipped after their first buffer) |
| content search for a phrase inside a PDF | 1 hit in 1.2s, no file transferred |

All three grep hits were cell values inside spreadsheets — the case `spreadsheets.get` alone cannot answer.

Drive's reported size predicts nothing about the JSON: it is storage accounting — embedded images, formatting, revisions — while the JSON is text, structure and style runs. Measured across the corpus the ratio runs from 0.01x to 130x, in both directions. A 377MB spreadsheet served 331KB — of its 104,709 allocated cells, 1,614 hold a value. A 26KB document served 3.4MB. A 60MB document served 2.5MB: 23,678 characters across 1,513 style runs, with 134 embedded images left behind as links.

## Also in here

- `content_type` on `DirEntry`/`FileStat`/`Stat`, carrying the provider's own idea of what a node is. WebDAV serves it as a dead property (`subject-content-type` in `urn:agent-k:workspace`), because dav-server derives `getcontenttype` from the request path.
- `Resource::resolve_size_on_stat` — Notion sizes a page during `stat`; Drive declines, or a listing of documents would export every one of them.
- `Resource::reports_content_type`, so the WebDAV layer can skip the property lookup for mounts that have nothing to say.
- `FileStat::serves_whole` — whether asking for part of a node is the same work as asking for all of it. A document is built on request, so `head -c 100` builds all of it, and the read strategy follows that rather than the size: fetched whole and kept up to the whole cache budget, while a real file passes its range down and is kept only while small. Conflating this with the size produced two bugs — a 256KB read of an object Drive listed without a size fetched the whole thing (2GB per chunk, uncacheable), and a document past the cache's 8MB limit re-rendered once per chunk: 42 renders and 420MB to deliver 10MB. Notion's `page.json` was whole-only all along and never said so.
- A ceiling where the memory is spent: a document over 64MB is refused rather than produced, which is under the cache budget, so anything producible is cacheable and no size degrades into re-rendering. And a file Drive lists without a size is measured rather than guessed — one ranged byte carries the total in `Content-Range` (119MB PDF: 1 byte, 0.68s).

## Test plan

- `cargo test -p workspace -p agent-k-backend` — 89 + 58, covering the naming rules (non-ASCII, collisions, path escapes), the placeholder's bounds, and the listing-cost split.
- `GDRIVE_BASE_URL=… cargo test -p workspace gdrive_mock -- --ignored` — full tree walk against enterprise-mock (84s, exercises a 10,000-entry paginated listing).
- `GDRIVE_CLIENT_ID/SECRET/REFRESH_TOKEN=… cargo test -p workspace gdrive_live -- --ignored` — tree, ranged reads of originals, and every native JSON against a real account.

## Known gaps

- Read-only. The mount holds `drive.readonly`; editing needs re-consent for `documents`/`spreadsheets`/`presentations` and a `.cmd` surface.
- Content search is implemented but unreachable. `Resource::command` has no transport: a write to `.cmd/<name>` goes to `write_bytes` and comes back `Unsupported`, so Drive's `search`, gmail's `send`/`reply` and notion's `page-create` are all dormant — the mount prompt documents `echo '<phrase>' > .cmd/search` ahead of the wiring, as notion's already does. Until that one hop lands, a PDF's contents remain unsearchable through the mount.
- `Resource::prompt()` is not wired into the agent's system prompt, so the mount's own instructions — verified line by line against the implementation in this branch — are not read by anything yet.
- An unread document reports 8MB, so `find -size` and `ls -l` are wrong for it until something reads it. Making them exact costs one render per document: 2.4s for a six-document folder concurrently, 6.3s serialised by per-entry `getattr`.


workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1663 @@
...
+    }
+}
+
+/// Whether this entry's size is a placeholder rather than a length: true for a
+/// document, whose JSON nobody has measured until it is produced.
+///
+/// Deliberately false for a file Drive reported no size for, even though that also
+/// gets the placeholder: resolving it would mean downloading the object — possibly
+/// gigabytes — to answer a `stat`.
+fn is_estimate(c: &Child) -> bool {
```
nuri-yoo, line 715: This is where the new contract comes apart. `entry_size` hands an unsized `Serves::Original` row the 8 MiB placeholder while `is_estimate` calls that length exact, so the listing asserts a number it does not have, and `CachedResource`'s stat fast-path (`cache.rs:610`, `Some(e) if e.size > 0`) returns it verbatim without ever reaching `stat` here. `probe_len` is unreachable for any caller that lists before it reads, which is all of them.

A 20 MB PDF that Drive listed without `size`, read as `ls` then four 256 KiB windows. Measured against a mock of the Drive endpoints, driving the real `GdriveResource` behind the real `CachedResource` that `build_mounts` installs, with a log of every request and its `Range`:

```
[warm] listing: size=8388608 size_is_estimate=false
[warm] stat: size=8388608 estimate=false serves_whole=false | probe requests=[]
[warm] 4x256KiB windows -> 4 media requests, ranges=[None, None, None, None], 80 MB moved

[cold] stat: size=20971520 estimate=false | probe requests=[Some("bytes=0-0")]
[cold] 4x256KiB windows -> 4 media requests, ranges=[Some("bytes=0-262143"), Some("bytes=262144-524287"), Some("bytes=524288-786431"), Some("bytes=786432-1048575")], 1 MB moved
```

80 MB to deliver 1 MB. `whole_or_small` reads 8388608 as a known length at or under `MAX_CONTENT_BYTES`, so each window fetches the object entire, and `content_put` then discards it, because 20 MB over the ranged limit with `whole = false` is not kept. This is the failure the description lists as fixed ("a 256KB read of an object Drive listed without a size fetched the whole thing (2GB per chunk, uncacheable)"); it is fixed on the cold-stat path only.

The same row through `WorkspaceFs`:

```
[fs] read_dir reports len=8388608 for a 20971520 byte file
[fs] open handle metadata().len = 8388608
```

`File::metadata` skips resolution because the flag says exact, so a WebDAV `GET` advertises `Content-Length: 8388608` for a 20 MB body, and `seek(SeekFrom::End)` lands at the placeholder.

The guard in `cache.rs` was written for this exact row and expects the opposite flag:

```rust
/// A file whose length the backend never reported must still be read a range at
/// a time. Its placeholder size looks small, and treating "small" as "cacheable"
/// downloaded the whole object once per chunk
let unsized_file = FileStat { size: 8 << 20, size_is_estimate: true, serves_whole: false, .. };
```

Returning `true` here for `(Serves::Original, None)` closes all three symptoms at once, and it costs a listing nothing: `resolve_size_on_stat()` stays `false`, so `readdir` still resolves nothing, and what a later `stat` or `open` performs is `probe_len` rather than a fetch. The doc comment's reason for `false` ("resolving it would mean downloading the object, possibly gigabytes") describes the world before `probe_len` existed; by your own measurement it is one byte and 0.68s.

workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1663 @@
...
+            kind: if child.kind.is_dir() {
+                FileKind::Dir
+            } else {
+                FileKind::File
+            },
+            size: size.unwrap_or_else(|| entry_size(&child)),
+            mtime: child.mtime,
+            created: child.created,
+            content_type: child.mime_type.clone(),
+            size_is_estimate: size.is_none() && is_estimate(&child),
```
nuri-yoo, line 563: The same gap on the other branch. When `probe_len` fails, `size` is `None`, `is_estimate` is `false` for this row, and the 8 MiB placeholder goes out as a measured length. `probed_len` documents the intent as "the placeholder stands rather than a wrong number", which is what happens, except nothing marks it as a placeholder, so `whole_or_small` and `File::metadata` both read it as fact. Fixing `is_estimate` covers this line too.

backend/src/router/webdav.rs
```diff
@@ -348,6 +397,22 @@ impl DavDirEntry for DavDirEntryAdapter {
...
+const SUBJECT_TYPE_PROP: &str = "subject-content-type";
+const PROP_NAMESPACE: &str = "urn:agent-k:workspace";
+const PROP_PREFIX: &str = "W";
+
+fn subject_type_prop(value: Option<String>) -> DavProp {
+    DavProp {
+        name: SUBJECT_TYPE_PROP.to_string(),
+        prefix: Some(PROP_PREFIX.to_string()),
+        namespace: Some(PROP_NAMESPACE.to_string()),
+        xml: value.map(String::into_bytes),
```
nuri-yoo: `DavProp.xml` holds a complete element, not the value. dav-server parses it back with `Element::parse2` and ships `DavProp::new(name, prefix, namespace, value)` for this purpose, so a bare mime string fails to parse: the property reaches the client empty and every node logs an error.

Real PROPFINDs through `DavHandler` on dav-server 0.11, against a `DavFileSystem` with these three methods copied over:

```
[ERROR dav_server::handle_props] davprop_to_element(): XmlParseError. Please check your
GuardedFileSystem.get_props() implementation. 'xml' should include complete xml tag.
Use DavProp::new() to easy create a DavProp with valid xml syntax.

# as written, allprop
... <W:subject-content-type xmlns:W="urn:agent-k:workspace"></W:subject-content-type> ...

# as written, named prop
... <W:subject-content-type xmlns:W="urn:agent-k:workspace"></W:subject-content-type> ...

# DavProp::new(), allprop
... <W:subject-content-type xmlns:W="urn:agent-k:workspace">application/vnd.google-apps.spreadsheet</W:subject-content-type> ...
```

Both request shapes end up empty, and not by the route I first assumed: the named-prop path never reaches `get_prop`, because the deadprop branch at `handle_props.rs:1157` is gated on `self.name == "prop"`. Both go through `davprop_to_element`.

Two things worth knowing while fixing it. `handle_props.rs:1226` appends `get_props` on top of whatever the per-property loop produced, so once the value parses, a named PROPFIND emits the element twice; having `get_prop` answer `Err(FsError::NotFound)` leaves exactly one:

```
# DavProp::new() + get_prop declines, named prop
<D:prop><W:subject-content-type xmlns:W="urn:agent-k:workspace">application/vnd.google-apps.spreadsheet</W:subject-content-type></D:prop>
```

And `DavProp::new` interpolates the value through `format!` without escaping, while a Drive `mimeType` is whatever the uploader set. One ampersand loses it:

```
# mime = "application/x-tar&gzip"
[ERROR ...] XmlParseError
... <W:subject-content-type xmlns:W="urn:agent-k:workspace"></W:subject-content-type> ...
```

Nothing is injectable, since xmltree rejects a second root before anything is emitted, so the cost is a silently dropped value rather than a hole. It still wants escaping to be correct.

A test that reads the property back out of a 207 body would have caught the original shape. `reports_content_type_is_answered_by_the_mount_not_a_stat` covers the capability gate and stops one layer short of the serialization.

workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1663 @@
...
+            .unwrap_or_default()
+            .into_iter();
+        let mut budget = GRID_BYTES_BUDGET;
+        for tab in v
+            .get_mut("sheets")
+            .and_then(|s| s.as_array_mut())
+            .into_iter()
+            .flatten()
+        {
+            let Some(range) = ranges.next() else { break };
```
nuri-yoo: This fold pairs `sheets` with `valueRanges` by position while `tab_titles` builds the request from a filtered and truncated view of the same array, so the two can disagree in two ways.

Past `MAX_TABS` the tail gets no values and no note, which makes it the one omission in this file that does not explain itself:

```
[70 tabs] served=70 tabs, with values=64, with valuesOmitted=0, silently empty tabs=[64, 65, 66, 67, 68, 69]
```

And a sheet with no `properties.title` is dropped by `tab_titles`' `filter_map` but still consumed here, which shifts every tab after it:

```
[titleless] sheets[0] title=Some("Alpha") values=[["ALPHA-CELL"]] omitted=None
[titleless] sheets[1] title=None          values=[["GAMMA-CELL"]] omitted=None
[titleless] sheets[2] title=Some("Gamma") values=None             omitted=None
```

`sheets[1]` is serving `Gamma`'s cells and `Gamma` is serving none, which is the outcome `quote_a1` exists to prevent, reached from the other side. `valueRanges[].range` carries the sheet title, so pairing on that instead of on position removes both cases, and the tabs past the cap can carry the `valuesOmitted` note the budget path already has.

workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1663 @@
...
+                continue;
+            };
+            let values = range.get("values").cloned().unwrap_or(Value::Array(vec![]));
+            let cost = served_len(&values);
+            if cost > budget {
+                obj.insert(
+                    "valuesOmitted".into(),
+                    serde_json::json!({ "bytes": cost, "remainingBudget": budget }),
+                );
+                budget = 0;
```
nuri-yoo: Zeroing the budget spends the remainder on one tab's behalf. A workbook whose second tab is large loses every later tab regardless of size:

```
[budget] sheets[0] small1 has_values=true  omitted=None
[budget] sheets[1] huge   has_values=false omitted={"bytes":14588892,"remainingBudget":8388588}
[budget] sheets[2] small2 has_values=false omitted={"bytes":20,"remainingBudget":0}
```

`small2` needed 20 bytes with 8,388,588 still unspent. A plain `continue` that leaves `budget` alone matches what the constant documents, "spent tab by tab in the workbook's own order until it runs out", and it keeps the note meaningful: `remainingBudget: 0` currently says more about the previous tab than about this one.

workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1663 @@
...
+            mtime: None,
+            created: None,
+            serves: Serves::Nothing,
+            size: None,
+        };
+        let mut children = vec![
+            section(MY_DRIVE_NAME, MY_DRIVE_ID, GKind::Folder, None),
+            section(SHARED_WITH_ME_NAME, SHARED_WITH_ME_ID, GKind::Folder, None),
+        ];
+        if let Ok(drives) = self.accessor.list_shared_drives().await {
```
nuri-yoo: A failure here is indistinguishable from an account that has no shared drives, and the reduced root is then cached for `DIR_TTL`. With `/drives` answering 500:

```
[drives 500] root sections = ["My Drive", "Shared with me"], /drives attempts = 6
[drives 500] second readdir -> same, new /drives attempts = 0
finished in 33.19s
```

Three costs stack up. The first `ls` of the mount blocks 33 seconds walking the retry ladder, the shared drives then disappear without a log line, and the second listing inside the TTL makes no attempt at all, so nothing recovers for five minutes. A `tracing::warn!` on the error arm names the first two, and skipping the `dir_cache` insert when the root was built from a failed drive listing fixes the third.

workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1663 @@
...
+pub struct GdriveResource {
+    accessor: GdriveAccessor,
+    /// Lengths learned by probing (file id → bytes), for the rows Drive listed
+    /// without a `size`. One probe per file per session: `stat` is asked once per
+    /// entry after every listing, and a probe is a request.
+    probed: Mutex<HashMap<String, u64>>,
+    /// Per-directory listing cache (folder path → children). Path resolution walks
+    /// parent listings, so one cached listing answers readdir, stat and the lookup
+    /// a read starts with; fetching the bytes themselves still costs a request.
+    dir_cache: Mutex<HashMap<String, CachedListing>>,
```
nuri-yoo, line 244: Neither map is ever pruned. `grep -nE 'dir_cache|probed' | grep -E 'remove|retain|clear|shrink'` finds nothing, so a listing keeps its `Arc<Vec<Child>>` for the life of the mount long after `DIR_TTL` has made it unusable, and the corpus has a folder that lists 10,000 entries.

Walking 40 folders against a mock, reading the two map lengths straight off the resource:

```
[caches] before any walk: (0, 0)
[caches] after walking 40 folders: dir_cache=42 entries, probed=40 entries
[caches] after a second walk: (42, 40)
```

Both are high-water marks: one entry per folder ever listed, one per file ever probed. `CachedResource` above bounds both of its equivalents (`CONTENT_CACHE_BUDGET`, `MAX_LEARNED_SIZES`); dropping entries older than `DIR_TTL` on insert gives these the same property in a few lines.

workspace/src/vfs/accessor/gdrive.rs
```diff
@@ -0,0 +1,781 @@
...
+        let url = format!(
+            "{}/files/{id}?alt=media&supportsAllDrives=true",
+            self.urls.drive
+        );
+        let resp = self
+            .send_with_refresh(|t| {
+                let req = self.client.get(&url).bearer_auth(t);
+                match &range {
+                    // HTTP byte ranges are inclusive at both ends.
+                    Some(r) if r.end > r.start => {
```
nuri-yoo, line 595: An empty range falls into the `_` arm, so no `Range` header is sent and Drive answers with the object:

```
[empty range] returned 0 bytes; request Range=[None]; body moved 20 MB
```

`File::read_bytes(0)` reaches this. `r.end == r.start` can answer `Ok(Vec::new())` before the request is built.

workspace/src/vfs/accessor/gdrive.rs
```diff
@@ -0,0 +1,781 @@
...
+        loop {
+            let resp = build(&token).send().await?;
+            let status = resp.status();
+            if status == reqwest::StatusCode::UNAUTHORIZED && !refreshed {
+                *self.access_token.lock().await = None;
+                token = self.token().await?;
+                refreshed = true;
+                continue;
+            }
+            let retryable =
```
nuri-yoo, line 386: Drive reports per-user rate limits as 403 with `rateLimitExceeded` or `userRateLimitExceeded`, which this classifies as terminal:

```
[403 rateLimitExceeded] files.list attempts = 1, elapsed = 1.8ms,  err = true
[429]                   files.list attempts = 6, elapsed = 33.9s,  err = true
```

Reading `reason` out of the 403 body would put it on the same ladder. The second number is worth a look on its own: the doc above says the worst-case total stays low because these calls sit behind a filesystem op the agent blocks on, and the measured worst case is 34 seconds. It is also what turns the shared-drives failure into a 33-second stall on the first `ls`.

workspace/src/vfs/accessor/gdrive.rs
```diff
@@ -0,0 +1,781 @@
...
+    /// its size), and the indented copy written back out — and where the only honest
+    /// limit can live: a reader cannot ask for part of a document, so serving one is
+    /// all-or-nothing, and past some size the answer has to be "no" rather than a
+    /// gigabyte of allocations per read.
+    async fn get_pretty(&self, url: &str) -> anyhow::Result<Vec<u8>> {
+        let resp = self
+            .send_with_refresh(|t| self.client.get(url).bearer_auth(t))
+            .await?
+            .error_for_status()?;
+        if let Some(len) = resp.content_length()
```
nuri-yoo: The pre-check fires only when the response declares a length; without one the body is buffered whole and the limit is then applied to what has already been allocated. Measured by counting what the server actually wrote:

```
[no content-length]   refused = true, server streamed 70 MB before the guard fired
[with content-length] refused = true, server body handed over 0 MB
```

`reqwest` is built here without `gzip`, so Google usually sends a length and the second line is the path this normally takes. Summing chunks off `bytes_stream()` and bailing past the limit would make the guard hold when the length is missing. Same shape at line 594.

workspace/src/vfs/cache.rs
```diff
@@ -510,6 +599,12 @@ impl Resource for CachedResource {
...
             }
+            // A file whose cached length is only an upper bound: fall through and ask
+            // the provider, which may be able to answer exactly and cheaply (Drive
+            // reads a file's true length out of one ranged response header). Serving
+            // the cached guess here would repeat the listing's placeholder as if it
+            // were the length — and hand a chunked read the wrong fetch strategy.
+            Some(e) if e.size_is_estimate => {}
             // A file with a known (>0) size: serve it (with the cached times so
             // `ls -l` shows real times, not the epoch — R2).
             Some(e) if e.size > 0 => {
```
nuri-yoo, line 617: This arm rebuilds the `FileStat` from `Default`, and `Entry` carries no `serves_whole` to restore, so the flag survives a cold `stat` and not a listed one:

```
[serves_whole] cold stat -> serves_whole=true
[serves_whole] stat after readdir -> serves_whole=false (provider says true)
```

Nothing trips it today: Notion lists `page.json` as 0 and Drive lists documents as estimates, and both of those fall through instead of taking this arm. That leaves "a `serves_whole` node is always either size 0 or an estimate" as an invariant the code depends on and no provider states, while the read strategy inverts on the flag by an order of magnitude. Carrying it on `Entry` next to `content_type` and `size_is_estimate` costs one field.

workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1663 @@
...
+    ///   #        then python -m uvicorn app.main:app --port 8000
+    ///   GDRIVE_BASE_URL=http://localhost:8000 \
+    ///     [GDRIVE_MOCK_TOKEN=…] cargo test -p workspace gdrive_mock -- --ignored --nocapture
+    ///
+    /// The walk is bounded so the same test runs against a five-file sample
+    /// corpus and a 25k-document hosted one; a real corpus also spans several
+    /// listing pages, which exercises the accessor's pagination for free.
+    #[tokio::test]
+    #[ignore = "requires a running enterprise-mock (GDRIVE_BASE_URL)"]
+    async fn gdrive_mock_tree_and_reads() {
```
nuri-yoo, line 1559: Every test in this file drives `GdriveResource` bare (`grep -c CachedResource` here is 0) while `build_mounts` always wraps it. The wrapper is where the listing feeds the stat and the stat picks the read strategy, so the mock and live tests both pass with the placeholder problem above present.

One mock-backed case through `CachedResource::new(Arc::new(GdriveResource::new(&cfg)?))` that does readdir, then stat, then a windowed read, asserting the request count and the `Range` headers, would cover the path real callers take. A hand-rolled `tokio::net::TcpListener` mock is enough, and no new dependency is needed; happy to hand over the one behind the numbers in this review.

Separately, a few lines down at 1187: `matches!(..., Err(_))` trips `clippy::redundant_pattern_matching`.

workspace/src/vfs/accessor/gdrive.rs
```diff
@@ -0,0 +1,781 @@
...
+/// duplicate/looping `nextPageToken` (a known Drive API pathology with some
+/// query/corpora combos) can't spin forever.
+const MAX_PAGES: usize = 50;
+
+/// Retry budget for a rate-limited/5xx request; same reasoning as the gmail
+/// accessor's (these calls sit behind a FUSE/WebDAV op the agent blocks on, so
+/// the worst-case total stays low).
+const MAX_RETRIES: u32 = 5;
+const MAX_BACKOFF: Duration = Duration::from_secs(16);
+const JITTER_MAX_MS: u64 = 1000;
```
nuri-yoo, line 76: Two mechanical things in this file. `rustfmt --edition 2024 --check` wants four changes: this stray blank line and three wraps inside `probe_len`. The two `assert!`s at 730 and 735 trip `clippy::assertions_on_constants`, and `const { assert!(..) }` keeps the intent while moving it to compile time.

workspace/src/vfs/resource/base.rs
```diff
@@ -137,4 +169,31 @@ pub trait Resource: Send + Sync {
...
+    /// 0 (see [`FileStat::size`]).
+    ///
+    /// `true` (the default) keeps `stat`/`HEAD`/`PROPFIND` exact for providers
+    /// that render on read: Notion's `page.json` is one file per page directory,
+    /// so one render per `ls -l` is a fair trade for a correct length.
+    ///
+    /// `false` is for providers where that trade inverts. Google Drive serves one
+    /// converted document *per file*, and each conversion takes seconds — so
+    /// `ls -l` in a folder of documents would convert every one of them just to
+    /// print their sizes. Such a provider reports 0 and lets the read fill it in.
```
nuri-yoo: This says a provider that opts out reports 0, while the provider it was written for reports 8 MiB and `UNKNOWN_LENGTH_SIZE` spends a paragraph on why 0 is the one value it cannot use. Two halves of the same contract describing different sentinels is part of what makes the unsized-file row easy to get wrong.

workspace/src/vfs/resource/gdrive.rs
selloriwoo: Correct on all three counts, and the reachability argument is the part I had missed: the fast path answers from the listing, so the probe never ran for anyone who lists first. `is_estimate` now covers `size.is_none()`, so the row is a placeholder until the probe answers and exact after. `a35aaf5b`.

workspace/src/vfs/resource/gdrive.rs
selloriwoo: Same fix covers this line: when the probe fails the length stays marked a placeholder, so nothing downstream reads it as measured. `a35aaf5b`.

backend/src/router/webdav.rs
selloriwoo: Confirmed in the crate source: `davprop_to_element` parses `xml`, so this property has never carried a value. It is now built as a complete element with the value escaped, not via `DavProp::new`, which interpolates raw. `get_prop` declines so the named path emits one element rather than two. `0763d788`.

workspace/src/vfs/resource/gdrive.rs
selloriwoo: Fixed by pairing on the title in `valueRanges[].range` rather than on position. A tab that ends up without values now says which of four reasons applies, including the two that were silent: past the cap, and no title to address it by. `42f863b0`.

workspace/src/vfs/resource/gdrive.rs
selloriwoo: Fixed: `continue` without zeroing, so a small tab after a large one still fits. `42f863b0`.

workspace/src/vfs/resource/gdrive.rs
selloriwoo: Two of the three costs fixed: off the retry ladder (one attempt) and a `warn!` on the error arm. The third is not cured by skipping the `dir_cache` insert, because the reduced root is held by `CachedResource`'s `IndexCache` for `LISTING_TTL`, and that layer has no per-listing way to mark a listing incomplete; a fresh mount retries, but within one mount it still waits out the TTL. Failing the root readdir instead I rejected: every path resolves through the root, so a transient `/drives` 500 would disable the mount. `c6d0c8cb`.

workspace/src/vfs/resource/gdrive.rs
selloriwoo: Fixed: expired listings are dropped when the next one arrives, and the probed lengths are capped. `8d4220e0`.

workspace/src/vfs/accessor/gdrive.rs
selloriwoo: Fixed, answered before the request is built. `4394743e`.

workspace/src/vfs/accessor/gdrive.rs
selloriwoo: Fixed: the three rate-limit reasons join the ladder, `insufficientFilePermissions` and `dailyLimitExceeded` stay terminal. A 403 that is not retried now carries its reason, since classifying it consumes the body. `fda945e6`.

workspace/src/vfs/accessor/gdrive.rs
selloriwoo: You were right and my first read was wrong: I assumed Google sends a length. Measured on Docs, Slides, `spreadsheets.get` and `values:batchGet`, all four answer as an HTTP/2 stream with no `Content-Length` at all, so the pre-check never fired. Now read frame by frame, stopping at the ceiling. `df823a8a`.

workspace/src/vfs/cache.rs
selloriwoo: Fixed by carrying the field, and Notion's `page.json` now says in its listing what its `stat` already said. That removes the invariant you named rather than relying on it. `6d75af77`.

workspace/src/vfs/resource/gdrive.rs
selloriwoo: This is the one that mattered most: bare-provider tests are why the placeholder bug passed. Added `vfs::resource::tests::gdrive_mounted`, your mock's shape on a loopback listener with no new dependency, with seven cases asserting request counts and `Range` headers through `CachedResource(GdriveResource)`. `a35aaf5b`, plus the `matches!` lint in `4b3a55d8`.

workspace/src/vfs/accessor/gdrive.rs
selloriwoo: Fixed, with the constant pair moved to `const { assert!(..) }`. Only the files this branch touches are formatted: the repo's rustfmt config asks for nightly-only import options, so a whole-tree run rewrites files this PR has no business in. `4b3a55d8`.

workspace/src/vfs/resource/base.rs
selloriwoo: Fixed: the flag now describes what it governs, whether a `stat` may produce the file, and leaves the placeholder to the provider with the reason Drive cannot use 0. `e41be624`.

nuri-yoo (approved):
LGTM. Checked all fourteen against the code rather than the replies: the placeholder flag now reaches `whole_or_small` and `File::metadata`, the tab pairing round-trips with `quote_a1` (including the doubled quote and an embedded `!`), and `body_within` holds the limit on the endpoints that turn out to declare no length. `cargo test -p workspace -p agent-k-backend` passes, and fmt and clippy are clean on the touched crates.

One thing stays open, and your reading of it is right: within a single mount the reduced root still waits out the listing TTL, because `CachedResource::readdir` caches unconditionally and `listings_complete()` is per-provider rather than per-listing. Out of scope here. Worth an issue so the five-minute window is written down somewhere other than a test comment.

workspace/src/vfs/accessor/gdrive.rs
```diff
@@ -0,0 +1,909 @@
...
+fn endpoints(base_url: Option<&str>) -> Endpoints {
+    match base_url {
+        Some(b) => {
+            let b = b.trim_end_matches('/');
+            Endpoints {
+                drive: format!("{b}/drive/v3"),
+                token: format!("{b}/oauth2/token"),
+                docs: format!("{b}/docs/v1"),
+                sheets: format!("{b}/sheets/v4"),
+                slides: format!("{b}/slides/v1"),
```
khj809: Hardcoding intermediate routes like `/oauth2`, `/docs`, `/sheets`, `/slides` reflects enterprise-mock's specific behavior, so these subpaths won't work for other mock servers.

Instead of designing around enterprise-mock's behavior specifically, I'd suggest supporting per-service `base_url` overrides, and appending the same subpaths used by the official APIs (`/token` for oauth, `/v1` for docs, `/v4` for sheets, `/v1` for slides, etc.).

For example, for enterprise-mock we could set the `base_url` for the token API to `https://enterprise-mock.brekkylab.com/oauth2` (falling back to `https://oauth2.googleapis.com` by default), and appending `/token` would work correctly in both the overridden and default cases.

I'm also not sure why Google hasn't consolidated their API domains, but given the current structure of their API urls, letting users override the base URL per service seems like the best approach.

P.S. This also applies to gmail case.

selloriwoo: Good catch, and it applied more than you knew: this branch changed `{base}/v1` to `{base}/docs/v1` precisely because the mock wanted it, hard-coding the mock further in rather than less.

Done in both branches. `Origins` overrides an origin per service and this code appends only the official suffix, so Sheets can sit on one host and Docs on another; `Origins::behind(host)` keeps the one-host case a single knob, still set by `GOOGLE_API_BASE_URL`. gmail is in #212 (`805cd0b6`) since this branch stacks on it, the Drive services here (`8f1a8cca`), and a test drives two gateways at once to prove a moved origin actually moved (`a2f6cbe9`).

Verified against enterprise-mock, real Gmail and real Drive.

workspace/src/vfs/accessor/gmail.rs
```diff
@@ -23,49 +23,73 @@ const GMAIL_ORIGIN: &str = "https://gmail.googleapis.com/gmail";
...
+    /// Serves the Docs API (`{docs}/v1/documents/…`).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub docs: Option<String>,
+    /// Serves the Sheets API (`{sheets}/v4/spreadsheets/…`).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sheets: Option<String>,
+    /// Serves the Slides API (`{slides}/v1/presentations/…`).
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub slides: Option<String>,
 }
```
khj809: This `Origins` covers more than gmail but it's located in `gmail.rs`, which seems quite awkward.
How about locating it in common module such as `google.rs`, and use it on both gmail and gdrive?

selloriwoo: Thanks, that is a fair thing to catch: the file said gmail while the type described five services, and the Drive accessor was reaching across for it.

Moved to `accessor::google` along with the OAuth origin both providers share, and its tests moved with it (`b3a1a048`).

workspace/src/vfs/resource/gdrive.rs
```diff
@@ -0,0 +1,1905 @@
...
+            eprintln!("pagination exercised: one listing spanned {biggest} entries");
+        }
+    }
+
+    fn live_config() -> Option<GdriveConfig> {
+        Some(GdriveConfig {
+            client_id: std::env::var("GDRIVE_CLIENT_ID").ok()?,
+            client_secret: std::env::var("GDRIVE_CLIENT_SECRET").ok()?,
+            refresh_token: std::env::var("GDRIVE_REFRESH_TOKEN").ok()?,
+            account_email: std::env::var("GDRIVE_EMAIL").unwrap_or_else(|_| "live-test".into()),
```
khj809: Same as in https://github.com/brekkylab/agent-k/pull/212#discussion_r3709466764.
Also gdrive API does not require gmail address, and it's actually not used anywhere.

```suggestion
            client_id: std::env::var("GOOGLE_CLIENT_ID").ok()?,
            client_secret: std::env::var("GOOGLE_CLIENT_SECRET").ok()?,
            refresh_token: std::env::var("GOOGLE_REFRESH_TOKEN").ok()?,
```

workspace/src/vfs/accessor/gdrive.rs
```diff
@@ -0,0 +1,907 @@
...
+        .await
+        .ok()?
+        .json()
+        .await
+        .ok()?;
+    v.get("user")
+        .and_then(|u| u.get("emailAddress"))
+        .and_then(|e| e.as_str())
+        .map(|s| s.trim().to_lowercase())
+}
```
khj809: While reviewing this, I found that the current enterprise-mock does not implement `/about` in gdrive routes. So now I understand why you tried to read `GDRIVE_EMAIL` in `live_config()`. Let me implement it on mock, and `GdriveConfig` won't need to hold `account_email` anymore.

selloriwoo: Thanks — that unblocks creating a Drive mount against the mock at all, since exchange_gdrive_code fails the create when about.get does.

The field I'd keep, though: one person can own several Google accounts and the consent screen lets them pick a different one per mount, so Gmail on the work account and Drive on the personal one both succeed. GET /mounts showing the two emails is how that gets noticed.

The GDRIVE_EMAIL read is gone either way (43295924) — live_config never reaches the code that reads the field.

khj809: <del>Ah I didn't consider such multi-accounts case.. got it.</del>
@selloriwoo I think I need to retract my earlier answer.

A refresh token is only issued after an account is picked on the consent screen, so it's already bound to a single account — which makes `GET /about` the single source of truth for the email. And keeping `account_email` configurable lets it disagree with the refresh token stored beside it, with nothing validating that the two agree. If it isn't configurable, that mismatch never happens.

Moreover, `GET /mounts` still reports the email. It's resolved per mount from that mount's own refresh token via `GET /about`, so the case you describe still surfaces: two mounts consented to two accounts return two different emails.

selloriwoo: You are right, and the argument that convinces me is the second one: asking each token answers the multi-account case too, since two mounts consented to two accounts return two different emails. The stored copy was a second answer with nothing keeping it in agreement.

Removed, along with the about.get at mount-create and its ability to fail a create; the lookup is now GdriveAccessor::account_email, verified against a real account (53b1edd9). gmail keeps its field, where the value names the mirror directory on disk.

That leaves ProviderInfo::Gdrive with nothing to report. Resolving it per mount in GET /mounts would make a listing depend on the network and on every stored token still being valid, so I left that to whoever adds it rather than folding it in here.

khj809 (approved):
I found there's many test codes or comments mentioning about enterprise-mock's specific behavior, but I think our code should not care about it and just focus on how `GDRIVE_BASE_URL` is set. But it would delay this PR too much, so I'll handle it on the following PR.

Thanks for the huge work!

PR #223, opened 2026-08-05
feat(vfs): Slack as a workspace source mount

selloriwoo:
Mounts a Slack workspace as a live tree in the workspace VFS. Nothing is kept on disk: unlike Gmail, whose `messages.list` returns bare ids and needs a mirror before anything can search it, a Slack listing hands back whole objects. Slack has no hierarchy at all, though — only conversations and a stream — so the tree is synthesized along the time axis, and what shapes it is what each level costs in requests.

```
/slack/
├── channels/<name>__C0123/<yyyy-mm-dd>/
│   ├── chat.jsonl                    that day's root messages
│   ├── files/<stem>__F045.pdf        attachments posted that day
│   └── threads/<root-ts>/
│       ├── chat.jsonl                the root plus every reply
│       └── files/<stem>__F046.pdf    attachments posted inside the thread
├── dms/<user>__D0456/<yyyy-mm-dd>/   identical shape
└── users/<name>__U0789.json          who each member is, found by name
```

Measurements below name where they were taken. A **live workspace** is real Slack and small; a **synthetic corpus** is a large offline fixture that exists so listings page and a walk hits its ceiling. Neither substitutes for the other — the live workspace is too small to page, and the corpus is not Slack.

Rate limits are the binding constraint, and they are per method and instantaneous rather than a per-minute budget: against a live workspace, 250 sequential `conversations.history` calls drew zero 429s while 5 concurrent ones tripped it. So one `conversations.history` window fills a whole day — `chat.jsonl`'s bytes, the `threads/` listing, and `files/` with exact sizes — and a thread is a *directory* rather than a file, because inlining replies would spend `1 + N` calls per day whether or not anything read them. That split also settles where an attachment posted inside a thread lives, which the day's window cannot see.

Every entry name carries its id (`<name>__C0123`), so a path resolves from its segments alone — no parent walk. That is what lets a cold deep path work after a restart, and why macOS NFD names are harmless: only the id is ever used.

## Which days a conversation has

Slack has no endpoint for that. The obvious substitute — the calendar between `created` and the newest message — answers it badly: one channel in the synthetic corpus spanned **6,438 days and had messages on 1,130**, so 82% of its tree would be days nothing happened. The reader here is an agent. It cannot tell a quiet day from a failed request, has nothing to rank 6,438 identical directories by, and pays a request to find out.

So the listing walks history instead, newest-first, up to ten pages, and names the days it saw. **Every date directory has messages** — reading all 1,130 of that corpus's days took 0.27s and none was empty. Those pages *are* the days' contents, so the walk is less an added cost than an earlier one.

A walk that stops at the ceiling leaves a region it cannot speak for, so listing and reachability are separate questions:

A conversation's timeline has three marks: when it was created, the oldest day the walk reached — its floor — and its newest message. Where a date falls between them decides the answer.

| the date | | requests |
|---|---|---|
| listed by the walk | served | 0 |
| unlisted, but above the floor | refused: the walk read everything up there, so this day is empty | 0 |
| below `created` | refused: the conversation did not exist yet | 0 |
| below the floor, above `created` | the walk never looked — fetched, and served only if it has messages | 1, and the empty answer is cached too |

The last row keeps old history reachable while search is dormant — naming a date is the only route in — without letting a made-up one become a directory. `created` comes from a listing the mount already holds, so the bound the calendar used to supply survives and `conversations.info` is gone.

Three things this depends on, each of which was wrong first: `listings_complete` must be `false`, or the metadata cache answers `stat` from the parent listing and the resource is never called; the conversation must be checked before its history, or an invented channel id spends a request every time and none of them is cached; and dates must be canonical, because they are compared as strings and chrono accepts `+12026-08-03`, whose first byte sorts below every digit.

## Failure is not emptiness

Slack reports application errors *inside* HTTP 200, so a status-only check reads every permission error as a successful empty response. `SlackAccessor::call` is the single gate, and the classification it produces is what the tree is built on:

- An unjoined public channel answers `ok: true`, `messages: []`, **`is_limited: true`** — withheld, not quiet (5 of 5 in a live workspace, one active three months earlier). Filtered at listing time rather than shown as a directory of empty days.
- `types=public_channel,private_channel` on a token without `groups:read` returns `missing_scope` and **zero** channels. The listing asks per kind and merges; `missing_scope` propagates naming the scope.
- A denied read is served empty once and never cached, or a transient `channel_not_found` would freeze a channel as empty for the whole TTL.
- **Archived channels are listed.** The exclusion cited a comment claiming their history is denied; it is not, and excluding them dropped finished projects entirely.
- Against live Slack, an unaccepted token makes `files.slack.com` 302 to a web login answering 200 — 61,461 bytes of HTML where the listing said 966,750, returned as the file's contents. Downloads now refuse redirects and check length against the listing.

## Who wrote a line

`chat.jsonl` is Slack's own message object plus two names that are deliberately **not** one field. `user_name` is resolved against the member list; `app_name` is what a post calls itself, and anyone who can add a webhook picks that string per message — merged, a forged line renders identically to a real one. Neither is an identity, since display names are not unique; the id is. A default webhook post carries neither name at all (its whole key set is `blocks, bot_id, subtype, text, ts, type`), so `bots.info` resolves it, cached per bot including the empty answer — Slackbot's `B01` answers `bot_not_found`, and every workspace has a Slackbot DM.

`users/<name>__<id>.json` is an index rather than a directory: it exists so a person can be found by filename from above the mount, without reading a single message. Resolving an author does not need it — `chat.jsonl` carries `user_name` beside the id already — so it serves an allowlist of what says who someone is (id, handle, display and real name, title, timezone, bot and deleted flags) and nothing that contacts or profiles them. `users.list` returns far more than that, including email, phone, custom profile fields and the workspace's `is_admin`/`has_2fa` flags; an allowlist keeps a field Slack adds later out of the tree until someone decides it belongs. The install does not request `users:read.email` at all.

## Also here

`ProviderSpec::resolve` becomes async because creating a mount now makes a network call: the browser runs the consent and posts the ten-minute `code`, and the backend exchanges it at `oauth.v2.access` with the app's client secret, which never leaves the server. `ls /users` listed profiles at size 0, which sent the cache wrapper down its eager path and made locating each member a linear scan — quadratic overall, 5.40s of rendering at the synthetic corpus's 10,000 members against 0.012s for sizing each once where the member is already in hand. And a `Retry-After` longer than the ordinary cap is sat out **once**, in full: clamping it to 16s meant five retries all landed back inside the window Slack named. Directory mtimes come from one place per level, so `readdir` and `stat` cannot disagree about the same directory — a thread's is its last reply, a date's is that day's last message, and the date's comes from the walk that named it rather than from whatever the day cache happens to hold. The walk reads newest-first, so it already has the tail of every day it touched, and neither side fetches to answer.

## Test plan

- `cargo test --workspace` — 255 passed, 12 ignored. Naming rules (non-ASCII, collisions, the 200-**byte** cap a 255-character assumption hides until deployment), depth dispatch, which dates a walk lists and which of the rest it still serves, cache eviction, the retry policy, and that `{"ok":false}` becomes an error rather than an empty success.
- `SLACK_USER_TOKEN=… cargo test --workspace slack_live -- --ignored --nocapture` — walks the tree, reads a day, a thread and an attachment whole and ranged, checks that every listed date has messages, and that every directory stats to the mtime it was listed with.
- `SLACK_BASE_URL=<origin> SLACK_USER_TOKEN=… cargo test --workspace slack_mock -- --ignored` — the same walk against the synthetic corpus, where listings page and a walk actually hits its ceiling.

Both integration tests self-skip when their env is unset. All three commands say `--workspace` rather than `-p workspace` on purpose: the crate does not declare the reqwest feature it uses, so building it alone fails. #225 fixes that separately.

## Known gaps

- Read-only, and no token refresh: a Slack token never expires unless the app opts into rotation, which this client does not implement — and which cannot be turned off again once turned on. A rotated token dies after 12 hours; the mount then errors rather than reading empty, since `not_authed`/`invalid_auth` are deliberately outside the soft-fail set.
- A thread belongs to the date of its **root**, so replies arriving later do not surface under the day they were written. Its mtime tracks the last reply, so a live old thread says so — but finding it means opening the day it started. Slack offers no cheaper route.
- An attachment whose declared size is not what Slack serves cannot be read at all, and costs a full download per `ls`. Measured on a live workspace: a Quip file declares 2,718 bytes and serves 3,979, and `download_file` treats the mismatch as fatal. That check is also what catches an HTML login page served as file bytes, so it wants relaxing rather than removing.
- `paginate` budgets retries per *page*, not per operation, so a sustained rate limit multiplies by page count. On the 1-request/minute tier one page sleeps about 91s, so a ten-page date listing blocks the FUSE op behind it for roughly a quarter hour. Nothing above holds a wall-clock ceiling.
- `chat.jsonl` is Slack's own message object with two names added, so it carries whatever else Slack sent with it. On a live sample, attachment metadata alone was 70% of its bytes and included `permalink_public`. `files/` projects the same objects down to four fields; the message line projects nothing.
- The channels listing asks for every public channel in the workspace and keeps the ones the person joined. `users.conversations` returns that set directly — one request instead of up to fifty, and no 10,000-channel ceiling.
- Nothing coalesces concurrent identical fetches, while the FUSE frontend spawns every operation, so a parallel reader multiplies each cold fetch by its thread count.
- `search.messages` is implemented and tested but unreachable: nothing routes `.cmd/<name>` to `Resource::command` yet, and `Resource::prompt()` has the same problem. Before either is wired, search needs a decision — it returns Slack's response unprojected, and it reaches public channels the tree deliberately excludes because the person never joined them.


workspace/src/vfs/cache.rs
```diff
@@ -642,19 +651,43 @@ fn basename(path: &str) -> &str {
...
+}
+
+/// Drop `p`'s own listing and nothing else — not its entry, and nothing beneath
+/// it.
+///
+/// For folding a newly discovered child into an incomplete listing: that child
+/// was just listed, and caching it is the whole point of having read it, so the
+/// sweep must not reach it. [`purge_prefix`] would, since the child's key starts
+/// with `p/`.
+fn purge_listing(inner: &mut Inner, p: &str) {
```
nuri-yoo, line 834: `children.remove(p)` disables vanished-child detection: `set_dir` finds disappeared entries by diffing the previous child list (cache.rs:135), so once this has removed it, `vanished` is always empty and a child missing from the fresh listing keeps stat'ing as present. A probe confirms it:

```
/a/c still present after vanish: true
panicked: vanished child still stats as present
```

Dropping just this line fixes it — `list_dir_entries` checks `expiry` first, so the re-listing is still forced and the child's own listing is still untouched. All 17 `vfs::cache` tests pass after the change, including the new `discovering_an_unlisted_child_disturbs_only_its_parents_listing`.

Reachable for Slack: `scan_history` walks the newest ten pages, so as messages arrive the floor rises and previously listed dates drop out. The setup — `ls` a conversation, then open a date below the floor — is what the mount prompt recommends.

Separately, the comment two lines above the `relist_parent` call — and `base.rs:137` and `cache.rs:545` — name Gmail as the `listings_complete() == false` example. It no longer is: Gmail does not override the method and never reaches this wrapper (`mount.rs:84` returns early with the mirror resource). Slack is the only provider on this path now.

workspace/src/vfs/accessor/slack.rs
```diff
@@ -0,0 +1,971 @@
...
+    fn a_config_with_no_token_is_rejected() {
+        assert!(SlackAccessor::new(&config(None, None)).is_err());
+        assert!(SlackAccessor::new(&config(Some(""), Some(""))).is_err());
+        // And an empty user token falls through to a real bot token.
+        let a = SlackAccessor::new(&config(Some(""), Some("xoxb-bot"))).unwrap();
+        assert_eq!(a.token(), "xoxb-bot");
+        assert!(!a.search_available());
+    }
+
+    /// The soft-fail classification must key off Slack's `error` code, not a
```
nuri-yoo: This sentence is split across two attributes — its second half ("formatted string, and must NOT swallow token-level failures…") is at :883 on `read_denied_covers_per_channel_errors_only`, where the whole thing belongs. Editing accident.

workspace/src/vfs/accessor/slack.rs
```diff
@@ -0,0 +1,971 @@
...
+    async fn download_refuses_a_non_slack_host() {
+        let a = SlackAccessor::new(&config(Some("xoxp-user"), None)).unwrap();
+        let e = a
+            .download_file("https://evil.example.com/f.pdf", None, 1)
+            .await
+            .expect_err("must refuse");
+        assert!(e.to_string().contains("not a Slack host"), "{e}");
+        // Slack's own hosts pass the check (this one fails later, on connect).
+        let e = a
+            .download_file("https://files.slack.com/nope", None, 1)
```
nuri-yoo: `expect_err("no network in tests")` isn't accurate — the host check passes for `files.slack.com`, so `send()` runs:

```
curl https://files.slack.com/nope                       -> 400, 0.12s
cargo test ... download_refuses_a_non_slack_host        -> 0.13s
HTTPS_PROXY=http://127.0.0.1:1 cargo test ... same      -> 0.01s
```

Not flaky (a connect failure is also an `Err`), but an un-ignored unit test reaching an external host. Lifting the host check into its own function and asserting on that covers the same thing without a request.

workspace/src/vfs/path.rs
```diff
@@ -4,6 +4,15 @@
 pub struct MountPath(String);
 
 impl MountPath {
+    /// The path of `name` inside this directory, without the `//` a root join makes.
+    pub fn child(&self, name: &str) -> Self {
```
nuri-yoo, line 8: Line-for-line the same as `cache.rs:384`'s `child_path`, root handling included. That one is what production uses (`cache.rs:491`); this is used only by `slack/tests.rs`. Worth pointing the existing caller here and deleting the free function, so the public API has one of these rather than two.

workspace/src/vfs/resource/slack.rs
```diff
@@ -0,0 +1,1803 @@
...
+/// beside the author fields, never in place of them — the id is the stable identity
+/// — and `<@U0BM…>` in `text` becomes `@name`. Two name keys because only one of
+/// them is Slack's own; see [`author_names`].
+fn message_line(
+    m: &Value,
+    names: &HashMap<String, String>,
+    bots: &HashMap<String, String>,
+) -> Vec<u8> {
+    let (verified, claimed) = author_names(m, names, bots);
+    let mut m = m.clone();
```
nuri-yoo, line 1537: `users/*.json` goes through an allowlist but the message line ships Slack's object verbatim. Known gaps already notes attachment metadata was 70% of the bytes and carried `permalink_public` — a link that opens without a token, which is a different kind of field from the rest of that metadata, and this mount carries DMs. `files/` already projects the same objects down to four fields, so applying that projection to the message line's `files[]` would close the asymmetry with machinery that is already here. Worth deciding whether that lands in this PR or a follow-up.

nuri-yoo (commented):
Read it all and ran it at 999cde6: `cargo test --workspace` gives 255 passed / 12 ignored, exactly what the body claims, and check + clippy (`-p workspace -p agent-k-backend --all-targets`) are clean. I also merged `origin/main` (5dd8800) — no conflicts, `cargo build -p workspace` then works standalone and the suite goes to 264/15, so the test-plan footnote can go.

The date listing is what carries this PR. Splitting "days the walk saw" from "days still reachable by name" beats the calendar, and the three traps around it — the `created` floor, canonical dates so string comparison holds, not caching a truncated walk's oldest day — are each pinned by a test.

One thing to fix before merge (`purge_listing`); the rest are small, and the `chat.jsonl` projection is a scope call rather than a fix.

.env.example
```diff
@@ -26,6 +26,34 @@ GOOGLE_REFRESH_TOKEN=
...
+# a bot token exercises less (no DMs, no search). Against a mock origin, use a
+# token that mock accepts rather than a real xoxp-/xoxb- one.
+SLACK_USER_TOKEN=
+SLACK_BOT_TOKEN=
+# Reroute all Slack calls to a mock/gateway ({base}/slack/api).
+# Deployment config only — never user input: the token exchange sends the client
+# secret, so the host it reaches must come from config, not from a request.
+# Default (unset): https://slack.com/api
+# SLACK_API_BASE_URL=
+# The mount's own tests take the same override as SLACK_BASE_URL (the workspace
```
nuri-yoo: `SLACK_BASE_URL` appears only in this prose, never as a key, so a copied `.env` can't run the mock tests. A commented `# SLACK_BASE_URL=` line would do it.

workspace/src/vfs/cache.rs
selloriwoo: Good catch — fixed in `03c75fe2`. Dropping the expiry alone was enough, as you said. Added a test with a discovery between the two listings; that's the combination none of mine covered.

Gmail comments fixed in `5afb4bf1`, though not as a rename — the `unlink` ones would have claimed Slack can delete messages, so they're marked unreachable instead. Found a fifth site while I was in there (the negative-cache branch).

workspace/src/vfs/accessor/slack.rs
selloriwoo: Fixed in `8a43b502`. Checked both crates for the same thing; this was the only one.

workspace/src/vfs/accessor/slack.rs
selloriwoo: Done in `8660d683` — host check is its own function now, test is a plain `#[test]` with no client. Turned out the leading dot in `.ends_with(".slack.com")` had no test either, so that's covered now.

workspace/src/vfs/path.rs
selloriwoo: Already gone — #222 deleted `child_path` in the merge, and `MountPath::child` is what production uses now.

That merge also removed the eager-sizing pass from `readdir`, which makes one of my Known-gaps entries wrong. Fixed in the body.

workspace/src/vfs/resource/slack.rs
selloriwoo: Landed here — `be2d48e2`. Built on `file_meta` so the line and the listing can't disagree. `{name, size}` and no URL, since reads go through `files/<name>`. That day's `chat.jsonl` went 6,649 → 1,936 bytes.

Envelope passthrough is still open (`blocks`, and `last_read`/`subscribed` are the reader's own state) — left in Known gaps.

.env.example
selloriwoo: Added in `db1ac1a4`.

nuri-yoo (commented):
Re-reviewed at `be2d48e2`. The six from the last round are fixed, and I checked each at this head rather than taking the commit messages for it, so they are resolved.

Verified locally on `be2d48e2`:

- `cargo test --workspace` — 313 passed, 0 failed, 21 ignored
- `cargo clippy -p workspace -p agent-k-backend --all-targets` — clean
- `rustfmt --check` over the touched files with `skip_children=true`, so the nightly-only options in `rustfmt.toml` do not drag untouched files in — clean
- `base.rs`'s new claim that Slack is the only provider returning `listings_complete() == false` — true. It is the only implementor, and Gmail no longer goes through `CachedResource` at all.

To check the design claims rather than read them, I stood a Slack-shaped origin up over a raw `TcpListener` and drove the whole public stack (`WorkspaceFs` -> `CachedResource` -> `SlackResource`) against it, with a paginated history so the walk actually truncates at `SCAN_PAGES`. Every row of the reachability table holds:

| the date | result | requests |
|---|---|---|
| listed by the walk | served | 0 |
| unlisted, above the floor | refused | 0 |
| below the floor, has messages | served | 1, then 0 on repeat |
| below the floor, no messages | refused | 1, then 0 on repeat |
| below `created` | refused | 0 |
| above the newest message | refused | 0 |

The empty answer really is cached, so a wrong guess costs one request per TTL rather than one per `stat`. Reading a below-floor day after that `stat` costs nothing further. Re-listing the conversation after the discovery re-runs `readdir` and spends no request, because the resource still holds the walk — exactly the cost the `listings_complete` doc claims, and no more.

Two more claims measured the same way. A day the walk saw whole reads at 0 requests while the oldest day, which `Dates::whole` deliberately drops, costs 1 — the prefill boundary behaves as written. And mtimes agree at every level: 10 date directories with 0 disagreements between the listing and a later `stat`, and the same for `chat.jsonl`, `files/`, `threads/` and a thread directory, whose mtime is its last reply rather than its root.

Four comments below. Only the 416 one seems worth fixing before merge.

One quantification of a gap you already list. "Nothing coalesces concurrent identical fetches ... a parallel reader multiplies each cold fetch by its thread count" — for `conversations.list` under a partial-scope install it is three times the thread count, because the merged call fails `missing_scope` and each reader then asks per kind. Eight parallel cold reads of one day, measured:

```
history = 8, users.list = 8, conversations.list = 24
```

`conversations.list` is Tier 2, so eight readers are past the per-minute allowance on their own, before any retry. The multiplier is not the same for every method, which the current wording implies.

Two sentences in the description went stale when main merged in: the test plan says 255 passed / 12 ignored (313 / 21 now that #222 is in), and the closing note that all three commands say `--workspace` because the crate does not declare the reqwest feature it uses. #225 merged on 08-10 and is in this PR's merge base, so `cargo test -p workspace --no-run` succeeds today.

workspace/src/vfs/accessor/slack.rs
```diff
@@ -0,0 +1,997 @@
...
+        let mut req = self.files.get(parsed).bearer_auth(self.token());
+        let ranged = range.is_some();
+        if let Some(r) = &range {
+            // Inclusive-end, per HTTP: a 0..10 read asks for bytes 0-9.
+            req = req.header(
+                reqwest::header::RANGE,
+                format!("bytes={}-{}", r.start, r.end.saturating_sub(1).max(r.start)),
+            );
+        }
+        let resp = req.send().await?.error_for_status()?;
```
nuri-yoo, line 711: A 416 is a hard error here, but to a reader walking a file to its end it is a clean EOF. `gdrive.rs:602` already draws that distinction, with the same reasoning in the comment; this is the one accessor that does not.

The path is reachable through the public stack. `whole_or_small` (`cache.rs:494`) sends anything over `MAX_CONTENT_BYTES` (8 MiB) down the ranged branch, and `WorkspaceFile::read_bytes` (`fs.rs:275`) does not clamp against `stat.size` — it issues `offset..offset+count` until the resource answers empty. So the last request of a sequential read starts at `size`, and the empty-range guard above does not catch it, since `end = size + want > start`.

Measured by driving `WorkspaceFs` -> `CachedResource` -> `SlackResource` against a Slack-shaped origin serving a 12 MiB attachment, with the origin answering 416 for a range starting at or past EOF:

```
attachment size  = 12582912
sequential read  = 12582912 bytes, 416s served = 1
EOF WALK: FAILED -> read at offset 12582912: GeneralFailure
read at offset==size -> Err(...)
```

Every byte arrives, and then the read fails. The gdrive branch transplants directly:

```rust
if resp.status() == reqwest::StatusCode::RANGE_NOT_SATISFIABLE {
    return Ok((Vec::new(), true));
}
```

The `true` is load-bearing: it tells `download` the range was applied, so it returns the empty vec instead of slicing it again.

One thing I could not settle without a token is whether `files.slack.com` answers 416 or 200 for that range. 416 is what RFC 9110 specifies and what the Google CDN does per your own comment in `gdrive.rs`. If Slack answers 200 with the whole body instead, the read does terminate — but it pays a full re-download at every EOF, so the branch is worth having either way.

workspace/src/vfs/accessor/slack.rs
```diff
@@ -0,0 +1,997 @@
...
+///
+/// `url_private_download` comes from API data, so a message's file metadata
+/// naming some other host would otherwise leak the token to it. A `base_url`
+/// deployment (mock or gateway) serves files from its own origin, which is then
+/// the only host allowed.
+///
+/// A free function so it can be tested for what it decides rather than by
+/// watching a request fail to leave: every url it accepts is one the caller then
+/// sends a token to, which is not a thing a unit test should be doing.
+fn check_file_host(url: &reqwest::Url, base_url: Option<&str>) -> anyhow::Result<()> {
```
nuri-yoo, line 82: This compares `host_str()` and nothing else, so the origin is not really pinned. Two cases fall out of the code as written:

- Production (`base_url = None`): `http://files.slack.com/...` is accepted, and the next line sends the mount's bearer token over plaintext. The url is API-supplied data, which is the case this gate exists for.
- A `base_url` deployment: with `http://localhost:8000` configured, `http://localhost:9999/...` is accepted, because the port is dropped along with the scheme.

Neither is reachable from Slack's own responses today. But the doc above calls this "the one gate on where an attachment download may send the mount's token", and a bare host match is weaker than that sentence. Comparing `Url::origin()` on both sides settles the `base_url` branch, and requiring `url.scheme() == "https"` alongside the existing suffix check settles the other.

workspace/src/vfs/resource/slack.rs
```diff
@@ -0,0 +1,1845 @@
...
+        } else {
+            Vec::new()
+        };
+        let mut files = Vec::new();
+        for m in &msgs {
+            jsonl.extend_from_slice(&message_line(m, &names, &bots));
+            // Skip the root's own attachments: the day's `files/` already lists
+            // those (it came from `conversations.history`, which returns roots),
+            // and listing them twice would make the same file look like two.
+            if m.get("ts").and_then(Value::as_str) == Some(ts) {
```
nuri-yoo, line 779: This is right, and the day's `files/` is the correct home for a root's attachments. What is missing is the reader-facing half: that same root's line in the thread's own `chat.jsonl` still names the file, because `message_line` projects `files[]` for every message it writes.

```
thread root line = {"files":[{"name":"spec__F1.pdf","size":10}], ...}
thread files/    = []
```

So an agent that opens a thread, reads a filename off the root line, and runs `ls files/` beside it gets nothing. The mount prompt states only the other direction (`slack.rs:124`): "a file posted inside a thread is in THAT thread's files/, never the day's". A clause for the root's own attachments — they stay in the day's `files/` — would make the two halves meet.

workspace/src/vfs/resource/slack.rs
```diff
@@ -0,0 +1,1845 @@
...
+/// link that opens with no token at all, on a mount that carries DMs.
+///
+/// Built from [`file_meta`], the same function the `files/` listing uses, so the
+/// two cannot disagree about which attachments a message has or what they are
+/// called. One Slack sends without a usable download url appears in neither.
+fn listed_files(m: &Value) -> Value {
+    let Some(files) = m.get("files").and_then(Value::as_array) else {
+        return Value::Array(Vec::new());
+    };
+    let mut out: Vec<FileMeta> = files.iter().filter_map(file_meta).collect();
```
nuri-yoo, line 1626: `dedup_names` runs per message here and per day in `build_day`, so the doc's "the two cannot disagree about which attachments a message has or what they are called" does not hold when one file id lands in two messages of the same day, which a re-share produces:

```
day files/ listing = ["spec__F1.pdf", "spec__F1(2).pdf"]
line files = [{"name":"spec__F1.pdf","size":10}]
line files = [{"name":"spec__F1.pdf","size":10}]
```

The listing grows a second entry no message line points at, and both lines point at the first. It still reads: `(2)` carries the same url and downloads the same bytes. So this is a listing wart and an overstated doc sentence rather than a broken read. Deduplicating by file id in `build_day` rather than by name would collapse it, since the id is already what makes the name unique.

PR #218, opened 2026-07-28
Automation events

grf53:
## Description

Extend automations beyond schedules and webhooks. A durable event outbox lets triggers react to internal and external domain events, and the event payload is fed into the prompt as render input. This PR contains the event infrastructure (triggers, dispatcher, render) and the external-source lazy polling built on top of it.

## Design: trigger = single event + rich filter

The real requirement is not pairing distinct events by a correlation key, but a predicate combination over a single event (`path ~ *.pdf` AND `date ≥ X`). So, with no correlation id or join-state, `TriggerSpec::Event { conditions }` is an OR of `{source, filter}` predicates. Multi-step work is handled by a single tool-using agent given a prompt, so chaining/DAG/fan-in orchestration is not built here.

## Design: durable outbox + dispatcher

Producers INSERT events into `events` (the outbox) alongside their domain change. A dispatcher worker matches undispatched events to enabled Event triggers and creates runs, using per-(trigger, event) dedup to prevent at-least-once duplicate firing. `run.succeeded` chaining is bounded by a `chain_depth` loop guard.

### Event infrastructure
- **`state/events.rs`** (new) — `EventsState` outbox.
- **`worker.rs`** — dispatcher (undispatched events → trigger matching → run creation).
- **`state/automation.rs`** — `TriggerSpec::Event`, filter matcher (`eq`/`glob`/`gte`/`lte`/`contains`), `render()` for `{{event.*}}`, run `input`/`event_id`, dedup, `chain_depth`.
- **Webhook body/filter** — `create_webhook_run` carries the POST body as render input; `TriggerSpec::Webhook` gains an optional body `filter` (reuses `event_filter_matches`) — a non-matching POST is accepted (202 `matched:false`) but fires no run.
- Internal producers: `workspace.created`, `knowledge.indexed/removed`.

### External-source lazy polling
- **`state/source_poll.rs`** (new) — per-mount snapshot store + `diff_snapshots`.
- **`worker.rs`** — `spawn_source_poller`/`poll_sources_once`/`poll_mount`/`scan_tree`: scans only provider mounts an enabled Event trigger subscribes to (`s3.*`/`notion.*`), diffs against the snapshot, and emits `s3.object_*`/`notion.page_*` events. The first poll only seeds a baseline.
- **`workspace/`** — hookless `scan_fs`, `WorkspaceMount::provider_str`.

### Migration
Add the `events` outbox and `source_poll_state` tables to `0003_automations.sql`.

## Tests

[...]

PR #226, opened 2026-08-11
fix(mounts): inject the Google OAuth client instead of storing a copy per mount

selloriwoo:
## Upgrading a development database

If you ran any earlier version of this branch, its migrations were numbered 3 and 4 and are
now 5 and 6, so sqlx refuses to start on an applied version it cannot find:

```sql
DELETE FROM _sqlx_migrations WHERE version IN (3, 4);
```

Both, not just 3: `validate_applied_migrations` returns on the first one it misses, so
clearing 3 alone gets you `migration 4 was previously applied but is missing`. `main` has
never had a version 3, so no deployed database is affected.

## What changed

A Gmail or Drive mount row stored three values that belong to the installation rather than
to any one mount, written from the backend's own config at mount-create and identical in
every row. The backend passes them in when it builds a mount instead.

```rust
pub struct GdriveConfig {
-   pub client_id: String,      // deployment-wide, copied into every row
-   pub client_secret: String,  // same
-   pub origins: Origins,       // same, and it carries the token endpoint
    pub refresh_token: String,  // this mount's own, still stored
}
```

Not an encryption change. Plaintext storage is the repo's stated stance (migration `0002`)
and is unchanged: `refresh_token` still sits in the row, as do S3's keys and Notion's API
key. Those are secrets the user supplied for that mount.

## The pair is still needed at runtime

Measured against Google, since the fix is not "stop needing them":

| request | response |
| --- | --- |
| full refresh grant | `200`, `expires_in=3599` |
| no `client_secret` | `400` `client_secret is missing.` |
| no `client_id` | `400` `Could not determine client ID from request.` |
| wrong `client_secret` | `401` `The provided client secret is invalid.` |

An access token lasts an hour, so a mount refreshes for as long as it lives.

## What this fixes

- **Secret rotation.** Each row held its own copy with nothing keeping it current, so
  rotating the deployment's client secret left every existing mount refreshing with the old
  one, and there is no mount-update route to repair them.
- **A row can no longer name the host the secret is POSTed to.** `origins` carries the token
  endpoint, so while it was read from the row, anyone able to write one could redirect the
  refresh POST. A test deserializes such a row and asserts the named host is never contacted.
- **A leaked row is not usable on its own.** The table above is the reason: a refresh token
  alone answers `400`. A narrowing, not a fix, since the refresh token is still in the row.
- **An account key cannot escape the mirror root.** `account_email` becomes a directory name
  and the sanitizer must allow `.`, so `..` survived it; `remove_dir_all` on
  `<mirror>/gmail/..` emptied every account's mirror. The key is now validated, refused
  rather than rewritten, since rewriting would land two accounts on one tree.

## When a mount cannot be built

It stays in the table and fails every read, rather than being left out. An absent prefix is
unroutable, so every operation answers `NotFound`, and the knowledge index treats that as
"this target is gone" and purges every document it had ingested from the source. Failing the
assembly instead would take `/files` and every other provider down with it. The two cases
that reach this are a deployment with no OAuth client and a row whose account key is
unusable.

The distinction survives to the guest: `ForwardFs::stat` answers "missing" only for
`NotFound` and `Forbidden`, and the tunnel maps anything else to `EIO`, so an unreachable
source does not read as an empty one inside the sandbox.

## Migrations

`0005` removes `client_id` and `client_secret`, `0006` removes `origins`, both with
`json_remove` so every other field survives byte for byte. Verified against a real sqlite
database across seven row shapes, including an already-clean row, another provider, invalid
JSON, and a valid non-object config, plus a second run.

Neither attempts to scrub the file, deliberately. An UPDATE leaves the bytes it replaced in
the page, and `secure_delete` does not reach what an earlier statement already freed
(measured: 9 of 20 secrets still recoverable after both migrations). Only a whole-file
rewrite gets to zero, and `VACUUM` is the wrong thing to put in front of a boot: it needs
free space equal to the database, writes the whole database in the clear through a temp file
in `TMPDIR`, and on failure is never recorded, so every restart repeats it. So the remedy is
rotating the Google client, which no deployed database needs here because this has not run in
production yet.

## Tests

`GoogleClient` derives neither `Serialize` nor `Deserialize`, so storing it is a compile
error rather than something a comment asks you not to do, and `build_workspace_vfs` takes it
as an argument so every caller had to say where theirs comes from.

Each new test was checked by removing the mechanism it covers and confirming it fails:

- the stored row carries the mount's own credential and none of the deployment's, read back
  by id, for both providers
- a missing client, and an unusable account key, each cost their own mount and not `/files`
- an unreachable mount reads as failure rather than absence, at the `ForwardFs` layer and
  through the real FUSE handlers
- the access-token refresh that keeps a mount alive past an hour: dropped 60s before expiry,
  and a 401 buys exactly one re-auth
- a stored `origins` cannot redirect the refresh POST
- both migrations, against rows written by hand in the old shape
- an account key that traverses upward destroys nothing, against a real mirror tree

Verified: `workspace` 121, `agent-k` 82, `agent-k-backend` 61, clippy and rustfmt clean on
the touched files.

## Follow-ups, not in this PR

- `POST /mounts/{id}/sync` answers `202 {started: false, running: false}` when no client is
  configured, which a client cannot tell from a benign no-op. Needs a third state.
- Notion collapses every accessor failure into `NotFound`, which is the direction that makes
  the knowledge index purge. Same bug class as the mount case above, one provider over.
- Gmail's accessor has its own copy of the token-refresh and 401-retry logic with no mock
  server behind it, unlike Drive. Folding the duplicate into `google.rs` would let Drive's
  tests cover both.


backend/migrations/0005_drop_google_client_credentials.sql
```diff
@@ -0,0 +1,22 @@
...
+-- rather than to any one mount: the backend already holds them in its own config and
+-- now injects them when it builds a mount. A stored copy was one credential duplicated
+-- per user row, and sitting beside a refresh token it turned a single leaked row into a
+-- working Google credential. A refresh token alone mints nothing.
+--
+-- `config` is a JSON object, so the keys are removed with json_remove rather than by
+-- rewriting the document; every other field (refresh_token, account_email, origins,
+-- index_cap) is preserved exactly. Rows already written without the keys are untouched,
+-- which makes this safe to run against a database that saw both code versions.
+UPDATE workspace_mounts
```
nuri-yoo, line 14: Version 3 is already claimed: #217 and #218 both add `0003_automations.sql`. sqlx derives the version from the numeric prefix and `_sqlx_migrations.version` is `BIGINT PRIMARY KEY`, so with both files present `migrate!("./migrations").run()` fails with `UNIQUE constraint failed: _sqlx_migrations.version` (code 1555). `cargo check` still passes, so the build does not catch it; the existing tests do, in about a dozen places that call `migrate!().run().unwrap()`.

Renumbering this pair to `0005`/`0006` sidesteps it, and #217 being older with #218 stacked on it makes this the cheaper side to move. The `include_str!` in `the_migrations_strip_deployment_config_from_rows_the_old_code_wrote` ties the filenames to the build, so the rename fails to compile rather than silently skipping the check.

workspace/src/vfs/resource/gmail.rs
```diff
@@ -43,15 +43,21 @@ impl GmailResource {
...
         mirror_root: Option<&std::path::Path>,
+        oauth: &GoogleClient,
     ) -> anyhow::Result<Self> {
-        let tree = mirror_root.map(|r| mirror_tree(&account_mirror_dir(r, &config.account_email)));
+        // A row whose account key cannot name a directory fails the mount rather than
+        // serving something: the key decides which mailbox is served, so falling back to a
+        // sanitized guess would hand over whichever tree that guess collided with.
+        let tree = mirror_root
+            .map(|r| account_mirror_dir(r, &config.account_email).map(|d| mirror_tree(&d)))
+            .transpose()?;
```
nuri-yoo: This `?` costs the whole workspace, not just this mount. `build_mounts` propagates it, so `WorkspaceFs::from_config` returns `Err` and the workspace loses `/files` and every other provider along with the Gmail one. Against this head:

```
"..":          WHOLE WORKSPACE FAILED: unusable gmail account key ".."
"no-at-sign":  WHOLE WORKSPACE FAILED: unusable gmail account key "no-at-sign"
"":            WHOLE WORKSPACE FAILED
".":           WHOLE WORKSPACE FAILED
control a@b.c: ok
```

That is the outcome the PR body rules out for the sibling case: "Erroring would take `/files` down with them in every workspace holding a Google source". `unavailable_google_mount` handles that one the way I would want this one handled.

Real Google is safe here, since `account_email` comes from `users.getProfile`. But the threat model this hardening is written for is a row written directly or by a hostile configured origin, and a `GOOGLE_API_BASE_URL` deployment is what supplies that profile response. Under that same premise the row now costs the workspace instead of its own mount, and `create_mount` accepts it, so the failure lands after the row is committed. Recovery is a `DELETE` of the mount, which does work (`a_traversing_account_key_cannot_delete_other_mirrors` covers it), but `/files` stays down until someone works out which mount to remove.

Standing it up through `unavailable_google_mount`, with the offending key in the message, would keep the failure where it belongs. `gc_gmail_mirror` is the one caller that has to keep hard-refusing, and it already does.

One asymmetry worth knowing either way: with `mirror_root: None` the `.map(...).transpose()` skips the check entirely, so the library path never validates the key at all.

workspace/src/vfs/resource/gmail.rs
```diff
@@ -1177,22 +1183,25 @@ mod tests {
...
             origins: match std::env::var("GOOGLE_API_BASE_URL") {
                 Ok(host) => crate::vfs::accessor::Origins::behind(&host),
                 Err(_) => Default::default(),
             },
+        };
+        let config = GmailConfig {
+            refresh_token: std::env::var("GOOGLE_REFRESH_TOKEN").ok()?,
+            // This one *is* read: it names the mirror directory, so a live run against
+            // two accounts must not have them share a tree.
+            account_email: std::env::var("GMAIL_EMAIL").unwrap_or_else(|_| "live-test".into()),
```
nuri-yoo: `"live-test"` has no `@`, which the new `account_mirror_dir` refuses, so `gmail_live_mirror` panics in its own setup whenever `GMAIL_EMAIL` is unset. With the other `GOOGLE_*` vars set to dummies, no network involved:

```
thread 'vfs::resource::gmail::tests::gmail_live_mirror' panicked at workspace/src/vfs/resource/gmail.rs:954:74:
called `Result::unwrap()` on an `Err` value: unusable gmail account key "live-test"
```

`a_key_that_does_not_name_a_leaf_is_refused` lists `"no-at-sign"` among the keys that must be refused, so this default now contradicts it, and the failure points at a line that has nothing to do with what broke. `live-test@example.invalid` keeps the comment above accurate and lets the test reach what it is for.

nuri-yoo (commented):
Reviewed at `0b7886a`, in a worktree pinned to that SHA. The split is right, and leaving `Serialize` off `GoogleClient` is a constraint that stays true after everyone forgets why. Three things I would change first, detail inline:

1. **The migration numbers collide with #217 / #218.** Both add `0003_automations.sql`, and two files at version 3 fail `migrate!().run()` on the `_sqlx_migrations.version` primary key. `cargo check` misses it; the existing tests catch it.
2. **An unusable `account_email` fails the whole `WorkspaceFs`, `/files` included** — the outcome the body rules out for the missing-client case, which is handled correctly.
3. **`live_config()`'s `"live-test"` default no longer passes the new key check**, so `gmail_live_mirror` panics in setup without `GMAIL_EMAIL`.

Smaller: `POST …/mounts/{id}/sync` answers `started: false, running: false` for both "no OAuth client" and "unusable key", which is the documented shape for "already in flight", so a client polls and sees nothing move with the reason only in a `warn!`. And `GoogleOAuth::client()`'s "the only way the pair leaves this struct" is not enforced, since `client_id` and `client_secret` are `pub`, though nothing reads them directly today.

The description trails the branch: `0004` is missing from "What changed", "Behavior with no client configured" still describes the option `787fc4e4d` replaced, the `account_mirror_dir` traversal fix and the `FwdStat` ENOENT-to-EIO change are unmentioned, and the counts are 120 and 61 rather than 117 and 59.

Nothing else I probed came apart, including the `Unavailable` shape, the `ForwardFs::stat` remap against tunnel `lookup`/`getattr`, and old rows still deserializing. Local at this head: `workspace` 120 passed / 9 ignored, `agent-k-backend` 61, `agent-k` 82, clippy clean on both crates after a forced re-lint, rustfmt clean on the touched files.

backend/migrations/0005_drop_google_client_credentials.sql
selloriwoo: Renumbered to `0005`/`0006` in `be134a5a`. A database that applied any earlier version of this branch needs `DELETE FROM _sqlx_migrations WHERE version = 3;`, which the PR body now says.

workspace/src/vfs/resource/gmail.rs
selloriwoo: Right, and it contradicts the branch's own missing-client branch, which I wrote. `731e931b` absorbs a build failure into an unavailable mount instead, with the reason logged; `gc_gmail_mirror` keeps refusing hard. Your asymmetry note too: `account_key` is split out, so the key is checked with or without a mirror root.

workspace/src/vfs/resource/gmail.rs
selloriwoo: Fixed in `731e931b`, `live-test@example.invalid`.

nuri-yoo (approved):
Approving. Re-checked all three at `731e931` with probes of my own: #217's `0003_automations.sql` alongside this branch now migrates clean (`[1, 2, 3, 5, 6]`, was `1555`), a bad account key costs only `/gmail` while `/files` keeps serving, and `gmail_live_mirror` reaches its first request instead of panicking in setup. `workspace` 121 / `agent-k-backend` 61 / `agent-k` 82, clippy and rustfmt clean on the touched files; no `.github/workflows/` here, so that run is the only gate.

One thing before merge, in the description rather than the code: the body does not actually carry the `DELETE FROM _sqlx_migrations` line, and version 3 alone still leaves `migration 4 was previously applied but is missing in the resolved migrations`, so it needs `version IN (3, 4)`. The description also still says `0003` and 117 / 59.

The `sync` response shape and the `GoogleOAuth::client()` comment can wait for a follow-up.

